PR #2, opened 2017-01-11
Allow routing from UDP port besides UART device

edersondisouza:
Added `-u <udp_port>` option to allow mavlink_router to route from an
UDP port on *localhost*.
To keep consistence, UART device now must be informed via `-s UART`
argument. Note that one can't route from both at the same time.

### DIFF
--- a/comm.cpp
+++ b/comm.cpp
@@ -371,7 +371,7 @@ UdpEndpoint::UdpEndpoint()
     bzero(&sockaddr, sizeof(sockaddr));
 }
 
-int UdpEndpoint::open(const char *ip, unsigned long port)
+int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
 {
     const int broadcast_val = 1;
     fd = socket(AF_INET, SOCK_DGRAM, 0);
@@ -384,9 +384,16 @@ int UdpEndpoint::open(const char *ip, unsigned long port)
     sockaddr.sin_addr.s_addr = inet_addr(ip);
     sockaddr.sin_port = htons(port);
 
-    if (setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcast_val, sizeof(broadcast_val))) {
-        log_error_errno(errno, "Error enabling broadcast in socket (%m)");
-        goto fail;
+    if (to_bind) {
+        if (bind(fd, (struct sockaddr *)&sockaddr, sizeof(sockaddr)) < 0) {
+            log_error_errno(errno, "Error binding socket (%m)");
+            goto fail;
+        }
+    } else {
+        if (setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcast_val, sizeof(broadcast_val))) {
+            log_error_errno(errno, "Error enabling broadcast in socket (%m)");
+            goto fail;
+        }
     }
 
     if (fcntl(fd, F_SETFL, O_NONBLOCK | FASYNC) < 0) {
@@ -394,7 +401,7 @@ int UdpEndpoint::open(const char *ip, unsigned long port)
         goto fail;
     }
 
-    log_info("Open %s:%lu", ip, port);
+    log_info("Open %s:%lu %c", ip, port, to_bind ? '*' : ' ');
 
     return fd;
 
--- a/main.cpp
+++ b/main.cpp
@@ -70,14 +70,16 @@ static volatile bool g_should_exit;
 
 static void help(FILE *fp) {
     fprintf(fp,
-            "%s [OPTIONS...] <uart>\n\n"
-            "  -h --help                    Print this message\n"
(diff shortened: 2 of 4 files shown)


comm.cpp
```diff
@@ -320,17 +321,24 @@ int UdpEndpoint::open(const char *ip, unsigned long port)
     sockaddr.sin_addr.s_addr = inet_addr(ip);
     sockaddr.sin_port = htons(port);
 
-    if (setsockopt(fd, SOL_SOCKET, SO_BROADCAST, &broadcast_val, sizeof(broadcast_val))) {
-        log_error_errno(errno, "Error enabling broadcast in socket (%m)");
-        goto fail;
+    if (to_bind) {
```
lucasdemarchi: why do you bind?

binding when using UDP is very weird.

comm.cpp
```diff
@@ -76,7 +76,8 @@ struct _packed_ mavlink_router_mavlink1_header {
 };
 
 Endpoint::Endpoint(const char *name)
-    : _name{name}
+    : _name{name},
+    _last_packet_len{0}
```
lucasdemarchi: use c++11 initialization in the header

Initialization list here is only for fields that are initialized according to the parameters.

And if you were to use it, then the comma should align with ":":

```c++
 Endpoint::Endpoint(const char *name)
     : _name{name}
     , _last_packet_len{0}
```

comm.cpp
lucasdemarchi: more on this: If a UDP socket is unbound and either sendto or connect are called on it, the system will automatically bind it for you and thus the recvfrom call later on will succeed.... so I still don't think this is needed.

comm.cpp
edersondisouza: Ack

comm.cpp
edersondisouza: Well... without binding how would the OS know that this socket is here to receive UDP packets on any given port? Don't forget that it's a different process (simulator) that is sending the UDP packets.
Without binding it just doesn't work: mavlink_router just don't receive any UDP packet.

lucasdemarchi: but you want to *connect* to that IP+port, not bind... the simulator should be running on a known port, not mavlink-router.

comm.cpp
lucasdemarchi: this part has already been applied on master

edersondisouza:
Rebased.

main.cpp
```diff
@@ -69,14 +69,16 @@ static volatile bool g_should_exit;
 
 static void help(FILE *fp) {
     fprintf(fp,
-            "%s [OPTIONS...] <uart>\n\n"
-            "  -h --help                    Print this message\n"
-            "  -b --baudrate                Use baudrate for UART\n"
+            "%s [OPTIONS...] [-s <uart>] [-u <udp_port>]\n\n"
+            "  -s <uart>                    UART device that will be routed\n"
+            "  -u <udp_port>                UDP port that will be routed\n"
```
zehortigoza: Maybe instead of this you cold parse the last parameter, for UART it will always start with '/' for UDP it will start with anything else.

lucasdemarchi: I agree... but you can also test if it's a file, a char device, etc

main.cpp
```diff
@@ -357,25 +377,35 @@ static bool add_endpoints(Mainloop &mainloop)
 
 int main(int argc, char *argv[])
 {
+    int udp_port = 0;
     const char *uartstr = NULL;
     UartEndpoint uart{};
+    UdpEndpoint udp{};
     Mainloop mainloop{};
```
zehortigoza: Better dynamically allocate the right one in g_master.

main.cpp
```diff
@@ -357,25 +377,35 @@ static bool add_endpoints(Mainloop &mainloop)
...
+    if (uartstr) {
+        if (uart.open(uartstr, opt.baudrate) < 0)
+            goto close_log;
 
-    g_master = &uart;
-    mainloop.add_fd(uart.fd, &uart, EPOLLIN);
+        g_master = &uart;
+        mainloop.add_fd(uart.fd, &uart, EPOLLIN);
+    } else {
+        if (udp.open("127.0.0.1", udp_port, true) < 0)
```
zehortigoza: Also get the IP from parameter, this way can make compute board forward messages to host computer than host computer with mavlink-router can forward messages to another networks.

main.cpp
edersondisouza: Testing if it's a file looks good

main.cpp
edersondisouza: Fair

main.cpp
edersondisouza: ack

edersondisouza:
Changed and rebased

edersondisouza:
Updated.

lucasdemarchi:
Applied together with the other PR.

PR #3, opened 2017-01-13
MAVLink CRC check

zehortigoza:


comm.cpp
```diff
@@ -215,6 +252,11 @@ int Endpoint::read_msg(struct buffer *pbuf)
     pbuf->len = expected_size;
 
     return 1;
+
+invalid_msg:
+    rx_buf.len -= expected_size;
+    memmove(rx_buf.data, rx_buf.data + expected_size, rx_buf.len);
```
lucasdemarchi: humn...  I don't think this is the right approach. You are discarding the entire buffer when the start byte should be in the middle, which means we will lose a bit more frames until we recover.

You should start the "Find magic byte as the start byte" above again.

This function is becoming to be unmanageable and should be split in smaller parts. "check_crc()" is an obviously independent part.

zehortigoza:
> Why did we have an increase in the CPU usage in the baseline test? It was consistently on 1.5% before... Or this pidstat tool is doing something different/wrong.

Using top I also had 1% of usage, but there is a increase of CPU usage in this test because I rose the datarate in PX4 to 50000.

main.cpp
```diff
@@ -55,9 +56,11 @@ struct endpoint_address {
 static struct opt {
     long unsigned baudrate;
     struct endpoint_address *ep_addrs;
+    bool report_msg_statistics;
```
lucasdemarchi: I think it's fine for now. But in the long run this would better be either
a) a signal you send via SIGUSR1 and it dumps statistics from an idler in the mainloop
or
b) we have a way to communicate with mavlink-router.... either a custom protocol over unix socket or dbus

comm.cpp
zehortigoza: It is not discarding the entire buffer just the message(expected_size), but this block can be removed and use the _last_packet_len logic to do the same job.

Ok moving to a new method.

zehortigoza:
Forced pushed with the requested fixes.

comm.cpp
lucasdemarchi: By buffer I meant packet.

lucasdemarchi:
Just some minor coding style and nitpicks... see comment on top. Applied, thanks.

lucasdemarchi:
https://github.com/01org/mavlink-router/commit/c322ea699787e7bcc2268b64c589a32919c3dd9e

PR #6, opened 2017-01-23
build: Add helpers for static analysis

anselmolsm:


dagar:
Coverity works pretty well on travis-ci.
https://scan.coverity.com/travis_ci

https://travis-ci.org/01org/mavlink-router

lucasdemarchi:
Yep, if we are only uploading to scan.coverity.com, then we could just setup a CI like travis to do that.  I'd rather configure a cron job on it to do once a day or once a week instead of going with the suggested configuration of having a separate branch.

anselmolsm:
No objections. Just need to obtain some rights on the github repo, to enable the sync with travis. I'll verify this.

lucasdemarchi:
Ok. Meanwhile this is applied so we can use it. Thanks.

PR #8, opened 2017-01-25
Routing

edersondisouza:
This PR is more of a RFC about routing. Note however that commit 6f7afe9 is an independent fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,3 +60,28 @@ connection there will also receive routed packets.
 Pull-requests are accepted on GitHub. Make sure to check coding style with the
 provided script in tools/checkpatch, check for memory leaks with valgrind and
 test on real hardware.
+
+### Samples ###
+
+#### Sender & receiver ####
+
+One can test mavlink-router by using `examples/sender.py` and
+`examples/receiver.py` to simulate traffic of mavlink messages.
+First script send mavlink *ping* messages to a target mavlink system-id, and
+second receives and respond them.
+For instance:
+
+    $ PYTHONPATH=<path-to-pymavlink> python examples/sender.py 127.0.0.1:3000 100 0
+
+Will send mavlink *pings* to UDP port 300. Those pings will have `100` as
+source system and will have `0` as target-system (`0` as target means broadcast).
+Receiver could be set as:
+
+    $ PYTHONPATH=<path-to-pymavlink> python examples/receiver.py 127.0.0.1:4000 50
+
+Where `50` is receiver system id. Then, to route between those:
+
+    $ mavlink-routerd -e 127.0.0.1:4000 0.0.0.0:3000
+
+Note that it's possible to setup multiple senders and receivers to see
+mavlink-router in action.
--- a/comm.cpp
+++ b/comm.cpp
@@ -448,7 +448,9 @@ int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
         goto fail;
     }
 
-    log_info("Open %s:%lu %c", ip, port, to_bind ? '*' : ' ');
+    if (to_bind)
+        sockaddr.sin_port = 0;
+    log_info("Open [%d] %s:%lu %c", fd, ip, port, to_bind ? '*' : ' ');
 
     return fd;
 
@@ -483,6 +485,11 @@ int UdpEndpoint::write_msg(const struct buffer *pbuf)
     /* TODO: send any pending data */
     if (tx_buf.len > 0) {
         ;
+    }
+
+    if (!sockaddr.sin_port) {
+        log_debug("No one ever connected to %d. No one to write for", fd);
+        return 0;
     }
 
     ssize_t r = ::sendto(fd, pbuf->data, pbuf->len, 0,
--- a/comm.cpp
+++ b/comm.cpp
@@ -95,9 +95,12 @@ Endpoint::~Endpoint()
     }
 }
 
-int Endpoint::read_msg(struct buffer *pbuf)
+int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compid)
 {
     bool should_read_more = true;
+    uint32_t msg_id;
+    const mavlink_msg_entry_t *msg_entry;
+    uint8_t *payload;
 
     if (fd < 0) {
         log_error("Trying to read invalid fd");
@@ -183,6 +186,21 @@ int Endpoint::read_msg(struct buffer *pbuf)
         if (rx_buf.len < sizeof(*hdr))
             return 0;
 
+        msg_id = hdr->msgid;
+        payload = rx_buf.data + sizeof(*hdr);
+
+        if (!system_id || !component_id) {
+            system_id = hdr->sysid;
+            component_id = hdr->compid;
+        }
+
+        if (system_id != hdr->sysid)
+            log_warning("Different system_id message for endpoint %d: Current: %u Read: %u", fd,
+                        system_id, hdr->sysid);
+        if (component_id != hdr->compid)
+            log_warning("Different component_id message for endpoint %d: Current: %u Read: %u", fd,
+                        component_id, hdr->compid);
+
         expected_size = sizeof(*hdr);
         expected_size += hdr->payload_len;
         expected_size += checksum_len;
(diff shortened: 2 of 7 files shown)


main.cpp
```diff
@@ -602,8 +602,9 @@ static bool add_endpoints(Mainloop &mainloop)
             return false;
         }
 
-        g_endpoints[i++] = udp.release();
-        mainloop.add_fd(udp->fd, g_endpoints[i], EPOLLIN);
+        g_endpoints[i] = udp.release();
```
lucasdemarchi: ugh. thanks. I fixed this locally and forgot to amend before pushing.
Could you send this fix independently?

Also,  `- Using unique_ptr after release()` is not true... the problem is that we would pass g_endpoints[i+1] to add_fd due to the second issue you pointed out.

comm.cpp
```diff
@@ -195,6 +204,12 @@ int Endpoint::read_msg(struct buffer *pbuf)
         if (rx_buf.len < sizeof(*hdr))
             return 0;
 
+        msg_id = hdr->msgid;
+        payload = rx_buf.data + sizeof(*hdr);
+
+        *sysid = hdr->sysid;
+        *compid = hdr->compid;
```
lucasdemarchi: I think this should be a member of Endpoint instead of reading it from the arguments.... And we only set it once and emit a warning (possibly throttled not to spam the console) if it's read again and different from what's already saved.

lucasdemarchi: Then you can add a getter if you need this outside of Endpoint class.

comm.cpp
```diff
@@ -213,6 +228,20 @@ int Endpoint::read_msg(struct buffer *pbuf)
     if (_crc_check_enabled && !_check_crc())
         return 0;
 
+    msg_entry = mavlink_get_msg_entry(msg_id);
+    if (msg_entry == nullptr) {
+        log_error("No message entry for %u", msg_id);
+    }
+
+    *target_sysid = -1;
+    *target_compid = -1;
```
lucasdemarchi: ok.. this makes sense to be as argument.

comm.cpp
```diff
@@ -213,6 +228,20 @@ int Endpoint::read_msg(struct buffer *pbuf)
...
+
+    if (msg_entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM)
+        *target_sysid = payload[msg_entry->target_system_ofs];
+
+    if (msg_entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_COMPONENT)
+        *target_compid = payload[msg_entry->target_component_ofs];
```
lucasdemarchi: I don't mind the missing `{ }` on if statements with a single line. However be consistent with what is there.... Just some lines above you used braces

main.cpp
edersondisouza: Hmmm... about `release`, I actually tested that - if I use `udp->fd` I still get segfault.

comm.cpp
edersondisouza: Ack

main.cpp
lucasdemarchi: You do, because the udp object doesn't contain the pointer anymore. That was the reason why I used g_endpoints when calling add_fd().  I forgot about the `i` variable though, fixed it locally and didn't amend.

lucasdemarchi: the fix should be:
```c++
g_endpoints[i] = udp.release();
mainloop.add_fd(udp->fd, g_endpoints[i], EPOLLIN);
i++;
```

edersondisouza: That was exactly my first approach - and still segfaulted:

```
==4203== Invalid read of size 4
==4203==    at 0x401707: add_endpoints (main.cpp:606)
==4203==    by 0x401707: main (main.cpp:679)
==4203==  Address 0x28 is not stack'd, malloc'd or (recently) free'd
==4203== 
==4203== 
==4203== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==4203==  Access not within mapped region at address 0x28
==4203==    at 0x401707: add_endpoints (main.cpp:606)
==4203==    by 0x401707: main (main.cpp:679)
```

lucasdemarchi: ahh... yeah, there's the `udp->fd` access which went unnoticed. Thanks.

lucasdemarchi: I cherry-picked this commit.

main.cpp
```diff
@@ -169,7 +176,39 @@ static int parse_argv(int argc, char *argv[], const char **uart)
         return -EINVAL;
     }
 
-    *uart = argv[optind++];
+    if (stat(argv[optind], &st) == -1) {
```
edersondisouza: Fair... but isn't this kind of detailed checking asking for a TOCTOU report from static analysis?

lucasdemarchi: It's already a TOCTOU. It's an argument the user is passing, there isn't much option.

edersondisouza:
Updated.

edersondisouza:
Rebased, updated with: 
 - Changes suggested by checkpatch
 - Removed routing entry from TODO list

edersondisouza:
Updates:
 - Fix segfault on 'Routing messags' patch
 - Two new patches: 
   - Fix an issue with UDP endpoints
   - Handle more than one 'master' endpoint, or endpoints which we listen from

edersondisouza:
One more patch with scripts that simulate mavlink messages. I'd like to squash them on 'routing' patch before merge. Didn't just squash to avoid messing someone review.

otaviobp:
I did some tests with the routing feature and it is working for the scenarios I tested

README.md
```diff
@@ -60,3 +60,28 @@ connection there will also receive routed packets.
...
+
+#### Sender & receiver ####
+
+One can test mavlink-router by using `examples/sender.py` and
+`examples/receiver.py` to simulate traffic of mavlink messages.
+First script send mavlink *ping* messages to a target mavlink system-id, and
+second receives and respond them.
+For instance:
+
+    $ PYTHONPATH=<path-to-pymavlink> python examples/sender.py 127.0.0.1:3000 100 0
```
lucasdemarchi: it's better to remove this PYTHONPATH from here. It's advised to install pymavlink with:

`pip install pymavlink`
or 
`pip2 install pymavlink`

examples/receiver.py
```diff
@@ -0,0 +1,44 @@
...
+
+if len(sys.argv) != 3:
+    print("Usage: %s <ip:udp_port> <system-id>" % (sys.argv[0]))
+    print("Receive mavlink heartbeats on specified interface. Respond with a ping message")
+    quit()
+
+srcSystem = int(sys.argv[2])
+mav = mavutil.mavlink_connection('udpin:' + sys.argv[1], source_system = srcSystem)
+
+while(True):
```
lucasdemarchi: could you add the equivalent of checkpatch but for python... you may use pep8 for that

main.cpp
```diff
@@ -198,44 +197,49 @@ static int parse_argv(int argc, char *argv[], const char **uart, char **udp_addr
...
-        return -EINVAL;
+            struct endpoint_address *e = (struct endpoint_address *)malloc(sizeof(*e));
+            e->next = opt.master_addrs;
+            e->ip = ip;
+            e->port = port;
+            opt.master_addrs = e;
+            log_error("add master %s %lu", ip, port);
+        } else {
+            if (!*uart)
+                *uart = argv[optind];
```
lucasdemarchi: not using braces may be ok, but unbalanced if/else are pretty hard to read.

main.cpp
```diff
@@ -198,44 +197,49 @@ static int parse_argv(int argc, char *argv[], const char **uart, char **udp_addr
...
-        help(stderr);
-        return -EINVAL;
-    } else if (!*uart && !*udp_port) {
-        log_error("Please, inform a UART device or UDP port to route!");
-        help(stderr);
-        return -EINVAL;
+            struct endpoint_address *e = (struct endpoint_address *)malloc(sizeof(*e));
+            e->next = opt.master_addrs;
+            e->ip = ip;
+            e->port = port;
```
lucasdemarchi: endpoint_address could be a class that receives ip and port and later sets its position in the list...
or at least have a function `create_endpoint_adress()`... makes it easier to handle the ownership of objects and avoids leaks.

lucasdemarchi:
Nice. There are some leftovers that can be done on top. Applied, thanks.

PR #9, opened 2017-01-26
Add script to keep coding style consistent

lucasdemarchi:
This is using git-clang-format from LLVM project. You need clang-format installed for this to work. I'm also adding a reference configuration file for clang-format, which is not written in stone, so no pre-commit hook yet.

Let's take a look if this produces a reasonable output. For my basic tests it's working fine.

Use `./tools/checkpatch --diff` to see any diff from the "correct" style and what you have in working tree. If you stage your changes, you can use `./tools/checkpatch` to automatically fix the coding style. Then you can add the changes with `git add` as usual.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,14 @@
+---
+BasedOnStyle:    WebKit
+
+AlignAfterOpenBracket: true
+AllowShortBlocksOnASingleLine: false
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+Language: Cpp
+PointerAlignment: Right
+Standard: Cpp11
+TabWidth: 4
+AllowShortFunctionsOnASingleLine: Inline
+NamespaceIndentation: None
+...
--- a/tools/checkpatch
+++ b/tools/checkpatch
@@ -1,0 +1,542 @@
+#!/usr/bin/env python2
+#
+#===- git-clang-format - ClangFormat Git Integration ---------*- python -*--===#
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+#===------------------------------------------------------------------------===#
+
+r"""                                                                             
+clang-format git integration                                                     
+============================                                                     
+                                                                                 
+This file provides a clang-format integration for git. Put it somewhere in your  
+path and ensure that it is executable. Then, "git clang-format" will invoke      
+clang-format on the changes in current files or a specific commit.               
+                                                                                 
+For further details, run:                                                        
+git clang-format -h                                                              
+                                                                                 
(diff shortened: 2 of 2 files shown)


lucasdemarchi:
@edersondisouza @zehortigoza ping

edersondisouza:
Looks ok. Do you intend to fix existing code? We can end up with inconsistent code as new patches follow the rules, but old ones don't...

lucasdemarchi:
Not sure... need to run clang-format and check it isn't causing any abnormal formatting like astyle does in PX4

lucasdemarchi:
Also, I think checkpatch is a more natural name, but I don't want to pretend this is our own code rather than the git-clang-format script from LLVM. What do you think?

edersondisouza:
Running it as is on all patches so far, leads to some 'controversial' format, indeed - not sure if it's tweak-able or we'll have to live with that.
Regarding the name, I would keep it with its own name - and be clear about using it on README, for instance. But I'm not happy with the 'git' part of the name - if we don't use it as a git sub-command. Maybe a wrapper? =P

lucasdemarchi:
trying on main.cpp, this was one that I don't really understand:

```diff
-    struct buffer buf{};   
+    struct buffer buf {    
+    };                     
```

I don't mind the missing spaces on initialization lists... just the getopt table we may want to add comment to turn off clang-format.  Can you take  a look on the case above?

edersondisouza:
I guess it prefers something like 

```diff
-struct buffer buf {};
+struct buffer buf = {};
```

I also didn't like this one:
```diff
-        { "baudrate",               required_argument,  NULL,   'b' },
-        { "endpoints",              required_argument,  NULL,   'e' },
-        { "report_msg_statistics",  no_argument,        NULL,   'r' },
-        { }
-    };
+        {"baudrate", required_argument, NULL, 'b'},
+        {"endpoints", required_argument, NULL, 'e'},
+        {"report_msg_statistics", no_argument, NULL, 'r'},
+        {}};
```

Besides messing with spacing, it kinda hides guard element...

edersondisouza:
Oh, just noticed you comment about turning it off for that tables.

zehortigoza:
Looks good to me, there is just this case:
-    struct epoll_event epev = { };
+    struct epoll_event epev = {};

But I did not found any parameter to change that and I think it is the same case as the tables.

lucasdemarchi:
- Added some changes to the `.clang-format` file
- Added reference to its original location naming it git-clang-format
- Added a "wrapper" (actually a symlink) named checkpatch
- Added small section on README.md about contributions to the project.

lucasdemarchi:
Applied.

lucasdemarchi:
I'm not applying changes to coding style right now since it would cause unnecessary conflicts and I think there still may have some options to tweak.

PR #10, opened 2017-01-28
Add ULog support

zehortigoza:


main.cpp
```diff
@@ -435,13 +444,31 @@ void Mainloop::loop()
...
+                    handle_read(e);
+
+                if (events[i].events & EPOLLOUT)
+                    handle_canwrite(e);
+
+                continue;
+            }
+
+            if (Timeout *t = dynamic_cast<Timeout *>(p)) {
+                uint64_t val = 0;
```
lucasdemarchi: You don't use dynamic_cast here.... Just make Timeout implement the same interface with `handle_read()` and an empty `handle_canwrite()`.... this implementation you are doing here should be the `handle_read()` of the Timeout.

main.cpp
```diff
@@ -54,10 +57,16 @@ class Mainloop {
...
     void process_tcp_hangups();
+    Timeout *timeout_add(uint32_t timeout_ms, bool (*cb)(void *data), const void *data);
+    void timeout_del(Timeout *t);
 
     int epollfd = -1;
     bool report_msg_statistics = false;
     bool should_process_tcp_hangups = false;
+
+private:
+    Timeout *_timeout_list[MAX_TIMEOUT];
```
lucasdemarchi: use inlist.

main.cpp
```diff
@@ -454,7 +481,84 @@ void Mainloop::loop()
...
+    uint8_t index_placed;
+
+    null_check(t, NULL);
+
+    for (index_placed = 0; index_placed < MAX_TIMEOUT; index_placed++) {
+        if (_timeout_list[index_placed] == NULL) {
+            _timeout_list[index_placed] = t;
+            placed = true;
+            break;
+        }
```
lucasdemarchi: just use a list... would simplify this.

main.cpp
```diff
@@ -454,7 +481,84 @@ void Mainloop::loop()
...
+    }
+
+    if (!placed) {
+        log_error("Maximum limit of timeouts reached");
```
lucasdemarchi: list, so you don'thave this limit

main.cpp
```diff
@@ -454,7 +481,84 @@ void Mainloop::loop()
...
+            continue;
+        }
+
+        _timeout_list[i] = NULL;
+        remove_fd(t->fd);
+        delete t;
+    }
+}
+
+Timeout *Mainloop::timeout_add(uint32_t timeout_ms, bool (*cb)(void *data), const void *data)
```
lucasdemarchi: just like in the structs you are actually using.... msec, not ms

comm.cpp
```diff
@@ -245,6 +245,17 @@ void Endpoint::print_statistics()
            _write_total);
 }
 
+int Endpoint::write_msg(const mavlink_message_t *msg)
+{
+    struct buffer buffer;
+    uint8_t data[MAVLINK_MAX_PACKET_LEN];
+
+    buffer.data = data;
+    buffer.len = mavlink_msg_to_send_buffer(data, msg);
```
lucasdemarchi: we should avoid this data copy... you will actually copy 3 times if you do this (from buffer to msg, from msg to temporary buffer, from temporary buffer to send buffer).... a router should be smarter than that.

comm.cpp
```diff
@@ -93,6 +93,9 @@ Endpoint::~Endpoint()
     if (fd >= 0) {
         ::close(fd);
     }
+
+    free(rx_buf.data);
+    free(tx_buf.data);
```
lucasdemarchi: Applied.

main.cpp
```diff
@@ -686,7 +688,8 @@ int main(int argc, char *argv[])
     if (!add_endpoints(mainloop))
         goto close_log;
 
-    mainloop.report_msg_statistics = opt.report_msg_statistics;
+    if (opt.report_msg_statistics)
+        mainloop.timeout_add(MSEC_PER_SEC, _report_msg_statistics_callback, NULL);
```
lucasdemarchi: I still prefer a signal.

comm.cpp
zehortigoza: There is only one copy here from mavlink_message_t to buffer and buffer will be written to file descriptor.

main.cpp
zehortigoza: So add this methods to Pollable:
```
virtual int handle_read(struct buffer *pbuf) = 0;
virtual int handle_write(const struct buffer *pbuf) = 0;
```
And rename this methods of Endpoint to those methods above?
```
int read_msg(struct buffer *pbuf);
virtual int write_msg(const struct buffer *pbuf) = 0;
```
And implement handle_read() of Timeout to handle everything intern and always return 0?

main.cpp
zehortigoza: I'm using array because you said that we should only support a few timerfd's if we needed more we would need to implement a sorted list of timeouts, but even if you want to keep this limit I don't see any problems in moving to std::list.

lucasdemarchi: why do they need to be sorted?

lucasdemarchi: a list would mean just a `next` pointer.

zehortigoza: Instead of having several timerfd's have just one timerfd and sort a list of timeouts by expiration.

main.cpp
zehortigoza: Waiting an ACK here to do the changes.

main.cpp
```diff
@@ -125,7 +130,7 @@ static int parse_argv(int argc, char *argv[], const char **uart, char **udp_addr
     *uart = NULL;
     *udp_port = 0;
 
-    while ((c = getopt_long(argc, argv, "hb:e:rt:", options, NULL)) >= 0) {
+    while ((c = getopt_long(argc, argv, "hb:e:rt:l", options, NULL)) >= 0) {
```
edersondisouza: I think you need the `:` to state that you option takes an argument.

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+    _waiting_first_msg_offset = false;
+    logging_start_timeout();
+
+    return _timeout;
+}
+
+void ULog::stop()
+{
+    mavlink_message_t msg;
+    mavlink_command_long_t cmd;
```
edersondisouza: Not sure if I got it right, but I'm with the impression that these `mavlink_command_long_t` are read only stuff. Would be useful to have them as `const` - so no need to explicitly `bzero` them - just define once, globally on this class/file, initialised with `{ }` and be happy?

comm.h
```diff
@@ -97,3 +158,31 @@ class TcpEndpoint : public Endpoint {
...
+    int add_fd(int fd, void *data, int events);
+    int mod_fd(int fd, void *data, int events);
+    int remove_fd(int fd);
+    void loop();
+    void handle_read(Endpoint *e);
+    void handle_canwrite(Endpoint *e);
+    void handle_tcp_connection();
+    int write_msg(Endpoint *e, const struct buffer *buf);
+    void process_tcp_hangups();
+    Timeout *timeout_add(uint32_t timeout_ms, bool (*cb)(void *data), const void *data);
```
bkueng: A method documentation would help here. Especially the `bool` return.

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    bzero(&cmd, sizeof(cmd));
+    cmd.command = MAV_CMD_LOGGING_START;
+    cmd.target_component = MAV_COMP_ID_ALL;
+    cmd.target_system = TARGET_SYSTEM_ID;
+
+    mavlink_msg_command_long_encode(SYSTEM_ID, 0, &msg, &cmd);
+    _endpoint->write_msg(&msg);
+
+    return true;
```
bkueng: Limit the number of retries?

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+    _endpoint->write_msg(&msg);
+
+    fflush(_file);
+    fclose(_file);
+    _file = NULL;
+    _expected_seq = 0;
+    _mainloop->set_on_master_msg_callback(NULL, NULL);
+}
+
+bool ULog::handle_fligh_stack_message(struct buffer *buffer)
```
bkueng: typo: fligh**t**

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+    uint32_t msg_id;
+
+    if (mavlink2) {
+        struct mavlink_router_mavlink2_header *msg
+            = (struct mavlink_router_mavlink2_header *)buffer->data;
+        msg_id = msg->msgid;
+    } else {
+        struct mavlink_router_mavlink1_header *msg
+            = (struct mavlink_router_mavlink1_header *)buffer->data;
+        msg_id = msg->msgid;
```
bkueng: We should output a warning here, and potentially abort since it cannot be used with mavlink 1

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+    return true;
+}
+
+void ULog::_logging_data_process(mavlink_logging_data_t *msg)
+{
+    bool drops = false;
+
+    /* Check for message drops */
+    if (_expected_seq != msg->sequence) {
+        drops = true;
```
bkueng: This will be wrong for a duplicated message, in case the flight stack does not receive an ack (in time), and re-sends the message.

ulog.cpp
```diff
@@ -0,0 +1,280 @@
...
+        _expected_seq = msg->sequence;
+    }
+    _expected_seq++;
+
+    /* Waiting for ULog header? */
+    if (_waiting_header) {
+        const uint8_t magic[] = ULOG_MAGIC;
+
+        if (memcmp(magic, msg->data, sizeof(magic)))
```
bkueng: Should log an error.

lucasdemarchi:
@bkueng thanks for the review.

In addition:

 > It would be good to have an option 'log only when armed'. This would avoid generating large log files when the vehicle only stands around.

This should be controlled by the flight stack alone so we don't even receive the messages

> some kind of log rotate (remove old logs) would be nice

Yep, it's one item on the TODO file.

bkueng:
> This should be controlled by the flight stack alone so we don't even receive the messages

Hmm that's not how it's currently designed. Start just means start now. We can extend it, but would need a way to tell the client when to start/end a new file. On the other hand the arming state can simply be retrieved from the `base_mode` in the heartbeat message.

main.cpp
zehortigoza: Thanks

ulog.cpp
zehortigoza: It is not worthy keep allocated in stack memory, normally it will only send 2 message one to start and another to stop.

ulog.cpp
zehortigoza: Thanks

ulog.cpp
edersondisouza: Ah, ok. But with `const`, it would live on read-only memory section afair, not on stack.

ulog.cpp
zehortigoza:
```
drops = (_expected_seq - 1) != msg->sequence;
```
Ok, that would be enough?

bkueng: No this needs more logic. In case of a duplicate, you can just ignore the message, since you know you already handled it before. A duplicate could be further back than just the previous message. Something like this: https://github.com/PX4/Firmware/blob/master/Tools/mavlink_ulog_streaming.py#L150 (looks somewhat ugly in python, you can probably write this better in C++).

zehortigoza:
Thanks, new version here: https://github.com/01org/mavlink-router/pull/18

PR #11, opened 2017-01-31
Use __BASE_FILE__ for logging macro

lucasdemarchi:
Using __FILE__ is not a problem on a dev environment, but it's
inconvenient when using distro-builders like yocto that build sources
out of tree.  It would use the following sring for logging:

	(../git/comm.cpp:333) Open 192.168.1.255:14550
	(../git/comm.cpp:333) Open 192.168.7.255:14550

While now we have:

	(comm.cpp:407) Open 192.168.7.255:14550
	(main.cpp:544) Open TCP 0.0.0.0:5760

__BASE_FILE__ is available on both GCC (tested with 6.3.1) and clang
(tested with 3.8.1)  so we can use it.

### DIFF
--- a/log.h
+++ b/log.h
@@ -41,7 +41,7 @@ int log_internal(int level, int error,
     ({                                                                  \
          int _level = (level), _e = (error);                            \
          (log_get_max_level() >= LOG_PRI(_level))                       \
-            ? log_internal(_level, _e, __FILE__, __LINE__, __VA_ARGS__) \
+            ? log_internal(_level, _e, __BASE_FILE__, __LINE__, __VA_ARGS__) \
             : -abs(_e);                                                 \
         })
 


otaviobp:
+1
Just s/sring/string/ in the commit message

log.h
```diff
@@ -41,7 +41,7 @@ int log_internal(int level, int error,
     ({                                                                  \
          int _level = (level), _e = (error);                            \
          (log_get_max_level() >= LOG_PRI(_level))                       \
-            ? log_internal(_level, _e, __FILE__, __LINE__, __VA_ARGS__) \
+            ? log_internal(_level, _e, __BASE_FILE__, __LINE__, __VA_ARGS__) \
```
edersondisouza: From gcc doc:

> This macro expands to the name of the main input file, in the form of a C string constant. This is the source file that was specified on the command line of the preprocessor or C compiler. 

So, AFAIU, this could still show `foo/bar/baz.cpp` if that's what was passed from build env to compiler.
You said that it's working - so it could be OK - but intended results may not be obtained in all build environments.
Note: clang doc says the same.

lucasdemarchi: you are right: it still adds the path

PR #12, opened 2017-02-01
Return in case of tcp connection error

zehortigoza:


lucasdemarchi:
@edersondisouza review and apply please

lucasdemarchi:
@zehortigoza LGTM

main.cpp
```diff
@@ -412,6 +408,13 @@ void Mainloop::handle_tcp_connection()
     add_fd(fd, tcp, EPOLLIN);
 
     log_debug("Received tcp connection on fd %d", fd);
+    return;
+
+calloc_error:
+    close(fd);
```
edersondisouza: There's a trick here: `close()` may change `errno`, so we could lose track of original error.

zehortigoza: done

PR #13, opened 2017-02-03
Just call mavlink_get_msg_entry() once per message

zehortigoza:
This function will do a binary search to find the msg entry that is expense
better just get it once per message.

### DIFF
--- a/comm.cpp
+++ b/comm.cpp
@@ -30,8 +30,6 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-#include <mavlink.h>
-
 #include "log.h"
 #include "util.h"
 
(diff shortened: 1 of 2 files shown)


comm.cpp
```diff
@@ -246,13 +244,24 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
     _last_packet_len = expected_size;
     _read_total++;
 
-    if (_crc_check_enabled && !_check_crc())
-        return 0;
+    if (_crc_check_enabled) {
+        msg_entry = mavlink_get_msg_entry(msg_id);
```
edersondisouza: As we are, in fact, always going to call `mavlink_get_msg_entry`, I would make this call before any check. It would save us the `if (!_crc_check_enabled)` below.

zehortigoza: If the CRC fails we will not call mavlink_get_msg_entry()

edersondisouza: How, if `_crc_check()` is called after `mavlink_get_msg_entry()`? Indeed, result of later is argument for first...

zehortigoza: You are right, new version pushed.

edersondisouza:
:+1:

zehortigoza:
Merged

PR #15, opened 2017-02-04
Move classes to his own files

zehortigoza:
Some static variables and functions of become member of classes but
in overal there is no logical changes.

### DIFF
--- a/comm.h
+++ b/comm.h
@@ -22,86 +22,46 @@
 #include <errno.h>
 #include <inttypes.h>
 
-#include <mavlink.h>
+#include "macro.h"
 
 struct buffer {
     unsigned int len;
     uint8_t *data;
 };
 
-class Endpoint {
-public:
-    Endpoint(const char *name, bool crc_check_enabled);
-    virtual ~Endpoint();
-
-    int read_msg(struct buffer *pbuf, int *target_system, int *target_component);
-    void print_statistics();
-    virtual int write_msg(const struct buffer *pbuf) = 0;
-    virtual int flush_pending_msgs() = 0;
-
-    uint8_t get_system_id() { return system_id; }
-    uint8_t get_component_id() { return component_id; }
-
-    struct buffer rx_buf;
-    struct buffer tx_buf;
-    int fd = -1;
-
-protected:
-    virtual ssize_t _read_msg(uint8_t *buf, size_t len) = 0;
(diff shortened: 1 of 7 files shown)


mainloop.cpp
```diff
@@ -0,0 +1,433 @@
...
+#include "log.h"
+
+static bool should_exit = false;
+
+static void exit_signal_handler(int signum)
+{
+    should_exit = true;
+}
+
+static void setup_signal_handlers()
```
edersondisouza: Not sure if I got why these should keep as static...

comm.h
```diff
@@ -22,86 +22,46 @@
 #include <errno.h>
```
edersondisouza: Commenting on file: not sure if it's really necessary - or if these structures could live on `endpoint.cpp`, only place they are really used.

mainloop.cpp
zehortigoza: keep static or make it private on mainloop class would be almost the same.

comm.h
zehortigoza: ok

zehortigoza:
Pushed with @edersondisouza requested changes.

mainloop.h
```diff
@@ -0,0 +1,69 @@
...
+    bool add_endpoints(Mainloop &mainloop, const char *uartstr, struct opt *opt);
+
+    int epollfd = -1;
+    bool report_msg_statistics = false;
+    bool should_process_tcp_hangups = false;
+
+private:
+    endpoint_entry *g_tcp_endpoints;
+    Endpoint **g_endpoints;
+    int g_tcp_fd;
```
edersondisouza: Minor nitpick: this `g_` prefix makes no sense anymore.

zehortigoza:
Rebased

mainloop.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+#include <assert.h>
+#include <memory>
+#include <signal.h>
+#include <sys/epoll.h>
+#include <unistd.h>
+
+#include "log.h"
+
+static bool should_exit = false;
```
lucasdemarchi: I thought it was ok since I was reviewing the idea, not the changes... it would only move code around, but then I went to check this one thing and it looked wrong.

It was `static volatile bool g_should_exit;` before and you removed the volatile.

PR #16, opened 2017-02-07
Add timeout and improve message statistics

zehortigoza:
This PR depends on https://github.com/01org/mavlink-router/pull/15

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -132,6 +132,8 @@ mavlink_routerd_SOURCES = \
 	pollable.h \
 	pollable.cpp \
 	router.h \
+	timeout.h \
+	timeout.cpp \
 	util.c \
 	util.h
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -129,6 +129,9 @@ mavlink_routerd_SOURCES = \
 	main.cpp \
 	mainloop.cpp \
 	mainloop.h \
+	pollable.h \
+	pollable.cpp \
+	router.h \
 	util.c \
 	util.h
 
--- a/endpoint.cpp
+++ b/endpoint.cpp
@@ -115,7 +115,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
     bool should_read_more = true;
     uint32_t msg_id;
     const mavlink_msg_entry_t *msg_entry;
-    uint8_t *payload, seq;
+    uint8_t *payload, seq, sysid;
 
     if (fd < 0) {
         log_error("Trying to read invalid fd");
@@ -204,13 +204,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
         msg_id = hdr->msgid;
         payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
-
-        if (!_system_id)
-            _system_id = hdr->sysid;
-
-        if (_system_id != hdr->sysid)
-            log_warning("Different system_id message for endpoint %d: Current: %u Read: %u", fd,
-                        _system_id, hdr->sysid);
+        sysid = hdr->sysid;
 
         expected_size = sizeof(*hdr);
         expected_size += hdr->payload_len;
@@ -227,13 +221,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
         msg_id = hdr->msgid;
         payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
-
-        if (!_system_id)
-            _system_id = hdr->sysid;
-
-        if (_system_id != hdr->sysid)
-            log_warning("Different system_id message for endpoint %d: Current: %u Read: %u", fd,
-                        _system_id, hdr->sysid);
+        sysid = hdr->sysid;
 
         expected_size = sizeof(*hdr);
         expected_size += hdr->payload_len;
@@ -267,6 +255,13 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
 
     _stat.read.handled++;
     _stat.read.handled_bytes += expected_size;
+
+    if (!_system_id && (!_crc_check_enabled || msg_entry))
+        _system_id = sysid;
+
+    if (_system_id && _system_id != sysid)
+        log_warning("Different system_id message for endpoint %d: Current: %u Read: %u", fd,
+                    _system_id, sysid);
 
     *target_sysid = -1;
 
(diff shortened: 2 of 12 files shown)


endpoint.cpp
```diff
@@ -95,7 +95,7 @@ Endpoint::~Endpoint()
     free(tx_buf.data);
 }
 
-int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compid)
+int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
```
lucasdemarchi: > One system id can have several components ids.

In reality we may have multiple systemdid with multiple compid on the same Endpoint.  When doing the route we decide we would assume only one sysid + compid per endpoint.

Have you seem multiple compids in a single endpoint and this is hurting what you are doing here?  We may need to go the other direction and add support for more sysid/compid on a single endpoint rather than removing compid.

lucasdemarchi: @edersondisouza ping

endpoint.cpp
```diff
@@ -87,14 +89,27 @@ Endpoint::Endpoint(const char *name, bool crc_check_enabled)
...
+
+void Endpoint::handle_read()
+{
+    assert(_router);
+
+    int target_sysid;
+    struct buffer buf{};
+
+    if (read_msg(&buf, &target_sysid) > 0)
+        _router->router_msg(&buf, target_sysid, _system_id);
```
lucasdemarchi: in the Endpoint we relied on read_msg() being called on a loop:

```
-    while (endpoint->read_msg(&buf, &target_sysid) > 0)
-        router_msg(endpoint, &buf, target_sysid);
-}
```

why did you change it to one read per loop iteration?

mainloop.cpp
```diff
@@ -114,7 +114,7 @@ int Mainloop::write_msg(Endpoint *e, const struct buffer *buf)
     return r;
 }
 
-void Mainloop::router_msg(Endpoint *sender, struct buffer *buf, int target_sysid)
+void Mainloop::router_msg(struct buffer *buf, int target_sysid, int sender_sysid)
```
lucasdemarchi: not from this commit, but this method should be called "route_msg"... i.e. a verb not a noun.

endpoint.cpp
zehortigoza: Ops my bad, I will go back again to while

mainloop.cpp
zehortigoza: yes it should, thanks

mainloop.h
```diff
@@ -19,21 +19,22 @@
...
+#include "router.h"
 
 struct endpoint_entry {
     struct endpoint_entry *next;
     Endpoint *endpoint;
     bool remove;
 };
 
-class Mainloop {
+class Mainloop : public Router {
```
zehortigoza: I don't think that we will need another Router implementation, Mainloop is only implementing Router because this way Endpoint can include router.h without cause a circular dependency.

zehortigoza:
Fixed Lucas comments

lucasdemarchi:
You can fix that by having some fwd declarations. No need to have Mainloop
implement the Router interface.

On Feb 8, 2017 4:51 PM, "José Roberto de Souza" <notifications@github.com>
wrote:

*@zehortigoza* commented on this pull request.
------------------------------

In mainloop.h <https://github.com/01org/mavlink-router/pull/16>:

>
 struct endpoint_entry {
     struct endpoint_entry *next;
     Endpoint *endpoint;
     bool remove;
 };

-class Mainloop {
+class Mainloop : public Router {

I don't think that we will need another Router implementation, Mainloop is
only implementing Router because this way Endpoint can include router.h
without cause a circular dependency.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/01org/mavlink-router/pull/16>, or mute the thread

lucasdemarchi:
Can you explain the numbers please?

```
                messages with CRC error: 916 0.163405% 48KBytes
                sequence of messages lost: 539945 96.320885%
                messages handled: 19708 792KBytes
```

So you counted 916 messages with crc error. But you are saying you had 539945 messages lost (due to resync, etc).... Given the seq number is uint8_t, it wraps quickly and you can only be sure of `256 * 916 == 234496`.... How come you concluded you have more than 2 times that maximum?

endpoint.cpp
```diff
@@ -251,22 +246,48 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
...
+
+            if (seq > _stat.read.expected_seq)
+                diff = (seq - _stat.read.expected_seq);
+            else
+                diff = (UINT8_MAX - _stat.read.expected_seq) + seq;
+
+            _stat.read.drop_seq_total += diff;
+            _stat.read.total += diff;
+        }
+        _stat.read.expected_seq = seq;
```
lucasdemarchi: checking the math here.... I can't see any obvious thing that's wrong.  The final numbers don't sum up though.  Are you sure the other end is sending correct numbers?  Or do you have any explanation?

Makefile.am
```diff
@@ -129,6 +129,9 @@ mavlink_routerd_SOURCES = \
 	main.cpp \
 	mainloop.cpp \
 	mainloop.h \
+	pollable.h \
```
otaviobp: I am not a c++ expert, but shouldn't we use .hh or .hpp extensions?

pollable.cpp
```diff
@@ -0,0 +1,26 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Copyright (C) 2016  Intel Corporation. All rights reserved.
```
otaviobp: s/2016/2017/
The same for other headers.

zehortigoza:
> Can you explain the numbers please?
>
>```
>                messages with CRC error: 916 0.163405% 48KBytes
>                sequence of messages lost: 539945 96.320885%
>                messages handled: 19708 792KBytes
>```
>
>So you counted 916 messages with crc error. But you are saying you had 539945 messages lost >(due to resync, etc).... Given the seq number is uint8_t, it wraps quickly and you can only be sure of >`256 * 916 == 234496`.... How come you concluded you have more than 2 times that maximum?

The calculation of sequence lost only takes place after CRC, so this `256 * 916 == 234496` do not apply.

Makefile.am
lucasdemarchi: no, please no

pollable.cpp
zehortigoza: thanks

zehortigoza:
Merged

lucasdemarchi:
To clarify here for possible future reference. The issue with the math is that for each message sucessfully received we can have 255 with error .. so the maximum theoretical errored messages are n_success * 255 and not n_crc_errored

PR #17, opened 2017-02-07
tests: Small test script for routing

edersondisouza:
Script creates some endpoints, a mavlink-routerd process and sends
packets trough it. Currently, checking is minimal: checks if all
received packets have correct destinatary and expected quantity of them.

Naturally, tests can be improved on future, but it works as a initial
test demonstration. An issue that may be tackled is non-determinism
caused by the use of threads to each endpoint. Current small testcase
already makes use of `sleep()` to ensure order of packets, but that
may become unfeasible if test case grows. An obvious alternative would
be keeping tests small. But what about stress tests? Maybe more flexible
metrics are the solution then.

'test' target added, so one can issue `make test` to test - pymavlink
on Python path is a requirement.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -143,6 +143,9 @@ heartbeat_print_SOURCES = \
 
 noinst_SCRIPTS += examples/heartbeat-print.py
 
+test: mavlink-routerd
+	/usr/bin/env python $(top_builddir)/tests/test.py $(top_builddir)/mavlink-routerd
+
 # ------------------------------------------------------------------------------
 # coverity
 # ------------------------------------------------------------------------------
--- a/tests/test.py
+++ b/tests/test.py
@@ -1,0 +1,186 @@
+#!/usr/bin/python
+
+# This file is part of the MAVLink Router project
+#
+# Copyright (C) 2017  Intel Corporation. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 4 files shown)


Makefile.am
```diff
@@ -135,6 +135,9 @@ heartbeat_print_SOURCES = \
 
 noinst_SCRIPTS += examples/heartbeat-print.py
 
+test: mavlink-routerd
+	/usr/bin/env python tests/test.py ./mavlink-routerd
```
lucasdemarchi: Srcdir is not necessarily the same as builddir

tests/test.py
```diff
@@ -0,0 +1,185 @@
...
+# Two senders: one send to all (target 0). The other sends to target 100
+sender0 = MavlinkSender("sender0", 10000, 1, 1, 0, 0)
+sender100 = MavlinkSender("sender100", 10000, 2, 1, 100, 1)
+
+# Two receivers.
+receiver100 = MavlinkReceiver("receiver100", 10100, 100, 0)
+receiver101 = MavlinkReceiver("receiver101", 10101, 101, 0)
+
+sender100.start()
+# Wait a bit before starting broadcaster pinger
```
lucasdemarchi: You could instead make everybody send heartbeat and wait for that instead arbitrary delay.

edersondisouza: Hmmm - that's true, but would 'break' this very test: if an endpoint has not been seen, a packet sent to it should be lost. Using this delay before the broadcast, I can ensure this case. Maybe I should change comment to be clearer...

Makefile.am
edersondisouza: ack

tests/test.py
```diff
@@ -0,0 +1,185 @@
...
+                self.mav.mav.ping_send(
+                    int(time.time() * 1000), msg.seq,
+                    msg.get_srcSystem(), msg.get_srcComponent())
+            else:
+                break
+
+    def accept(self, func):
+        return self.success and func(self.received)
+
+
```
otaviobp: It is a good habit to add:
if __name__ == "__main__":
to check if the script is being run or if it is being included. If the idea is to run the code even if it is being included, ignore that. hehe

edersondisouza:
Rebased and updated following comments by @lucasdemarchi and @otaviobp , pending @lucasdemarchi comment about heartbeats.

PR #18, opened 2017-02-09
Add ULog support V2

zehortigoza:
Changes from V1 https://github.com/01org/mavlink-router/pull/10:
- Using new routing features
- Taking care of same sequence retransmission


mainloop.h
```diff
@@ -28,7 +28,7 @@ struct endpoint_entry {
     bool remove;
 };
 
-class Mainloop : public Router {
+class Mainloop : public Router, public Timeout_Manager {
```
lucasdemarchi: Mainloop should **not** inherit from Router in the first place. Don't add double inheritance.  All you need to do is to add some forward declarations in some headers

endpoint.cpp
```diff
@@ -344,6 +344,48 @@ void Endpoint::print_statistics()
     printf("\n}\n");
 }
 
+void Endpoint::untrim_msg(const struct buffer *from, struct buffer *to)
```
lucasdemarchi: Since this is for our internal use... why don't we just mark that the message is trimmed rather than having to copy it around?

zehortigoza: Because the data of buffer can have valid bytes at the end of the current message and the buffer len only have the length of the current message.

lucasdemarchi: But if you mark the message as trimmed you could just write zeros in the only place you need it. You don't need to copy the entire message.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+    /*
+     * As MAVLINK_MSG_ID_LOGGING_DATA_ACKED and MAVLINK_MSG_ID_LOGGING_DATA
+     * are MAVLink 2 only messages lets untrim all messages that match
+     * with the interested msg_ids.
+     */
+    uint8_t data[MAVLINK_MAX_PACKET_LEN * 4];
+    struct buffer msg_buffer { 0, data };
+    uint8_t *payload;
+
+    untrim_msg(buffer, &msg_buffer);
```
lucasdemarchi: here: rather than "untrim" message, you can have another 0-filled buffer that you write after writing the data.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+        mavlink_logging_data_acked_t *ulog_data_acked = (mavlink_logging_data_acked_t *)payload;
+        mavlink_message_t msg;
+        mavlink_logging_ack_t ack;
+
+        ack.sequence = ulog_data_acked->sequence;
+        ack.target_component = MAV_COMP_ID_ALL;
+        ack.target_system = _target_system_id;
+        mavlink_msg_logging_ack_encode(SYSTEM_ID, MAV_COMP_ID_ALL, &msg, &ack);
+        _send_msg(&msg, _target_system_id);
+    }
```
lucasdemarchi: missing fall-through comment

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+        if (!_timeout || cmd->command != MAV_CMD_LOGGING_START)
+            break;
+
+        if (cmd->result == MAV_RESULT_ACCEPTED) {
+            _timeout_manager->del_timeout(_timeout);
+            _timeout = NULL;
+        } else
+            log_error("MAV_CMD_LOGGING_START result(%u) is different than accepted", cmd->result);
+        break;
```
lucasdemarchi: you don't need the "untrimmed" message here.

endpoint.cpp
zehortigoza: what do you mean with: in the only place you need ?
the payload is where the zeros needs to be written.

I can not move everything after payload and add the missing zeros because it would overlap the next message.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+    if (_waiting_header) {
+        const uint8_t magic[] = ULOG_MAGIC;
+
+        if (memcmp(magic, msg->data, sizeof(magic))) {
+            log_error("Invalid ULog Magic number");
+            return;
+        }
+
+        _waiting_header = false;
+        fwrite(msg->data, 1, ULOG_HEADER_SIZE, _file);
```
lucasdemarchi: you should only ever write to the final place if you are certain you received everything.  Or you need a mechanism to undo what you've done before. Imagine the situation in which you received a partial write, you wrote the header but you can't complete the message. You will have a wrong log.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+bool ULog::start()
+{
+    time_t t = time(NULL);
+    struct tm *timeinfo = localtime(&t);
+    char path[1024];
+
+    snprintf(path, sizeof(path), "%s/%i-%02i-%02i_%02i-%02i-%02i.ulg", _logs_dir,
+             timeinfo->tm_year + 1900, timeinfo->tm_mon + 1, timeinfo->tm_mday, timeinfo->tm_hour,
+             timeinfo->tm_min, timeinfo->tm_sec);
+    _file = fopen(path, "wb");
```
lucasdemarchi: this file needs to be opened with cloexec flag and even more important, it should be non-blocking.  And code changed to account for it being non-blocking.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+{
+    while (_buffer_len >= sizeof(struct ulog_msg_header)) {
+        struct ulog_msg_header *header = (struct ulog_msg_header *)_buffer;
+        const uint16_t full_msg_size = header->msg_size + sizeof(struct ulog_msg_header);
+
+        if (full_msg_size > _buffer_len) {
+            break;
+        }
+
+        fwrite(_buffer, 1, full_msg_size, _file);
```
lucasdemarchi: 2 problems here:

1. It's operating on a blocking FILE and this is executing on mavlink-router's mainloop which means you can (and will) slowdown mavlink-router when you start to get bigger files.  It will basically be waiting a lot for IO to complete... you are using the default buffer mode and size from the libc so it will block every time you reach BUFSIZ.
1. You are ignoring the return values.  Returning from fwrite doesn't mean you actually really wrote anything. You can easily test this by creating a 1MB tmpfs mountpoint and log to a file there. When you have full file you will start to receive errors and we shouldn't silently ignore those errors.

endpoint.cpp
lucasdemarchi: 1. write the data you have
1. write the missing zeros
1. write the remaining data you need, if you do

instead of

1. copy everything to a new buffer, zero-extending it
1. write everything

ulog.cpp
zehortigoza: Why not? when the result is MAV_RESULT_ACCEPTED it is zero so it will be trimmed.

ulog.cpp
zehortigoza: ok

main.cpp
```diff
@@ -152,6 +155,10 @@ static int parse_argv(int argc, char *argv[], const char **uart, char **udp_addr
             }
             break;
         }
+        case 'l': {
+            opt.logs_dir = strdup(optarg);
```
edersondisouza: Where do you free it? Anyway, as `optarg` lives while program lives, if you are not changing it, no need to `strdup` it.

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+};
+
+ULog::ULog(const char *logs_dir, Timeout_Manager *timeout_manager)
+    : Endpoint{"ULog", false}
+{
+    assert(timeout_manager);
+    assert(logs_dir);
+
+    _timeout_manager = timeout_manager;
+    _logs_dir = strdup(logs_dir);
```
edersondisouza: Oh, and here it's dupped again - no need to it. Believe in `const char*` magic =D

ulog.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    mavlink_msg_command_long_encode(_system_id, MAV_COMP_ID_ALL, &msg, &cmd);
+    _send_msg(&msg, _target_system_id);
+
+    if (_timeout) {
+        _timeout_manager->del_timeout(_timeout);
+        _timeout = nullptr;
+    }
+
+    fflush(_file);
```
edersondisouza: Unless accounting for errors - which is not the case, no need to `fflush()` before `fclose()`, as the later already performs the first.

endpoint.cpp
zehortigoza: > 1. write the data you have
> 1. write the missing zeros
> 1. write the remaining data you need, if you do
This is exactly what untrim does. Or you mean write to file? That would make the implementation I little bit more complicated even more changing no non-blocking writing.

ulog.cpp
lucasdemarchi: make sure to check ferror()

zehortigoza:
New version with the requested fixes.

ulog.cpp
```diff
@@ -0,0 +1,350 @@
...
+    time_t t = time(NULL);
+    struct tm *timeinfo = localtime(&t);
+    char path[1024];
+
+    if (_file != -1) {
+        log_warning("ULog already started");
+        return false;
+    }
+
+    snprintf(path, sizeof(path), "%s/%i-%02i-%02i_%02i-%02i-%02i.ulg", _logs_dir,
```
lucasdemarchi: _logs_dir + all the names can be more than 1024. Use asprintf here. It's not like we are starting and stop thousand times per second.

ulog.cpp
```diff
@@ -0,0 +1,350 @@
...
+{
+    _stat.read.total++;
+    _stat.read.handled++;
+    _stat.read.handled_bytes += msg->len;
+    if (msg->magic == MAVLINK_STX)
+        _stat.read.handled_bytes += sizeof(struct mavlink_router_mavlink2_header);
+    else
+        _stat.read.handled_bytes += sizeof(struct mavlink_router_mavlink1_header);
+
+    _mainloop->route_msg(msg, target_sysid);
```
lucasdemarchi: normally the mainloop/router shouldn't copy message. The endpoints copy them when they needed. This is the only place in which `route_msg()` is called with a `mavlink_message_t*` as parameter which changes the "normal" behavior of route_msg(): instead of just routing it will actually copy data to stack and call the write_msg() on the other endpoint.

Make the copy here, remove the additional route_msg() methodd  and then call the older one

ulog.cpp
```diff
@@ -0,0 +1,350 @@
...
+{
+    /* Check for message drops */
+    if (_expected_seq != seq) {
+        *drop = true;
+
+        /*
+         * Handle case were Flight Stack did not received a ACK and sent the
+         * same message again.
+         */
+        if (_expected_seq - 1 == seq)
```
lucasdemarchi: this doesn't seem to handle the case @bkueng commented here: https://github.com/01org/mavlink-router/pull/10/files#r100004968 - "A duplicate could be further back than just the previous message"

ulog.cpp
```diff
@@ -0,0 +1,350 @@
...
+        struct ulog_msg_header *header = (struct ulog_msg_header *)_buffer;
+        const uint16_t full_msg_size = header->msg_size + sizeof(struct ulog_msg_header);
+
+        if (full_msg_size > _buffer_len) {
+            break;
+        }
+
+        const ssize_t r = write(_file, _buffer, full_msg_size);
+        if (r == -EAGAIN)
+            return;
```
lucasdemarchi: we are silently dropping the message here

ulog.cpp
```diff
@@ -0,0 +1,350 @@
...
+        if (r != full_msg_size) {
```
lucasdemarchi: we potentially wrote part of the message. We should still recalculate _buffer_len below or we will corrupt the log when we enter this function again on next write.

ulog.cpp
lucasdemarchi: Ok... I guess it's ok to log only on partial writes below since what will happen is that  we may end up filling our buffer later if the underlining storage keeps returning -EAGAIN.

lucasdemarchi: comment below still needs to be handled though.

zehortigoza:
New version with the requested fixes.

ulog.cpp
```diff
@@ -0,0 +1,415 @@
...
+            return;
+        }
+        if (msg->length < ULOG_HEADER_SIZE) {
+            /* This should never happen */
+            log_error("ULog header is not completed");
+            return;
+        }
+
+        const ssize_t r = write(_file, msg->data, ULOG_HEADER_SIZE);
+        if (r != ULOG_HEADER_SIZE) {
```
edersondisouza: Is this check meaningful? I mean, `_file` is `O_NONBLOCK`, so does it return anything meaningful here? AFAICU, it all depends on underlying implementation - default being just ignoring `O_NONBLOCK` for regular files. Default makes check meaningful - but `O_NONBLOCK` not.

ulog.cpp
```diff
@@ -0,0 +1,415 @@
...
+    memcpy(&_buffer[_buffer_len], &msg->data[begin], msg->length - begin);
+    _buffer_len += (msg->length - begin);
+    _logging_flush();
+}
+
+void ULog::_logging_flush()
+{
+    while (_buffer_partial_len) {
+        const ssize_t r = write(_file, _buffer_partial, _buffer_partial_len);
+        if (r == -EAGAIN)
```
edersondisouza: `write` returns `-1` and sets `errno`, it doesn't return `-errno`.

ulog.cpp
```diff
@@ -0,0 +1,415 @@
...
+            return;
+        if (r < 1) {
+            log_error_errno(errno, "Unable to write to ULog file: (%m)");
```
edersondisouza: Ditto regarding meaning of `write` return.

ulog.cpp
```diff
@@ -0,0 +1,415 @@
...
+            break;
+        }
+
+        const ssize_t r = write(_file, _buffer, full_msg_size);
+        if (r == full_msg_size) {
+            _buffer_len -= full_msg_size;
+            memmove(_buffer, &_buffer[full_msg_size], _buffer_len);
+            continue;
+        }
+        if (r == -EAGAIN)
```
edersondisouza: Ditto regarding `-errno`.

ulog.cpp
```diff
@@ -0,0 +1,415 @@
...
+        const uint16_t full_msg_size = header->msg_size + sizeof(struct ulog_msg_header);
+
+        if (full_msg_size > _buffer_len) {
+            break;
+        }
+
+        const ssize_t r = write(_file, _buffer, full_msg_size);
+        if (r == full_msg_size) {
+            _buffer_len -= full_msg_size;
+            memmove(_buffer, &_buffer[full_msg_size], _buffer_len);
```
edersondisouza: Why the `memove`? Isn't possible to just adjust some offset to next write?

ulog.h
```diff
@@ -0,0 +1,61 @@
...
+#include "endpoint.h"
+#include "timeout.h"
+
+#define TARGET_SYSTEM_ID 1
+#define SYSTEM_ID 2
+
+class ULog : public Endpoint {
+public:
+    ULog(const char *logs_dir);
+    virtual ~ULog();
```
edersondisouza: Why destructor is `virtual` but its implementation is empty?

ulog.cpp
zehortigoza: It is possible reduce the number of memmoves() I will try that.

zehortigoza:
Rebased and fixed @edersondisouza comments

ulog.h
```diff
@@ -0,0 +1,64 @@
...
+#define SYSTEM_ID 2
+
+#define BUFFER_LEN 2048
+
+class ULog : public Endpoint {
+public:
+    ULog(const char *logs_dir);
+
+    bool start();
+    bool logging_start_timeout();
```
edersondisouza: Why is this public?

ulog.cpp
```diff
@@ -0,0 +1,429 @@
...
+        const ssize_t r = write(_file, _buffer_partial, _buffer_partial_len);
+        if (r == 0 || (r == -1 && errno == EAGAIN))
+            return true;
+        if (r < 0) {
+            log_error_errno(errno, "Unable to write to ULog file: (%m)");
+            return false;
+        }
+
+        _buffer_partial_len -= r;
+        memmove(_buffer_partial, &_buffer_partial[r], _buffer_partial_len);
```
edersondisouza: Why not use this `r` as an offset to write?
Then you'd have something like:

```cpp
ssize_t r = 0;
while (_buffer_partial_len) {
    r = write(_file, _buffer_partial + r, _buffer_partial_len);
[...]
```

And you could get rid of this `memmove` altogether?

ulog.cpp
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    while (_buffer_len >= sizeof(struct ulog_msg_header) && !_buffer_partial_len) {
+        struct ulog_msg_header *header = (struct ulog_msg_header *)&_buffer[_buffer_index];
+        const uint16_t full_msg_size = header->msg_size + sizeof(struct ulog_msg_header);
+
+        if (full_msg_size > _buffer_len) {
+            break;
```
edersondisouza: Why return `true` if message couldn't be written by being too big?

ulog.h
```diff
@@ -0,0 +1,64 @@
...
+
+    uint16_t _expected_seq;
+    bool _waiting_header;
+    bool _waiting_first_msg_offset;
+    uint8_t _buffer[BUFFER_LEN];
+    uint16_t _buffer_len;
+    /* Where valid data starts on buffer */
+    uint16_t _buffer_index;
+    uint8_t _buffer_partial[BUFFER_LEN / 2];
+    uint16_t _buffer_partial_len;
```
edersondisouza: Maybe it's a good idea explain why this buffer partial, and about writing strategy being used. I'm with the impression that - maybe to avoid blocking? - it tries to write all messages in `_buffer` (message by message, why? why not the whole buffer?), but if some partial writing happens, it will store remaining messages on a partial buffer and try to write later.
If I got anything wrong, maybe it's a good idea to explain what this code is trying to do and why, to save ourselves from future =D

ulog.cpp
zehortigoza: It is only returning false in case of a write() error, this way at the stop() it will try write the _buffer_partial.

ulog.cpp
zehortigoza: I don't want to add more complexity to the buffer_partial, in 99% of the time it will not be used, the write() of the block bellow will write all the bytes at once and even if there some by bytes left this block will write all missing bytes causing a memmove() of 0 bytes.

ulog.h
zehortigoza: Because it will be called from the timeout callback.

```
static bool _ulog_timeout_cb(void *data)
{
    ULog *ulog = static_cast<ULog *>(data);
    return ulog->logging_start_timeout();
}
```

ulog.h
zehortigoza: It is because of the ULog messages, we can not write only part of a ULog message, so if write() only writes half of the ULog message we copy it to _buffer_partial and write it to file before parsing and writing another ULog message.

edersondisouza: Hmmm... think I've got it. Maybe some comment on code or commit message, to make this information more perennial.

ulog.h
edersondisouza: Ah - ok, didn't notice it was a `static` thing.

zehortigoza:
@lucasdemarchi ping

lucasdemarchi:
Applied

PR #20, opened 2017-02-16
Conf files - v2

edersondisouza:
Changes from v1:
  - Accept `#` as a comment to Linuxfy a bit
  - New `strcasecmp` macro to compare strings, ignoring case
  - `null_check` macro is now `assert_or_return`.
  - New patch adding possibility to have a dir in which .conf files can be put, allowing one to add and override previous configuration. Files on this dir are read in alphabetical order - allowing one to chain overrides.
   -- This last patch involved some refactor - review it with care =D

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,27 @@ Port = 11000
 Note that `Port` is optional for endpoints whose `mode` is `normal`, as
 well as `baud` for UART endpoints.
 
+#### Conf dirs ####
+
+Besides default conf file, it's also possible to use a directory in where to
+put some extra configuration files. Files on such directory will be read in
+alphabetical order, and can add or override configurations found on previous
+files.
+
+By default, `/etc/mavlink-router/config.d` is the directory, but it can be
+overriden via `MAVLINK_ROUTER_CONF_DIR` environment variable, or via `-d`
+switch when running mavlink-routerd.
+
+Suppose default configuration file contents are the same of section above,
+an example of overriding configuration would be:
+
+```ini
+[Endpoint bravo]
+Baud = 115200
+```
+
+That would change `Endpoint bravo` baudrate to `115200`.
+
 ### Contributing ###
 
 Pull-requests are accepted on GitHub. Make sure to check coding style with the
--- a/README.md
+++ b/README.md
@@ -43,17 +43,52 @@ Install:
 To route mavlink packets from master `ttyS1` to 2 other UDP endpoints, do as
 following:
 
-    $ mavlink-routerd -b 1500000 -e 192.168.7.1:14550 -e 127.0.0.1:14550 /dev/ttyS1
+    $ mavlink-routerd -e 192.168.7.1:14550 -e 127.0.0.1:14550 /dev/ttyS1:1500000
 
-The `-b` switch above is used to set the UART baudrate. See more options with
-`mavlink-routerd --help`
+The `1500000` after colon above on `/dev/ttyS1:1500000` is used to set the
+UART baudrate. See more options with `mavlink-routerd --help`
 
 It's also possible to route mavlinks packets from any interface using:
 
-    $ mavlink-routerd -b 1500000 -e 192.168.7.1:14550 -e 127.0.0.1:14550  0.0.0.0:24550
+    $ mavlink-routerd -e 192.168.7.1:14550 -e 127.0.0.1:14550  0.0.0.0:24550
 
 mavlink-router also listens, by default, port 5760 for TCP connections. Any
 connection there will also receive routed packets.
+
+### Conf file ###
+
+It's also possible to use a .conf file to set options for mavlink-routerd.
+By default, mavlink-routerd looks for a file
+`/etc/mavlink-router/main.conf`. File location can be overriden via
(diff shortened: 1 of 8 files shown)


lucasdemarchi:
> -- This last patch involved some refactor - review it with care =D

where is the testsuite? :)

log.h
```diff
@@ -61,7 +61,7 @@ int log_internal(int level, int error,
 #define log_warning_errno(error, ...)   log_full_errno(LOG_WARNING, error, __VA_ARGS__)
 #define log_error_errno(error, ...)     log_full_errno(LOG_ERR,     error, __VA_ARGS__)
 
-#define null_check(ptr, ...)                 \
+#define assert_or_return(ptr, ...)           \
```
lucasdemarchi: s/ptr/exp/ and handle it as an expression

edersondisouza:
> where is the testsuite? :)

```
$ make test
/usr/bin/env python ./tests/test.py ./mavlink-routerd
ERROR LOADING MAVNATIVE - falling back to python implementation
All tests OK
```

=D

README.md
```diff
@@ -43,18 +43,53 @@ Install:
 To route mavlink packets from master `ttyS1` to 2 other UDP endpoints, do as
 following:
 
-    $ mavlink-routerd -b 1500000 -e 192.168.7.1:14550 -e 127.0.0.1:14550 /dev/ttyS1
+    $ mavlink-routerd -e 192.168.7.1:14550 -e 127.0.0.1:14550 /dev/ttyS1:1500000
```
otaviobp: ++ for this change

conf.cpp
```diff
@@ -121,7 +121,7 @@ int ConfFile::parse_file()
     return ret;
 }
 
-const char* ConfFile::next_section(bool reset)
+const char *ConfFile::next_section(bool reset)
```
otaviobp: move to previous commit

conf.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    while ((read = getline(&entry, &n, file)) >= 0) {
+        line++;
+
+        _trim(&entry);
+
+        switch (entry[0]) {
+        case ';':
+        case '#':
```
lucasdemarchi: people can just comment on the line above.   commenting inline in the end of the line just causes trouble... you would need to add a way to escape # when it's needed

edersondisouza: Yeah - or to be more natural, accept quotes `"` to enclose a value, a need to escape them...
Not sure if worth to tackle now

conf.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    char *end = strchr(entry, ']');
+    if (!end) {
+        log_error("On file %s: Line %d: Unfinished section name. Expected ']'", _filename, line);
+        return -EINVAL;
+    }
+
+    // So section name is the string between '[]'
+    *entry = ' ';
+    *end = '\0';
```
edersondisouza: Do you mean like `[foo]bar`? Sounds fair

conf.cpp
```diff
@@ -0,0 +1,276 @@
...
+            _current_section->current_config = config;
+            return config->value;
+        }
+        config = config->next;
+    }
+
+    return NULL;
+}
+
+void ConfFile::_trim(char **str)
```
edersondisouza: Sure! when I started writing it I thought I would... hehe

edersondisouza:
Updated to address @otaviobp comments.

conf.cpp
```diff
@@ -0,0 +1,276 @@
...
+        return -EINVAL;
+    }
+
+end:
+    free(entry);
+    fclose(file);
+    return ret;
+}
+
+const char* ConfFile::next_section(bool reset)
```
edersondisouza: Hmmm... forgot this one. I'll try it.

edersondisouza:
Now all comments should've been addressed.

conf.cpp
otaviobp: yep

main.cpp
```diff
@@ -376,6 +463,68 @@ static int parse_conf()
...
+    if (ret < 0 && ret != -EIO) {
+        return ret;
+    }
+
+    dirname = get_conf_dir();
+    // Then, parse all files on configuration directory
+    dir = opendir(dirname);
+    if (!dir) {
+        return 0;
+    }
```
lucasdemarchi: Every return here one leaks dir.

lucasdemarchi:
quickly scanned the changes, LGTM. There's certainly space for improvement, but it can be done later.

One small fix to do that I commented above. Please do on top.

main.cpp
edersondisouza: ? This `return` happens when `dir` is `NULL`, maybe you commented on wrong line?

lucasdemarchi:
I meant all returns below this line.

main.cpp
lucasdemarchi: Here on... Like in "returns below this"

edersondisouza: Ah, maybe I'd have got 'from here' - sorry.
Will fix it - although any early return here leads to program termination.

PR #23, opened 2017-02-22
Set the default UDP port when parsing command line

zehortigoza:


main.cpp
```diff
@@ -125,7 +125,9 @@ static int add_endpoint_address(struct endpoint_config *conf, const char *name,
         assert_or_return(conf->address, -ENOMEM);
     }
 
-    if (port != ULONG_MAX) {
+    if (port == ULONG_MAX) {
```
edersondisouza: Hmmm... I guess that is not that simple: `add_endpoint_address` can receive `ULONG_MAX` if this endpoint conf is being overriden and `port` is not being set. So, you'll need to check after current code, if `conf->port == ULONG_MAX`, as it's done for `baudrate` on `add_uart_endpoint`. Which made me see that I forgot to initialise baudrate there...  :sob:

edersondisouza:
As #24 was merged, this one was covered.

PR #27, opened 2017-02-28
Log

edersondisouza:


endpoint.cpp
```diff
@@ -230,7 +230,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
     *target_sysid = -1;
 
     if (msg_entry == nullptr)
-        log_error("No message entry for %u", msg_id);
+        log_debug("No message entry for %u", msg_id);
```
lucasdemarchi: we also need to fix this unbalanced use of braces on if and else.  And update the coding style script to catch it.

main.cpp
```diff
@@ -369,6 +376,15 @@ static int parse_conf(const char *conf_file_name)
         opt.logs_dir = value;
     }
 
+    value = conf.next_from_section("General", "verbose");
```
lucasdemarchi: I think it would be better to allow setting the log level on configuration and parse the argument as a string.

LogLevel = error | warning | info | debug

edersondisouza: Makes sense

edersondisouza:
Updated to address review comments.

edersondisouza:
There's a log dir feature that it's not related to this one. I've used `debug` to differentiate them.

lucasdemarchi:
> There's a log dir feature that it's not related to this one. I've used debug to differentiate them.

ohhh... ok. Just handle the -v vs --debug-log-level then?

edersondisouza:
Update to address comments regarding -v vs --debug-log-level

lucasdemarchi:
Applied, thanks.

PR #28, opened 2017-02-28
configure: allow different mavlink dialect to be used

edersondisouza:
This way, mavlink-router is able to understand different MAVLink
messages definitions. To do so, one needs to inform the different
`common.xml` file via MAVLINK_XML option to configure:

    $ ./configure MAVLINK_XML=path/to/common.xml

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,24 +14,28 @@ AM_MAKEFLAGS = --no-print-directory
 GCC_COLORS ?= 'yes'
 export GCC_COLORS
 
-BUILT_SOURCES = include/mavlink/common/mavlink.h
+DIALECT=@DIALECT@
+
+BUILT_SOURCES = include/mavlink/$(DIALECT)/mavlink.h
+
+MAVLINK_XML=$(DIALECT).xml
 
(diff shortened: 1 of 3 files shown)


lucasdemarchi:
Humn..  I wouldn't like to support external xml, only those present on the submodule. The thing is, there may be different dialects of mavlink. Look at modules/mavlink/message_definitions/v1.0/ardupilotmega.xml and how it includes the common.xml.  The generator will do the right thing if you just point it to ardupilotmega.xml, like is done here:
https://github.com/ArduPilot/ardupilot/blob/master/wscript#L235

lucasdemarchi:
I think that what we want is a `--with-mavlink-dialect=`, that accepts ardupilotmega.xml or common.xml.

mbelluzzo:
Wasn't hard enough to keep everything in think :-)

edersondisouza:
> Wasn't hard enough to keep everything in think :-)
?

edersondisouza:
Update to allow only changing mavlink dialect, between `common` and `ardupilotmega`.

PR #30, opened 2017-03-03
Ulog improvements

zehortigoza:


ulog.cpp
```diff
@@ -73,19 +73,41 @@ bool ULog::start()
...
     }
 
-    r = asprintf(&filename, "%s/%i-%02i-%02i_%02i-%02i-%02i.ulg", _logs_dir,
-                 timeinfo->tm_year + 1900, timeinfo->tm_mon + 1, timeinfo->tm_mday,
-                 timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec);
-    if (r < 1) {
-        log_error_errno(errno, "Error formatting ULog file name: (%m)");
+    for (uint16_t i = 0; i < UINT16_MAX; i++) {
+        if (i) {
+            r = asprintf(&filename, "%s/%i-%02i-%02i_%02i-%02i-%02i_%u.ulg", _logs_dir,
```
edersondisouza: Not sure how many times do we expect this to run, but a loop of `malloc` and `free` doesn't look very good. Why not the famous `buffer[PATH_MAX]` and `snprintf`? At the end, if found some name, a `strdup` would suffice.

zehortigoza: it should not run more than 10 times because we will also implement some log rotation.
for me both solutions are fine as this chain of malloc and free will run just once and at the initialization.

edersondisouza: Fair then =D

edersondisouza:
lgtm

PR #31, opened 2017-03-13
Ardupilot log

edersondisouza:
Initial attempt - data gathered from ardupilot sitl appears to make sense.
There are some `TODO` that are more of doubts on code - any help/idea is appreciated!

### DIFF
--- a/binlog.cpp
+++ b/binlog.cpp
@@ -1,0 +1,168 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Copyright (C) 2017  Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "binlog.h"
+
+#include <assert.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "log.h"
+#include "mainloop.h"
+#include "util.h"
+
+bool BinLog::_start_timeout()
+{
+    mavlink_message_t msg;
+
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, MAV_REMOTE_LOG_DATA_BLOCK_START, 1);
+    _send_msg(&msg, _target_system_id);
+
+    return true;
+}
+
+bool BinLog::start()
+{
+    if (!LogEndpoint::start()) {
+        return false;
+    }
+
+    _last_acked_seqno = 0;
+    _system_id = SYSTEM_ID;
+
+    return true;
+}
+
+void BinLog::stop()
+{
+    mavlink_message_t msg;
+
+    if (_file == -1) {
+        log_error("BinLog not started");
+        return;
+    }
+
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, MAV_REMOTE_LOG_DATA_BLOCK_STOP, 1);
+    _send_msg(&msg, _target_system_id);
+
+    LogEndpoint::stop();
+}
+
+int BinLog::write_msg(const struct buffer *buffer)
+{
+    const bool mavlink2 = buffer->data[0] == MAVLINK_STX;
+    uint32_t msg_id;
+    uint16_t payload_len;
+    uint8_t trimmed_zeros;
+
+    if (mavlink2) {
+        struct mavlink_router_mavlink2_header *msg
+            = (struct mavlink_router_mavlink2_header *)buffer->data;
+        msg_id = msg->msgid;
+        payload_len = msg->payload_len;
+    } else {
+        struct mavlink_router_mavlink1_header *msg
+            = (struct mavlink_router_mavlink1_header *)buffer->data;
+        msg_id = msg->msgid;
+        payload_len = msg->payload_len;
+    }
+
+    /* Check if we are interested in this msg_id */
+    if (msg_id != MAVLINK_MSG_ID_REMOTE_LOG_DATA_BLOCK) {
+        return buffer->len;
+    }
+
+    _stat.write.total++;
+    _stat.write.bytes += buffer->len;
+
+    if (_logging_start_timeout) {
+        _remove_start_timeout();
+        if (!_start_alive_timeout()) {
+            log_warning("Could not start liveness timeout - mavlink router log won't be able to detect if flight stack stopped");
+        }
+    }
+
+    uint8_t *payload;
+
+    if (mavlink2) {
+        payload = buffer->data + sizeof(struct mavlink_router_mavlink2_header);
+        trimmed_zeros = get_trimmed_zeros(buffer);
+    } else {
+        payload = buffer->data + sizeof(struct mavlink_router_mavlink1_header);
+        trimmed_zeros = 0;
+    }
+
+    if (trimmed_zeros) {
+        mavlink_remote_log_data_block_t binlog_data;
+        memcpy(&binlog_data, payload, payload_len);
+        memset(((uint8_t *)&binlog_data) + payload_len, 0, trimmed_zeros);
+        _logging_data_process(&binlog_data);
+    } else {
+        mavlink_remote_log_data_block_t *binlog_data = (mavlink_remote_log_data_block_t *)payload;
+        _logging_data_process(binlog_data);
+    }
+
+    return buffer->len;
+}
+
+void BinLog::_send_ack(uint32_t seqno)
+{
+    mavlink_message_t msg;
+
+    // Message filled a gap, or is duplicated. just send the ack
+    if (seqno < _last_acked_seqno) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
+        _send_msg(&msg, _target_system_id);
+        return;
+    }
+
+    // TODO maybe a threshould of [n]acks to be sent?
+    // TODO send ack to source only?
+    // Send nacks regarding unseen seqno
+    for (uint32_t i = _last_acked_seqno; i < seqno; i++) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 0);
+        _send_msg(&msg, _target_system_id);
+    }
+
+    // Send ack to seen seqno
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
+    _send_msg(&msg, _target_system_id);
+    _last_acked_seqno = seqno;
+}
+
+void BinLog::_logging_data_process(mavlink_remote_log_data_block_t *msg)
+{
+    ssize_t r;
+
+    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
+        log_error_errno(errno, "lseek failed (%m)");
+        return;
+    }
+
+    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
+    if (r != MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN) {
+        // partial writes are handled by not sending ack. Flight stack should resend
+        // msg. We hope that partial writes are rare enough
+        log_error_errno(errno, "Error writing data (%m)");
+        return;
+    }
+
+    // TODO should we send acks on a different fashion. e.g. queueing and sending?
+    _send_ack(msg->seqno);
+}
--- a/logendpoint.cpp
+++ b/logendpoint.cpp
@@ -1,0 +1,185 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Copyright (C) 2017  Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "logendpoint.h"
+
+#include <fcntl.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "log.h"
+#include "mainloop.h"
+#include "util.h"
+
+static bool _logging_start_timeout_cb(void *data)
+{
+    LogEndpoint *log = static_cast<LogEndpoint *>(data);
+    return log->_start_timeout();
+}
+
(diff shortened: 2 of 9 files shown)


binlog.cpp
```diff
@@ -0,0 +1,235 @@
...
+void BinLog::_logging_data_process(mavlink_remote_log_data_block_t *msg)
+{
+    ssize_t r;
+
+    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
+        log_error_errno(errno, "lseek failed (%m)");
+        return;
+    }
+
+    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
```
zehortigoza: You should handle partial writes.

binlog.cpp
```diff
@@ -0,0 +1,235 @@
...
+    if (r != MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN) {
+        log_error_errno(errno, "Error writing data (%m)");
+        return;
+    }
+
+    // TODO should we send acks on a different fashion. e.g. queueing and sending?
+    _send_ack(msg->seqno);
```
zehortigoza: If we need do queueing the "Router" aka Mainloop will do it.

binlog.cpp
```diff
@@ -0,0 +1,235 @@
...
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
+        _send_msg(&msg, _target_system_id);
+        return;
+    }
+
+    // TODO maybe a threshould of [n]acks to be sent?
+    // TODO send ack to source only?
+    // Send nacks regarding unseen seqno
+    for (uint32_t i = _last_acked_seqno; i < seqno; i++) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 0);
```
zehortigoza: Why this block?
You send a ack to each msg received. We are missing some messages? If we are I think that we should not send a ACK for those ones.

binlog.cpp
```diff
@@ -0,0 +1,235 @@
...
+    _file = open(filename, O_WRONLY | O_CLOEXEC | O_CREAT | O_NONBLOCK | O_TRUNC,
+                 S_IRUSR | S_IROTH | S_IRGRP);
+    if (_file < 0) {
+        log_error_errno(errno, "Unable to open BinLog file(%s): (%m)", filename);
+        goto open_error;
+    }
+
+    // TODO check why in a timeout
+    // TODO ulog also check for liveness - should we do the same here?
+    _timeout = _mainloop->add_timeout(MSEC_PER_SEC, _binlog_timeout_cb, this);
```
zehortigoza: You never stops this timeout... you should do it as soon you got an ack from MAV_REMOTE_LOG_DATA_BLOCK_START.

Also you should check for liveness as user can reboot the flight stack anytime.

lucasdemarchi:
@fnoop @peterbarker can you take a look?

mainloop.cpp
```diff
@@ -118,7 +118,7 @@ int Mainloop::write_msg(Endpoint *e, const struct buffer *buf)
 
 void Mainloop::route_msg(struct buffer *buf, int target_sysid, int sender_sysid)
 {
-    if (target_sysid > 0) {
+    if (target_sysid > 0 &&  target_sysid < 255) {
```
zehortigoza: 255 is the ID of QGC so it is still a valid id

edersondisouza: Ouch - this was some debug stuff left...

binlog.cpp
edersondisouza: This code sends (or should send) nacks for messages that weren't seen. For instance, we last saw msg `40`, and we see msg `45`, we send nacks for messages `41` to `44`.

binlog.cpp
edersondisouza: ok

binlog.cpp
edersondisouza: Partial writes are naturally handled by the ringbuffer...

peterbarker:
@lucasdemarchi I'll have a squiz after @zehortigoza's comments have been looked at :-)

I might question the use of a ringbuffer to hold the data, however, just based on the commit messages.  It introduces complication.  You're not running on spinning rust on 2017, so seeks aren't painful.  You're probably running on a relatively powerful computer with actual filesystem buffering.  Any buffering you do yourself is logging which will be lost in the event of Significant Unplanned Aircraft Altitude Divergence.  Sometimes naive is better.

edersondisouza:
v2:
  - Much more code is now shared between ulog and binlog implementations; 
  - Fix a bug on mainloop timeout code

edersondisouza:
@peterbarker Sure, the ringbuffer adds some complication to this code - but I tried to keep it as separated as possible. It should also handle partial writes naturally - if we opt to drop ringbuffer, it's a thing that we need to account also.
My worries when I opt for a ringbuffer were not only lseek performance - but the need to ask flight stack again for some data (maybe I should explicit this on commit message). If we have the data, and a partial write happens, we'll still need to buffer what wasn't written somehow, to try again on a next time. We could just ask flight stack to resend data, but I'm not sure if this is a better or less costly approach.
Any thoughts?

peterbarker:
On Tue, 14 Mar 2017, Ederson de Souza wrote:

> the need to ask flight stack again for some data (maybe I should explicit
> this on commit message). If we have the data, and a partial write happens,
> we'll still need to buffer what wasn't written somehow, to try again on a
> next time. We could just ask flight stack to resend data, but I'm not sure

That's what my implementations have done so far :-)  On a modern machine 
one would hope a short write is an exceptional circumstance, so handling 
it was shrug-worthy.  I do emit a message whenever it happens - and I've 
never seen it come out.

Still, sounds like you're well advanced with your ring-buffering, so don't 
let my laziness infect you  :-)

edersondisouza:
@peterbarker Nah, if ringbuffer adds to much complexity for too little gain (or loss, as in Significant Unplanned Aircraft Altitude Divergence [that's a cool name]), I'll just drop it =D

mainloop.cpp
```diff
@@ -515,7 +515,7 @@ void Mainloop::_del_timeouts()
         while (current) {
             if (current->remove_me) {
                 prev->next = current->next;
-                remove_fd(_timeouts->fd);
+                remove_fd(current->fd);
```
zehortigoza: nice catch

binlog.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+int BinLog::write_msg(const struct buffer *buffer)
+{
+    const bool mavlink2 = buffer->data[0] == MAVLINK_STX;
+    uint32_t msg_id;
+    uint16_t payload_len;
+    uint8_t trimmed_zeros;
+
+    _stat.write.total++;
+    _stat.write.bytes += buffer->len;
```
zehortigoza: This block should be after `if (msg_id != MAVLINK_MSG_ID_REMOTE_LOG_DATA_BLOCK) {` or it will never detect that flight stack have stopped sent Log packages because it will receive heartbeat and other messages without target.

binlog.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (_logging_start_timeout) {
+        _remove_start_timeout();
+        if (!_start_alive_timeout()) {
+            log_warning("Could not start liveness timeout - mavlink router log won't be able to detect if flight stack stopped");
+        }
+    }
+
+    uint8_t *payload;
+
+    // TODO do we really need this?
```
zehortigoza: You need, because MAVLINK_MSG_ID_REMOTE_LOG_DATA_BLOCK is compatible with Mavlink 1 and 2.

ringbuffer.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+template<typename T>
+Ringbuffer<T>::~Ringbuffer()
+{
+    free(_buffer);
+}
+
+template<typename T>
+enum Ringbuffer<T>::add_result Ringbuffer<T>::add(T *item, uint32_t seqno)
```
zehortigoza: This is not a regular Ringbuffer implementation it is specialized for BinLog usage so maybe is better have a more specific name.

binlog.cpp
```diff
@@ -154,16 +156,22 @@ void BinLog::_send_ack(uint32_t seqno)
...
+    result = _ringbuffer.add(&msg->data, msg->seqno);
+    switch (result) {
+    case _ringbuffer.overflow:
+        // This is kinda desperate, because when ringbuffer filled, we tried to write
+        // TODO control these writes - maybe stop logging and starting again?
+        log_warning("Buffer overflow, discarding message, writing");
+        _ringbuffer.write_partial(_file);
+        break;
+    case _ringbuffer.full:
+        log_warning("Buffer full, writing");
```
zehortigoza: This is a normal case right? So the log level should be info or debug otherwise it will print a lot of information that is not necessary.

binlog.cpp
edersondisouza: Actually, the handling is done by not sending `ack` to flight stack - it will resend the message. Each message has a fixed position on file, so it will be just overwritten.

zehortigoza: Let say that it is running over a file system that can write 100bytes without blocking and you are trying to write 101bytes, so it will loop here forever.

zehortigoza: Another thing that can happen is: the last write before user power off the board(forced or not) it do a partial write(lets say after several writes the file system now can not write all the bytes without blocking), the BinLog parsers will be able to work with this file?

edersondisouza:
On 15-03-2017 16:10, José Roberto de Souza wrote:
> *@zehortigoza* commented on this pull request.
>
> --------------------------------------------------------------------------------------------
>
> In binlog.cpp <https://github.com/01org/mavlink-router/pull/31#discussion_r106305471>:
>
>> +    // Send ack to seen seqno
> +    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
> +    _send_msg(&msg, _target_system_id);
> +    _last_acked_seqno = seqno;
> +}
> +
> +void BinLog::_logging_data_process(mavlink_remote_log_data_block_t *msg)
> +{
> +    ssize_t r;
> +
> +    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
> +        log_error_errno(errno, "lseek failed (%m)");
> +        return;
> +    }
> +
> +    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
>
> Another thing that can happen is: the last write before user power off the board(forced or
> not) it do a partial write(lets say after several writes the file system now can not write
> all the bytes without blocking), the BinLog parsers will be able to work with this file?

That could happen on any scenario really - even on a buffered one. So, I hope they are able...

peterbarker:
On Wed, 15 Mar 2017, Ederson de Souza wrote:

> That could happen on any scenario really - even on a buffered one. So, I
> hope they are able...

The parsers have to put up with all sorts of garbage.  They are able.

mbelluzzo:
> Any buffering you do yourself is logging which will be lost in the event of **Significant Unplanned Aircraft Altitude Divergence**. Sometimes naive is better.

Is that the official terminology for crashing the drone?

binlog.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "log.h"
+#include "mainloop.h"
+#include "util.h"
+
+struct _packed_ binlog_msg_header {
```
edersondisouza: Fixed

edersondisouza:
@peterbarker Hi, could you take a look? I think that now most of @zehortigoza points have already been dealt with...

binlog.cpp
peterbarker: You can nack things, or you can just wait for the server to resend.  The "spec" is a little loose on this.  Also, there is no "spec".

binlog.cpp
```diff
@@ -0,0 +1,168 @@
...
+{
+    ssize_t r;
+
+    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
+        log_error_errno(errno, "lseek failed (%m)");
+        return;
+    }
+
+    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
+    if (r != MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN) {
```
peterbarker: You're not checking for `0` or  `-1` here, which may indicate a permanent failure to write.

logendpoint.cpp
```diff
@@ -0,0 +1,185 @@
...
+                         timeinfo->tm_year + 1900, timeinfo->tm_mon + 1, timeinfo->tm_mday,
+                         timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec, extension);
+        }
+
+        if (r < 1 || (size_t)r >= sizeof(filename)) {
+            log_error_errno(errno, "Error formatting Log file name: (%m)");
+            return nullptr;
+        }
+
+        if (access(filename, F_OK)) {
```
peterbarker: So you're going to a lot of trouble to get a unique filename.  You still have a TOCTOU issue here.  Instead of `access(...)` you would ususally have a `_get_fh(..)` method and use open(..., O_EXCL) to ensure you have an open filehandle to a file you created.

logendpoint.cpp
```diff
@@ -0,0 +1,185 @@
...
+    else
+        _stat.read.handled_bytes += sizeof(struct mavlink_router_mavlink1_header);
+}
+
+char *LogEndpoint::_get_filename(const char *extension)
+{
+    char filename[PATH_MAX] = { };
+
+    time_t t = time(NULL);
+    struct tm *timeinfo = localtime(&t);
```
peterbarker: This is awesome - if it works.

I initially wrote the dronekit-la implementation doing this, but had to fall back to sequential numbering; the issue is that the real-time-clock on whatever computer you are running this on may not be initialised from an external source yet.  On Solo this was from the GPS passed through the flight control board 0 if you're inside you may never get a time update, and when it comes wackiness ensues.

binlog.cpp
```diff
@@ -0,0 +1,168 @@
...
+    /* Check if we are interested in this msg_id */
+    if (msg_id != MAVLINK_MSG_ID_REMOTE_LOG_DATA_BLOCK) {
+        return buffer->len;
+    }
+
+    _stat.write.total++;
+    _stat.write.bytes += buffer->len;
+
+    if (_logging_start_timeout) {
+        _remove_start_timeout();
```
peterbarker: What if this was an old sessoin that is still alive at the moment?  e.g. if you killed the mavrouter process and instantly restarted it?  I typically NAK any packets I receive when not in the started state - unless they have a seqno of 0, in which case it is time to start.

logendpoint.cpp
edersondisouza: Hmmm.. that's sad. No update at all is fine, mavlink-router would just add an increasing `_x` on filename. But a 'wild time update' could mess everything. 
@zehortigoza as the one who wrote this (I just moved code here), any thoughts?

logendpoint.cpp
edersondisouza: Fair =D

binlog.cpp
edersondisouza: Oh - got it. Thanks!

logendpoint.cpp
lucasdemarchi: @edersondisouza I think we already talked about this, no?

Log names need to start with a number... I think 3 decimal is sufficient and followed by date.  We have to handle some options:

1. user is happy with just this scheme, we just create the file name with this information
2. user would prefer to use the same date as the one from GPS.  In this case we would need to parse the GPS msg and get the date from there. Then we later *rename* the file with the new date.  However we still maintain the same number prefix.

The number prefix doesn't mean we won't be able to handle more than 999 logs... it's just that they will be easily sorted without using numerical sorting. When we have more than that, we just keep increasing the number.

binlog.cpp
edersondisouza: It's tempting to just do nothing, then. Why not? Maybe here is useful to say to flight stack: it's time to resend these, I haven't seen them.

binlog.cpp
edersondisouza: I'm tempted to just ignore such messages... I mean, if I send nack for them, won't the flight stack just resend it again? With same `seqno`?

logendpoint.cpp
edersondisouza: If we talked, I have no recollection of this, sorry. Thank you for (re)explaining =D

binlog.cpp
peterbarker: @edersondisouza Sorry, yes, you should be dropping them.  Just have to rely on the server to get the idea that noone's listening.

binlog.cpp
edersondisouza: Not sure what would be different in the reaction here. When I tackle excessive logging, it may have a different logging strategy, though.

edersondisouza:
New version on #41

Issue #32, opened 2017-03-16
tcpendpoint doesn't retry on lost link


fnoop:
Using the new tcpendpoint when connecting as a client, if the remote/server drops connection (for example if the flight controller reboots or SITL restarts) then mavlink-router loses the connection forever.
Ideally, it should retry periodically to regain the connection.

lucasdemarchi:
A RetryTimeout value in the TcpEndpoint entry in the config file may make sense. Then we default it to 5s or so.

edersondisouza:
@fnoop Once #34 was merged, do you believe that this issue can be closed?

fnoop:
Confirmed fixed now, thanks :)

fnoop:
Oh, except when it does reconnect the mavlink-routerd process goes to 100% cpu

edersondisouza:
Almost there =D 
#57 should fix that!
Thanks for reporting!

edersondisouza:
@fnoop is it all working well now? Can we close this issue?

fnoop:
Yup confirmed fixed, thanks :)

PR #34, opened 2017-03-17
Tcp reconnect

edersondisouza:
#32

### DIFF
--- a/endpoint.h
+++ b/endpoint.h
@@ -145,15 +145,30 @@ protected:
 class TcpEndpoint : public Endpoint {
 public:
     TcpEndpoint();
+    ~TcpEndpoint();
 
     int accept(int listener_fd);
     int open(const char *ip, unsigned long port);
+    void close();
 
     int write_msg(const struct buffer *pbuf) override;
     int flush_pending_msgs() override { return -ENOSYS; }
 
     struct sockaddr_in sockaddr;
+    int retry_timeout = 0;
+
+    inline const char *get_ip() {
+        return _ip;
+    }
+
+    inline unsigned long get_port() {
+        return _port;
+    }
 
 protected:
     ssize_t _read_msg(uint8_t *buf, size_t len) override;
+
+private:
+    char *_ip = nullptr;
+    unsigned long _port = 0;
 };
--- a/endpoint.cpp
+++ b/endpoint.cpp
@@ -526,6 +526,11 @@ TcpEndpoint::TcpEndpoint()
     bzero(&sockaddr, sizeof(sockaddr));
 }
 
+TcpEndpoint::~TcpEndpoint()
+{
+    free(_ip);
+}
+
 int TcpEndpoint::accept(int listener_fd)
 {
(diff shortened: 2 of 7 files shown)


lucasdemarchi:
I don't really like the use of std::function because it doesn't have a constexpr constructor... meaning there will be an allocated object per call. Or am I not up to date on its usage?

edersondisouza:
> I don't really like the use of std::function because it doesn't have a constexpr constructor... meaning there will be an allocated object per call. Or am I not up to date on its usage?

Glorious C++ and infinite different (or new to me) information to consider...

Maybe did you mean that for `std::bind`? We are going to have one `std::function` per `Timeout` - nothing really different from what we do have now.
If you are worried about many `std::bind` calls, them we might consider having it as field of every log object - not sure if worth, as it's not a code path that I would expect to have many runs: two binds for each start.
Or maybe you meant something else...

lucasdemarchi:
Yep, std::bind. Context:

~2 years ago I wanted to use std::function on ardupilot to replace its awful "fastdelegate" implementation. I stopped pursuing that due to: 

- There's no way to convince stdc++ to create objects at compile time. You always need an extra allocation  - this is due to the missing constexpr constructor. In ArduPilot this was really critical because then the object would not stay in flash
- Some benchmarks (on Linux) showed it to be slower and you can find some benchmarks available on the web... This may not be that important in the current use case, but we may need to think if this will continue to be true in future... and it's mind-blowing since the only thing we really want is a callback with a data pointer to be called in future, wth would we need the extra allocation and be so slow?

So... I ended up in a slightly different/simpler functor implementation for ardupilot:
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_HAL/utility/functor.h

I'm not saying to use it. I'm all in favor of user the standard interfaces if they cover our use cases (e.g. we could make more use here of std::string, std::unique_ptr, etc). In order to make such decisions it's nice to take a look on what other projects are doing... I always use chromium/webkit as a reference for c++ when it makes sense. See for example this discussion: https://groups.google.com/a/chromium.org/forum/#!topic/cxx/SoEj7oIDNuA

endpoint.h
```diff
@@ -145,15 +145,30 @@ class UdpEndpoint : public Endpoint {
...
     int flush_pending_msgs() override { return -ENOSYS; }
 
     struct sockaddr_in sockaddr;
+    int retry_timeout = 0;
+
+    inline const char *get_ip() {
+        return _ip;
+    }
+
+    inline unsigned long get_port() {
```
lucasdemarchi: it's implicitly inline when definition is on the header

fnoop:
New parameter should be added to Readme.
So it looks like the tcp endpoint does reconnect if connection lost, but the routing seems to break.  After reconnection I can no longer connect to SITL (the tcp endpoint) through another endpoint.

edersondisouza:
@fnoop that's sad... I'll look into it

edersondisouza:
@fnoop #50 should fix it.

PR #41, opened 2017-03-21
Ardupilot log v3

edersondisouza:
Created new PR as old one is full of comments.

- No more TOCTOU
- New log file structure - starting with increasing XXX-date-time
- Checkpatch compliant

### DIFF
--- a/binlog.cpp
+++ b/binlog.cpp
@@ -1,0 +1,180 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Copyright (C) 2017  Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "binlog.h"
+
+#include <assert.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "log.h"
+#include "mainloop.h"
+#include "util.h"
+
+bool BinLog::_start_timeout()
+{
+    mavlink_message_t msg;
+
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id,
+                                             MAV_COMP_ID_ALL, MAV_REMOTE_LOG_DATA_BLOCK_START, 1);
+    _send_msg(&msg, _target_system_id);
+
+    return true;
+}
+
+bool BinLog::start()
+{
+    if (!LogEndpoint::start()) {
+        return false;
+    }
+
+    _last_acked_seqno = 0;
+    _system_id = SYSTEM_ID;
+
+    return true;
+}
+
+void BinLog::stop()
+{
+    mavlink_message_t msg;
+
+    if (_file == -1) {
+        log_error("BinLog not started");
+        return;
+    }
+
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id,
+                                             MAV_COMP_ID_ALL, MAV_REMOTE_LOG_DATA_BLOCK_STOP, 1);
+    _send_msg(&msg, _target_system_id);
+
+    LogEndpoint::stop();
+}
+
+int BinLog::write_msg(const struct buffer *buffer)
+{
+    const bool mavlink2 = buffer->data[0] == MAVLINK_STX;
+    uint32_t msg_id;
+    uint16_t payload_len;
+    uint8_t trimmed_zeros;
+    mavlink_remote_log_data_block_t *binlog_data;
+
+    if (mavlink2) {
+        struct mavlink_router_mavlink2_header *msg
+            = (struct mavlink_router_mavlink2_header *)buffer->data;
+        msg_id = msg->msgid;
+        payload_len = msg->payload_len;
+    } else {
+        struct mavlink_router_mavlink1_header *msg
+            = (struct mavlink_router_mavlink1_header *)buffer->data;
+        msg_id = msg->msgid;
+        payload_len = msg->payload_len;
+    }
+
+    /* Check if we are interested in this msg_id */
+    if (msg_id != MAVLINK_MSG_ID_REMOTE_LOG_DATA_BLOCK) {
+        return buffer->len;
+    }
+
+    uint8_t *payload;
+
+    if (mavlink2) {
+        payload = buffer->data + sizeof(struct mavlink_router_mavlink2_header);
+        trimmed_zeros = get_trimmed_zeros(buffer);
+    } else {
+        payload = buffer->data + sizeof(struct mavlink_router_mavlink1_header);
+        trimmed_zeros = 0;
+    }
+
+    if (trimmed_zeros) {
+        binlog_data
+            = (mavlink_remote_log_data_block_t *)alloca(sizeof(mavlink_remote_log_data_block_t));
+        memcpy(binlog_data, payload, payload_len);
+        memset(binlog_data + payload_len, 0, trimmed_zeros);
+    } else {
+        binlog_data = (mavlink_remote_log_data_block_t *)payload;
+    }
+
+    if (_logging_start_timeout) {
+        // If message doesn't start a log session, just ignore it
+        if (binlog_data->seqno == 0) {
+            _remove_start_timeout();
+            if (!_start_alive_timeout()) {
+                log_warning("Could not start liveness timeout - mavlink router log won't be able "
+                            "to detect if flight stack stopped");
+            }
+        } else {
+            return buffer->len;
+        }
+    }
+    _logging_data_process(binlog_data);
+
+    _stat.write.total++;
+    _stat.write.bytes += buffer->len;
+
+    return buffer->len;
+}
+
+void BinLog::_send_ack(uint32_t seqno)
+{
+    mavlink_message_t msg;
+
+    // Message filled a gap, or is duplicated. just send the ack
+    if (seqno < _last_acked_seqno) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg,
+                                                 _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
+        _send_msg(&msg, _target_system_id);
+        return;
+    }
+
+    // TODO maybe a threshould of [n]acks to be sent?
+    // TODO send ack to source only?
+    // Send nacks regarding unseen seqno
+    for (uint32_t i = _last_acked_seqno; i < seqno; i++) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg,
+                                                 _target_system_id, MAV_COMP_ID_ALL, seqno, 0);
+        _send_msg(&msg, _target_system_id);
+    }
+
+    // Send ack to seen seqno
+    mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg, _target_system_id,
+                                             MAV_COMP_ID_ALL, seqno, 1);
+    _send_msg(&msg, _target_system_id);
+    _last_acked_seqno = seqno;
+}
+
+void BinLog::_logging_data_process(mavlink_remote_log_data_block_t *msg)
+{
+    ssize_t r;
+
+    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
+        log_error_errno(errno, "lseek failed (%m)");
+        return;
+    }
+
+    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
+    if (r != MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN) {
+        // partial writes are handled by not sending ack. Flight stack should resend
+        // msg. We hope that partial writes are rare enough
+        log_error_errno(errno, "Error writing data (%m)");
+        return;
+    }
+
+    // TODO should we send acks on a different fashion. e.g. queueing and sending?
+    _send_ack(msg->seqno);
+}
--- a/logendpoint.cpp
+++ b/logendpoint.cpp
@@ -1,0 +1,187 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Copyright (C) 2017  Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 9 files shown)


edersondisouza:
@peterbarker @zehortigoza @lucasdemarchi @fnoop Could you please review this new version? Tks!

lucasdemarchi:
@edersondisouza I saw a couple of bugs there but I didn't take note. I'll apply, fix and go through a test round. Thanks.

binlog.cpp
```diff
@@ -0,0 +1,180 @@
...
+        binlog_data
+            = (mavlink_remote_log_data_block_t *)alloca(sizeof(mavlink_remote_log_data_block_t));
+        memcpy(binlog_data, payload, payload_len);
+        memset(binlog_data + payload_len, 0, trimmed_zeros);
+    } else {
+        binlog_data = (mavlink_remote_log_data_block_t *)payload;
+    }
+
+    if (_logging_start_timeout) {
+        // If message doesn't start a log session, just ignore it
```
peterbarker: You could send a stop-message instead of ignoring it.

binlog.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    // Message filled a gap, or is duplicated. just send the ack
+    if (seqno < _last_acked_seqno) {
+        mavlink_msg_remote_log_block_status_pack(_system_id, MAV_COMP_ID_ALL, &msg,
+                                                 _target_system_id, MAV_COMP_ID_ALL, seqno, 1);
+        _send_msg(&msg, _target_system_id);
+        return;
+    }
+
+    // TODO maybe a threshould of [n]acks to be sent?
```
peterbarker: This won't nack for packets in a "gap" e.g. you have packets 1 and 3, and have acked both of them, but you're missing packet 2.

binlog.cpp
```diff
@@ -0,0 +1,180 @@
...
+    _send_msg(&msg, _target_system_id);
+    _last_acked_seqno = seqno;
+}
+
+void BinLog::_logging_data_process(mavlink_remote_log_data_block_t *msg)
+{
+    ssize_t r;
+
+    if (lseek(_file, msg->seqno * MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN, SEEK_SET) < 0) {
+        log_error_errno(errno, "lseek failed (%m)");
```
peterbarker: This probably should be something a little more serious than logging an error.  e.g. closing the session off and trying on a fresh file.

binlog.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return;
+    }
+
+    r = write(_file, msg->data, MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN);
+    if (r != MAVLINK_MSG_REMOTE_LOG_DATA_BLOCK_FIELD_DATA_LEN) {
+        // partial writes are handled by not sending ack. Flight stack should resend
```
peterbarker: if `r == -1` then you should do something more serious than logging an error.  There's a good chance that that log message will be lost because you've just filled the disk up!

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    for (; i <= MAX_LOG_FILE_PREFIX; i++) {
+        r = snprintf(filename, sizeof(filename), "%s/%03u-%i-%02i-%02i_%02i-%02i-%02i.%s",
+                     _logs_dir, i, timeinfo->tm_year + 1900, timeinfo->tm_mon + 1,
+                     timeinfo->tm_mday, timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec,
+                     extension);
+
+        if (r < 1 || (size_t)r >= sizeof(filename)) {
+            log_error_errno(errno, "Error formatting Log file name: (%m)");
+            return -EINVAL;
```
peterbarker: You're inconsistent in return values from this function (some places are returning -EPERM...).  Return -1 and rely on errno, I suggest.

mainloop.cpp
```diff
@@ -390,8 +390,12 @@ bool Mainloop::add_endpoints(Mainloop &mainloop, struct opt *opt)
         g_tcp_fd = tcp_open(opt->tcp_port);
 
     if (opt->logs_dir) {
-        _ulog = new ULog(opt->logs_dir);
-        g_endpoints[i] = _ulog;
+        if (streq(DIALECT, "common")) {
+            _log_endpoint = new ULog(opt->logs_dir);
+        } else {
```
peterbarker: If I'm guessing correctly, you should check for the `ardupilotmega` dialect here.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "log.h"
+#include "mainloop.h"
+#include "util.h"
+
+#define MAX_LOG_FILE_PREFIX 999
```
peterbarker: I wonder if this is really a long-enough prefix.  I went with 8 digits for the dataflash-file backend in ArduPilot recently!  Admittedly, that's overkill.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+        0, data
+    };
+
+    buffer.len = mavlink_msg_to_send_buffer(data, msg);
+    _mainloop->route_msg(&buffer, target_sysid, msg->sysid);
+
+    _stat.read.total++;
+    _stat.read.handled++;
+    _stat.read.handled_bytes += msg->len;
+    if (msg->magic == MAVLINK_STX)
```
peterbarker: I'm still not sure about this handle-mavlink-2-differently stuff.

But you should check for MAVLINK1_STX indicating a mavlink1 packet, rather than assuming MAVLINK_STX is mavlink2.  There are several places that do this.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+int LogEndpoint::_get_prefix(const char *extension, const char *dirname, uint16_t *prefix)
+{
+    struct dirent *ent;
+    DIR *dir;
+    uint16_t u;
+
+    dir = opendir(dirname);
+    if (!dir) {
+        return -errno;
```
peterbarker: Caller only checks for negative, and looks at errno; should return -1 and rely on errno

Also, some comments at the top of a function like, "returns -1 on failure" are nice.  Comments in general are nice, actually....

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    closedir(dir);
+
+    return 0;
+}
+
+int LogEndpoint::_get_file(const char *extension, char **filename_result)
+{
+    char filename[PATH_MAX] = {};
```
peterbarker: You could leave allocation to the caller rather than doing a "caller must free" in this function.  Passing in a pointer/pointer-len pair....

mainloop.cpp
edersondisouza: Currently, build can only be done with `common` or `ardupilotmega` - so it should be safe - for now, at least.

logendpoint.cpp
edersondisouza: Oh, will check to keep consistency, thanks. Relying on `errno` and having logs at the same time can be cumbersome imho (you have to save `errno` before logging and so on...)

binlog.cpp
edersondisouza: Fair

binlog.cpp
lucasdemarchi: Since the fd is opened as nonblock, just make sure to handle (r == -1 && erno == EAGAIN) the same way you handle partial writes. Other errors situations can be handled as @peterbarker said.

logendpoint.cpp
lucasdemarchi: The wrong thing in this implementation is not allowing to go beyond that value.  A prefix of 3 would be ok if that wouldn't be limiting to 1000 log files.

@edersondisouza let's increase this to 5 and always allow to go beyond that number... in fact you only have to remove code to allow that.

lucasdemarchi: I mean... 5 decimal places

logendpoint.cpp
lucasdemarchi: mavlink1 vs mavlink2 is seemless if you use mavlink library functions.

The intention was not to use it in most of the places, and use the wire format directly on the core router part. Then this ended up being spread across different places, which is in need of a cleanup patch.

@edersondisouza in this particular case, why are you not using `buffer.len` since you only want the size of the msg including header, which should be already in this variable.

logendpoint.cpp
lucasdemarchi: Actually the purpose of log_error_errno() to take errno as argument is exactly so you can ger errno back. IMO would be ok to just `return log_error_errno(...)` and comment that it returns negative on error.  If the caller needs to use the errno then it should also be the one responsible for logging the error and log_* shouldn't be called in this function.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+                     _logs_dir, i, timeinfo->tm_year + 1900, timeinfo->tm_mon + 1,
+                     timeinfo->tm_mday, timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec,
+                     extension);
+
+        if (r < 1 || (size_t)r >= sizeof(filename)) {
+            log_error_errno(errno, "Error formatting Log file name: (%m)");
+            return -EINVAL;
+        }
+
+        r = open(filename, O_WRONLY | O_CLOEXEC | O_CREAT | O_NONBLOCK | O_TRUNC,
```
lucasdemarchi: you are missing `O_EXCL` here. You don't want O_TRUNC, i.e. to overwrite files... you want to fail if the file already exists

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+                 S_IRUSR | S_IROTH | S_IRGRP);
```
lucasdemarchi: IMO 0644 would be clearer and the correct permissions for this file.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+        if (r >= 0) {
```
lucasdemarchi: return early on error, not on success.... the only exception is if it makes way simpler to handle the success case before, which is not in this case.

btw the for above should be a fixed number of retries, not the number available in the prefix range

logendpoint.h
```diff
@@ -0,0 +1,62 @@
...
+#define SYSTEM_ID 2
+
+class LogEndpoint : public Endpoint {
+public:
+    LogEndpoint(const char *logs_dir)
+        : Endpoint{"BinLog", false}
+        , _logs_dir{logs_dir}
+    {
+        assert(_logs_dir);
+    };
```
lucasdemarchi: dangling ;

logendpoint.h
```diff
@@ -0,0 +1,62 @@
...
+#pragma once
+
+#include <assert.h>
+
+#include "endpoint.h"
+#include "timeout.h"
+
+#define ALIVE_TIMEOUT 5
+#define TARGET_SYSTEM_ID 1
+#define SYSTEM_ID 2
```
lucasdemarchi: these defines shouldn't be in the header if you don't really need them here... move to .cpp.

If you need, you should either:
1) add appropriate "namespace", e.g. LOGENDPOINT_SYSTEM_ID
or
2) move it to be public `static const unsigned int` inside LogEndpoint

logendpoint.cpp
edersondisouza: LOL - i must've seen `O_CLOEXEC` and thought about `O_EXECL`... tks!

lucasdemarchi: but don't remove O_CLOEXEC ;-)

logendpoint.h
```diff
@@ -0,0 +1,62 @@
...
+        , _logs_dir{logs_dir}
+    {
+        assert(_logs_dir);
+    };
+
+    virtual bool start();
+    virtual void stop();
+
+    virtual bool _start_timeout() = 0;
+    virtual bool _alive_timeout();
```
lucasdemarchi: methods with _ are supposed to be private or protected.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+    _stat.read.total++;
+    _stat.read.handled++;
+    _stat.read.handled_bytes += msg->len;
+    if (msg->magic == MAVLINK_STX)
+        _stat.read.handled_bytes += sizeof(struct mavlink_router_mavlink2_header);
+    else
+        _stat.read.handled_bytes += sizeof(struct mavlink_router_mavlink1_header);
+}
+
+int LogEndpoint::_get_prefix(const char *extension, const char *dirname, uint16_t *prefix)
```
lucasdemarchi: this method doesn't need/use extension. You could actually return int with `< 0` being the special case of errors.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+int LogEndpoint::_get_file(const char *extension, char **filename_result)
+{
+    char filename[PATH_MAX] = {};
+
+    time_t t = time(NULL);
+    struct tm *timeinfo = localtime(&t);
+    int r;
+    uint16_t i;
+
+    if (_get_prefix(extension, _logs_dir, &i) < 0) {
```
lucasdemarchi: 1. open dir on this function and pass it to _get_prefix()
2. use opendirat() below
3. closedir on this function

this makes the _get_prefix simpler and removes the first %s of the snprintf() below.

logendpoint.cpp
```diff
@@ -0,0 +1,213 @@
...
+        log_error_errno(errno, "Error formatting Log file name: (%m)");
+    }
+
+    for (; i <= MAX_LOG_FILE_PREFIX; i++) {
+        r = snprintf(filename, sizeof(filename), "%s/%03u-%i-%02i-%02i_%02i-%02i-%02i.%s",
+                     _logs_dir, i, timeinfo->tm_year + 1900, timeinfo->tm_mon + 1,
+                     timeinfo->tm_mday, timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec,
+                     extension);
+
+        if (r < 1 || (size_t)r >= sizeof(filename)) {
```
lucasdemarchi: PATH_MAX may not be sufficient depending on mount point... not critical here since it's unlikely to fail, but something to keep in mind.

lucasdemarchi:
@edersondisouza I think I commented on everything I had seen and was planning to fix myself. Since @peterbarker added more comments, if left everything for you ;-)

logendpoint.cpp
edersondisouza: Ah, sure =D

logendpoint.cpp
edersondisouza: I haven't noticed this property of `log_error_errno`... cool!

logendpoint.h
edersondisouza: Yes, they are supposed. They are like this however because they were accessed from a free function callback (timeout callback). Since my patch adding support to `std::function` was merged, this can board the same train now =D

lucasdemarchi: Yep. Although it was not an excuse to make the method public. They could very well be accessed by the static function it the static function was part of the class ;)

edersondisouza: Rá!

logendpoint.cpp
edersondisouza: At the cost of making this function having to care about closing the dir... but I'll look into it.

lucasdemarchi: well... the other function already needed to do it, so it's basically moving the code.

binlog.cpp
edersondisouza: You mean, I have received packets 1 and 3, packet 5 arrives, I'll only send nack for packet 4?
If so, that's true, but a nack for packet 2 was already sent. How important is to keep resending nacks?

logendpoint.cpp
edersondisouza: @lucasdemarchi I'm ok to change to just use `buffer.len`, but a quick comparison shows it to be two bytes bigger than current calculation - is it any problem for static gathering?

lucasdemarchi: humn... can you take a look on the implementation why exactly it's different? From what I see, it shouldn't.

edersondisouza: Basically, because currently implementation appears wrong =D
It follows comments https://github.com/01org/mavlink-router/blob/master/endpoint.h#L27-L54 on other places that get stats, but not here.

edersondisouza:
v4 on #48

peterbarker:
On Tue, 28 Mar 2017, Ederson de Souza wrote:

> You mean, I have received packets 1 and 3, packet 5 arrives, I'll only send
> nack for packet 4?

Yes.

> If so, that's true, but a nack for packet 2 was already sent. How important
> is to keep resending nacks?

It isn't; ArduPilot send you a packet until you ack it :-)

PR #48, opened 2017-03-28
Ardupilot log v4

edersondisouza:
- Should fix issues noted on v3, mainly:
  - Restarting log after failure to write log file
  - Some code paths simplified
  - Sending `stop` packet if received messages while waiting for log to start
  - Statistics collection was underestimating bytes on log

### DIFF
--- a/logendpoint.cpp
+++ b/logendpoint.cpp
@@ -17,6 +17,7 @@
  */
 #include "logendpoint.h"
 
+#include <dirent.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <stdio.h>
@@ -27,11 +28,14 @@
 #include <time.h>
 #include <unistd.h>
 
+#include <memory>
+
 #include "log.h"
 #include "mainloop.h"
 #include "util.h"
 
 #define ALIVE_TIMEOUT 5
+#define MAX_RETRIES 10
 
 void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)
 {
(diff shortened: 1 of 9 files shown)


edersondisouza:
@peterbarker @zehortigoza @lucasdemarchi @fnoop Could you please review this new version? Tks again!

lucasdemarchi:
Failing CI.  You probably need some ifdef between ulog and binlog?

lucasdemarchi:
Or... switch do ardupilot dialect by default since it contains all messages from common.xml

logendpoint.cpp
```diff
@@ -48,44 +52,62 @@ void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)
...
+    uint32_t i;
+    int j, r;
+    DIR *dir;
+
+    dir = opendir(_logs_dir);
+    if (!dir) {
+        return log_error_errno(errno, "Could not open log dir (%m)");
+    }
+    // Close dir when leaving function.
+    std::shared_ptr<void> defer(dir, [](auto p) { closedir(p); });
```
lucasdemarchi: nice c++-ization of the cleanup attribute :). We may want to make it generic later

lucasdemarchi: not really important but it could be a std::unique_ptr, no?

lucasdemarchi:
Applied the 3 first patches, fixing the commit message for the third one.

As for the CI build break, please take a look on other programs that support multiple mavlink dialects to check what they do - would be nice to have this as a runtime configuration rather than compile time. QGC would be a good example I think.

logendpoint.cpp
edersondisouza: It could, but construction would be a bit different, after some tries I got this working:

```c++
std::unique_ptr<DIR, std::function<void(DIR*)>> defer(dir, [](auto p) { closedir(p); });
```
Worth changing? This one is more specific than using `shared_ptr`. We could create a new generic type to solve this...

edersondisouza:
v5 #56

PR #50, opened 2017-03-29
tcp endpoint: After reconnection, mavlink-router wasn't listening it

edersondisouza:
Missing addition of file descriptor to epoll. Fixed - now, every time
an endpoint is added to tcp endpoint list, it will automatically enroll
file descriptor on epoll.

Also, RetryTimeout property documented on README.

### DIFF
--- a/mainloop.cpp
+++ b/mainloop.cpp
@@ -229,6 +229,8 @@ int Mainloop::_add_tcp_endpoint(TcpEndpoint *tcp)
     tcp_entry->endpoint = tcp;
     g_tcp_endpoints = tcp_entry;
 
+    add_fd(tcp->fd, tcp, EPOLLIN);
+
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


mainloop.cpp
```diff
@@ -384,12 +384,10 @@ bool Mainloop::add_endpoints(Mainloop &mainloop, struct opt *opt)
                 continue;
             }
 
-            if (_add_tcp_endpoint(tcp.get()) < 0) {
+            if (_add_tcp_endpoint(tcp.release()) < 0) {
```
lucasdemarchi: this means you transferred ownership to _add_tcp_endpoint() and it is the new owner of TcpEndpoint. Which includes freeing it in the error path, which you are not doing.

lucasdemarchi: I'd say to leave the get() here and the release() below. Then everything is correct again.

edersondisouza:
Update to address comments

PR #52, opened 2017-04-01
conf file: Fix log dir usage

edersondisouza:
When getting log dir from any conf file, mavlink-router was actually
not using it - it was being forgot on stack.


edersondisouza:
Hold on - has a bug.

edersondisouza:
Bug fixed - can be reviewed.

Issue #58, opened 2017-04-05
Create flightstack log dir


edersondisouza:
If flightstack log directory, informed - either via `-l` or `Log` config, doesn't exist, mavlink-router should create it. From https://github.com/intel-aero/meta-intel-aero/pull/74:

> See https://github.com/lucasdemarchi/kmod/blob/master/shared/util.c#L397

edersondisouza:
Done by #59

PR #59, opened 2017-04-05
log: Create log directory if it doesn't exist

edersondisouza:
#58


edersondisouza:
Update commit message.

PR #60, opened 2017-04-06
conf file: Explain all possible configurations

edersondisouza:
Added a example of a config file, with a header that explain in details
all configurations.

Readme changed to point to this file.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -1,0 +1,115 @@
+# mavlink-router configuration file is composed of sections,
+# each section have some keys and values. They
+# are case insensitive, so `Key=Value` is the same as `key=value`.
+#
+# [This-is-a-section name-of-section]
+# ThisIsAKey=ThisIsAValuye
+#
+# Following specifications of expected sessions and key/values.
+#
+# Section [General]: This section doesn't have a name.
+#
+# Keys:
+#   TcpServerPort
+#       A numeric value defining in which port mavlink-router will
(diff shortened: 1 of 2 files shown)


examples/config.sample
```diff
@@ -0,0 +1,115 @@
...
+# Keys:
+#   Device
+#       Path to UART device, like `/dev/ttyS0`
+#       No default value. Must be defined.
+#
+#   Baud
+#       Numeric value stating baudrate of UART device
+#       Default: 115200
+#
+# Section [UdpEndpoint]: This section must have a name
```
anselmolsm: I'm just trying to make it clearer - What about something like:
Section UdpEndpoint: This section must have a name - [UdpEndpoint mandatory-section-name]

anselmolsm: OK, reading the rest made the format clear.

examples/config.sample
```diff
@@ -0,0 +1,115 @@
...
+#       One of <common> or <ardupilotmega>. Defines which MAVLink dialect
+#       is being used by flight stack, so mavlink-router can log
+#       appropiately.
+#       Default: common
+#
+#   Log
+#       Path to directory where to store flightstack log.
+#       No default value. If absent, no flightstack log will be stored.
+#
+#   LogDebugLevel
```
lucasdemarchi: this should be changed to DebugLogLevel before release

PR #63, opened 2017-04-06
log: Only log different sysid message on 'trusted' messages

edersondisouza:
Messages with unknown message id _may_ be corrupted, so we don't trust
its sysid. However, we were logging a possible incosistency based on
such messages.


zehortigoza:
LGTM

edersondisouza:
Merged, tks!

PR #65, opened 2017-04-07
Change ConfFile API

otaviobp:
Followed suggestions on https://github.com/01org/camera-streaming-daemon/pull/29

### DIFF
--- a/conf_file.cpp
+++ b/conf_file.cpp
@@ -16,268 +16,509 @@
  * limitations under the License.
  */
 
-#include "conf_file.h"
-
 #include <assert.h>
 #include <ctype.h>
-#include <stdio.h>
-#include <stdlib.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <fnmatch.h>
+#include <limits.h>
 #include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
 
+#include "conf_file.h"
 #include "log.h"
 #include "util.h"
 
+#define MAX_SECTION_NAME 100
+
 struct config {
-    char *key;
-    char *value;
+    const char *key;
+    const char *value;
+    size_t key_len, value_len;
+    const char *filename;
+    int line;
     struct config *next;
 };
 
 struct section {
-    char *name;
+    const char *name;
+    size_t len;
     struct config *configs;
-    struct config *current_config;
     struct section *next;
+};
+
+struct conffile {
+    int fd;
+    void *addr;
+    size_t len;
+    char *filename;
+    struct conffile *next;
 };
 
 ConfFile::~ConfFile()
 {
-    struct section *section;
-    struct config *config;
-
-    section = _sections;
-    while (section) {
-        struct section *tmp = section;
-
-        config = section->configs;
-        while (config) {
-            struct config *tmp_config = config;
-            free(config->key);
-            // config->value is on the same allocation as config->key, no need to free it
-            config = config->next;
-            free(tmp_config);
-        }
-
-        free(section->name);
-        section = section->next;
-        free(tmp);
-    }
+    release_all();
 }
 
-int ConfFile::parse_file()
+int ConfFile::parse(const char *filename)
 {
-    FILE *file;
-    char *entry = NULL;
-    int line = 0, ret = 0;
-    size_t n = 0;
-    ssize_t read;
+    int fd, ret = 0;
+    void *addr;
+    struct stat fstat;
 
-    assert(_filename);
+    assert(filename);
 
-    file = fopen(_filename, "re");
-    if (!file) {
-        log_error_errno(errno, "Could not open conf file '%s' (%m)", _filename);
-        return -EIO;
+    fd = open(filename, O_RDONLY | O_CLOEXEC);
+    if (fd < 0) {
+        log_error_errno(errno, "Could not open conf file '%s' (%m)", filename);
+        return -errno;
     }
 
-    while ((read = getline(&entry, &n, file)) >= 0) {
-        line++;
-
-        _trim(entry);
-
-        switch (entry[0]) {
-        case ';':
-        case '#':
-        case '\0':
-            // Discards comment or blank line
-            free(entry);
-            break;
-        case '[':
-            ret = _add_section(entry, line);
-            if (ret < 0) {
-                goto end;
-            }
-            break;
-        default:
-            ret = _add_config(entry, line);
-            if (ret < 0) {
-                goto end;
-            }
-        }
-
-        entry = NULL;
+    if (stat(filename, &fstat) < 0) {
+        ret = -errno;
+        goto error;
     }
 
-    _current_section = nullptr;
-
-    if (!_sections || !_sections->configs) {
-        free(entry);
-        log_error("Invalid conf file %s", _filename);
-        return -EINVAL;
+    addr = mmap(0, (size_t)fstat.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (addr == MAP_FAILED) {
+        ret = -errno;
+        goto error;
     }
 
-end:
-    free(entry);
-    fclose(file);
+    _files = new conffile{fd, addr, (size_t)fstat.st_size, strdup(filename), _files};
+
+    ret = _parse_file((char *)addr, (size_t)fstat.st_size, _files->filename);
+    if (ret < 0) {
+        log_error("Conf file parcially parsed. Configuration keys defined before the error were "
+                  "included");
+        return ret;
+    }
+
+    return 0;
+error:
+    close(fd);
     return ret;
 }
 
-const char *ConfFile::first_section()
+int ConfFile::_parse_file(const char *addr, size_t len, const char *filename)
 {
-    _current_section = _sections;
-    return _current_section->name;
-}
+    int line = 0, ret = 0;
+    char *end;
+    size_t line_len, full_line_len;
 
-const char *ConfFile::next_section()
-{
-    if (!_current_section->next)
-        return NULL;
+    struct section *s = nullptr;
+    while ((end = (char *)memchr((void *)addr, '\n', len))) {
+        line++;
+        full_line_len = line_len = end - addr;
 
-    _current_section = _current_section->next;
-    return _current_section->name;
-}
+        // Make sure all casts to int are safe. No need to support options or sections
+        // larger than INT_MAX
+        if (line_len > INT_MAX)
+            return -EINVAL;
+        _trim(&addr, &line_len);
+        if (!line_len)
+            goto next;
 
-int ConfFile::_add_section(char *entry, int line)
-{
-    struct section *section;
+        switch (addr[0]) {
+        case ';':
+        case '#':
+            goto next;
+        case '[':
+            s = _add_section(addr, line_len, line, filename);
+            if (!s)
+                return -EINVAL;
+            break;
+        default:
+            if (!s) {
+                log_error(
+                    "On file %s: Line %d: Expected section before the definition of first option.",
+                    filename, line);
+                return -EINVAL;
+            }
 
-    char *end = strchr(entry, ']');
-    if (!end) {
-        log_error("On file %s: Line %d: Unfinished section name. Expected ']'", _filename, line);
-        return -EINVAL;
-    }
-
-    if (*(end + 1) != '\0') {
-        log_error("On file %s: Line %d: Unexpected characters after session name", _filename, line);
-        return -EINVAL;
-    }
-
-    // So section name is the string between '[]'
-    *entry = ' ';
-    *end = '\0';
-    _trim(entry);
-
-    // Ensure section is not duplicated. If it is,
-    // just make parser add on known section
-    section = _sections;
-    while (section) {
-        if (strcaseeq(section->name, entry)) {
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
First 4 patches applied

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+    }
 
-    _current_section = _current_section->next;
-    return _current_section->name;
+    s = _find_section(addr + 1, len - 2);
+    if (s)
+        return s;
+
+    log_debug("ConfFile: Adding section '%.*s'", (int)(len - 2), addr + 1);
+    _sections = new section{addr + 1, len - 2, nullptr, _sections};
```
edersondisouza: Nice thing this `_list = new item{..., _list}`

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+    if (*ptr == NULL)
+        s = _sections;
+    else
+        s = ((section *)*ptr)->next;
+
+    for (; s; s = s->next) {
+        if (s->len - 1 > section_name_len && strncaseeq(section_name, s->name, section_name_len)
+            && s->name[section_name_len] == separator) {
+            *sub_name = s->name + section_name_len + 1;
+            *sub_len = s->len - (*sub_name - s->name);
```
edersondisouza: This ends up allowing 'unnamed' sub sections - is this desired?

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+    assert(storage_len);
+    assert(val_len);
 
-    *end = '\0';
+    if (storage_len - 1 < val_len)
+        return -ENOBUFS;
 
-    memmove(str, s, end - s + 1);
+    memcpy(storage, val, val_len);
+    ((char *)storage)[val_len] = 0;
```
edersondisouza: Minor, but `\0` is more clear about the intent of this `0`.

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+    struct config *c;
+    int ret;
+    size_t i;
+    void *storage;
+
+    for (i = 0; i < table_len; i++) {
+        c = _find_config(s, table[i].key, strlen(table[i].key));
+        if (!c) {
+            if (table[i].required) {
+                log_error("Required field '%s' not found in section '%.*s'", table[i].key,
```
edersondisouza: Maybe inform last file on which this section was found? Not sure if showing the whole list if files in which this section appears is easy, but could be a plus, don't you think?

main.cpp
```diff
@@ -474,212 +460,174 @@ static const char *get_conf_dir()
     return DEFAULT_CONF_DIR;
 }
 
-static struct endpoint_config *search_endpoints(const char *name)
+static int parse_mavlink_dialect(const char *val, size_t val_len, void *storage, size_t storage_len)
```
edersondisouza: World would be a better place if there were a easy way to generate an enum from string...

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+    void *addr;
+    struct stat fstat;
+
+    assert(filename);
+
+    fd = open(filename, O_RDONLY | O_CLOEXEC);
+    if (fd < 0)
+        return -errno;
+
+    if (stat(filename, &fstat) < 0)
```
edersondisouza: Hint: this trick can be handy https://github.com/01org/mavlink-router/blob/master/logendpoint.cpp#L101-L102

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
-    return _current_section->name;
-}
+    const char *end;
+    struct section *s;
 
-const char *ConfFile::next_section()
-{
-    if (!_current_section->next)
-        return NULL;
+    end = (char *)memrchr(addr, ']', len);
```
edersondisouza: Hmmm... `memrchr` allows things like `[really ] weird] option]`. Is it desirable?

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
-    file = fopen(_filename, "re");
-    if (!file) {
-        log_error_errno(errno, "Could not open conf file '%s' (%m)", _filename);
-        return -EIO;
+    _files = new conffile{fd, addr, (size_t)fstat.st_size, _files};
+
+    ret = _parse_file((char *)addr, (size_t)fstat.st_size, filename);
+    if (ret < 0) {
+        log_error("Conf file parcially parsed. Configuration keys defined before the error were "
+                  "included");
```
otaviobp: I am just logging what has happened here and I am returning an error code.
I decided to print this because if you are parsing multiple files and the last one fails, some data from last file will be added anyway. I don't want people to assume that only data from previous files that were successfully parsed were added.

conf_file.cpp
```diff
@@ -16,268 +16,416 @@
...
+            if (ret < 0)
+                return ret;
+        }
 
-    if (!_sections || !_sections->configs) {
-        free(entry);
-        log_error("Invalid conf file %s", _filename);
-        return -EINVAL;
+    next:
+        len = len - (end - addr) - 1;
```
otaviobp: That's true. I need to subtract the original line len before trim. thanks

conf_file.cpp
otaviobp: We needed to keep a pointer to last file the section appeared. It is not hard, just some pointers added. A list to all files that included the section is a bit more complicated, but easy anyway.
If you think that this is a needed feature, I can added it. But I don't know if it is worth keeping it. Up to you.

conf_file.cpp
otaviobp: We don't allow unnamed sub sections because the if is checking if we have 2 chars (separator + 1) after the section name.
But, it is not trimming the subsection and section names, so we may have a subsection called " ".
So, do we need to trim section names, like:
[     section_name        sub_section_name       ]

otaviobp: github removed my trailing spaces. heheh.  Now with spaces:

    [      section_name             sub_section_name           ]

main.cpp
otaviobp: I though about creating another struct to map this, but I didn't work on that. We could have this, but we would need a user_data pointer to be used in parser functions, with the mapping. Not sure if it is worth it.

conf_file.cpp
otaviobp: No. it is not. my bad.

otaviobp:
Branched updated. All proposed fixes that I said that I would fix, or that were not commented were fixed.
I added some documentation on how to use ConfFile too.
And I removed the global_instance() static function because we won't need to use ConfFile as singleton in mavlink-router nor camera-streaming-daemon and it makes more sense to have it as a regular class.

conf_file.cpp
```diff
@@ -16,268 +16,412 @@
...
-    _sections = section;
+    _trim(&key, &key_len);
+    _trim(&value, &value_len);
 
-    _current_section = section;
+    c = _find_config(s, key, key_len);
+    if (c) {
+        c->value = value;
+        c->value_len = value_len;
+    } else
```
edersondisouza: Unbalanced braces should be avoided on mavlink-router - couldn't find a way to check for them on checkpatch..

conf_file.cpp
edersondisouza: Yeah, I've had an error here and had to grep to find which file caused the error. It's more of a nice to have... not sure if someone would have a long chain of overriding configs...

conf_file.cpp
edersondisouza: Well.... we already trim everywhere =D

confs/b.conf
```diff
@@ -0,0 +1,102 @@
+# mavlink-router configuration file is composed of sections,
```
edersondisouza: A wild conf file appeared!!

conf_file.cpp
```diff
@@ -16,268 +16,412 @@
...
             break;
         default:
-            ret = _add_config(entry, line);
-            if (ret < 0) {
-                goto end;
+            if (!s) {
+                log_error(
+                    "On file %s: Line %d: Expected section before the definition of first option.",
+                    filename, line);
             }
```
edersondisouza: Missing return or break here?

edersondisouza:
Another issue I've found: if I run mavlink-router without pointing to any config file, it just exits, no error, no nothing.

otaviobp:
Pushed new branch with all proposed fixes

main.cpp
```diff
@@ -474,212 +460,174 @@ static const char *get_conf_dir()
...
-                    return -EINVAL;
-                }
-            } else {
-                eavesdropping = endpoint->eavesdropping;
-            }
+    bool *eavesdropping = (bool *)storage;
+    if (memcaseeq(val, val_len, "normal", sizeof("normal") - 1)) {
+        *eavesdropping = false;
+    } else if (memcaseeq(val, val_len, "eavesdropping", sizeof("eavesdropping") - 1)) {
+        *eavesdropping = false;
```
edersondisouza: Guess here you wanted `true`

conf_file.h
```diff
@@ -17,24 +17,123 @@
  */
 #pragma once
 
+#define DECLARE_PARSE_INT(_type) \
+    static int parse_##_type(const char *val, size_t val_len, void *storage, size_t storage_len)
```
lucasdemarchi: this is only useful for "internal" usage by the class. You don't need this defined for who includes this header. As such, a better approach would be to move the define inside the class and after it's used you undef it.

conf_file.h
```diff
@@ -17,24 +17,123 @@
...
+     * @param sub_len A pointer to size_t element be filled with the length of the current
+     * sub-section name.
+     * @param table Array of TableOption with information on which fields are going to be
+     * extracted and how to extract them.
+     * @param table_len The number of elements in @a table.
+     * @param data A pointer to the struct that will be used to hold the extracted data.
+     * @param ptr A void pointer to be used internally in the loop. Must be initialized with @c NULL
+     * for first execution.
+     */
+    int loop_sub_sections(const char *section_name, char separator, const char **sub_name,
```
lucasdemarchi: separator will always be a space, no?

conf_file.h
```diff
@@ -17,24 +17,123 @@
...
+                          size_t *sub_len, const OptionsTable table[], size_t table_len, void *data,
+                          void **ptr);
```
lucasdemarchi: I'm not sure about the approach on this function. I'd rather try to reuse extract_options() above, which would make most of these arguments go away.  The idea would be to have a `get_sections(const char *glob)` that accepts a glob and returns an iterator. You then loop on this iterator and call `extract_options()`. The "sub section" name is saved on this iterator.

I'd also avoid the "sub section" nomenclature. It's just a section IMO.

lucasdemarchi: Maybe a change like this is very intrusive. If we agree on this, we can do it on top if you prefer.

conf_file.h
```diff
@@ -17,24 +17,123 @@
...
+                    int line);
+    void _trim(const char **str, size_t *len);
+    int _extract_options_from_section(struct section *s, const OptionsTable table[],
+                                      size_t table_len, void *data);
 };
+
+/*
+ * Return the number of elements in a OptionTable array.
+ */
+#define OPTIONS_TABLE_SIZE(_table) (sizeof(_table) / sizeof(ConfFile::OptionsTable))
```
lucasdemarchi: this is ARRAY_SIZE, no need to define it with another name.

conf_file.h
```diff
@@ -17,24 +17,123 @@
...
+
+/*
+ * Create the last two elements in OptionsTable, storage_offset and storage_len, using _field
```
lucasdemarchi: "last two" is a good candidate for a forgotten comment when the table is changed

conf_file.h
```diff
@@ -17,24 +17,123 @@
...
+ * from _struct.
+ */
+#define OPTIONS_TABLE_STRUCT_FIELD(_struct, _field) \
+    offsetof(_struct, _field), sizeof(_struct::_field)
```
lucasdemarchi: I think it's better to define OptionsTable with sub-struct containing offset and size. Then this would become:
```c++
#define OPTIONS_TABLE_STRUCT_FIELD(_struct, _field) \
     { offsetof(_struct, _field), sizeof(_struct::_field) }
```

This way it's easier to check that the number of fields initialized in the struct match the number of fields it should be.

main.cpp
lucasdemarchi: > World would be a better place if there were a easy way to generate an enum from string...

well... this actually exists in several forms. P99 macros is one of them:
http://p99.gforge.inria.fr/p99-html/group__types_ga5d811e46f48e01d79d1f931817d9c0f1.html#ga5d811e46f48e01d79d1f931817d9c0f1

Other projects like systemd use code generation for similar thing.

lucasdemarchi: and you will find solutions around using "modern c++XX" with some limitations.

edersondisouza: See: 'several forms' and 'with some limitations' aren't exactly 'easy way' =D

lucasdemarchi: eheheh... fair enough.

otaviobp:
Branch updated with proposed fixes and new api for looping subsections

lucasdemarchi:
Great stuff!!

Thanks for all the rework. It looks like a really good API now.  I added 2 small commits on top with minor things/nitpicks. Applied.

PR #66, opened 2017-04-07
log: Aggregate some log messages that could flood output

edersondisouza:
To avoiding flooding output, some error/warning messages on routing
code path were increased to debug, and an aggregator sums how many of
those error code path were taken, printing the aggregate, if any,
each five seconds.

This way, we don't lose track of these code paths without flooding
output. As original messages are still there with debug log level, if any
interest on such situations appear, one can simply raise debug log level
to `debug` and get more datails.

### DIFF
--- a/endpoint.cpp
+++ b/endpoint.cpp
@@ -333,6 +333,15 @@ uint8_t Endpoint::get_trimmed_zeros(const struct buffer *buffer)
     return msg_entry->msg_len - msg->payload_len;
 }
 
+void Endpoint::log_aggregate(int interval)
+{
+    if (_incomplete_msgs > 0) {
+        log_warning("Endpoint %s [%d] sysid %u: %u incomplete messages in the last %d seconds",
+                    _name, fd, _system_id, _incomplete_msgs, interval);
+        _incomplete_msgs = 0;
+    }
+}
+
 int UartEndpoint::open(const char *path, speed_t baudrate)
 {
     struct termios2 tc;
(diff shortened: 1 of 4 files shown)


endpoint.cpp
```diff
@@ -413,7 +413,8 @@ int UartEndpoint::write_msg(const struct buffer *pbuf)
 
     /* Incomplete packet, we warn and discard the rest */
     if (r != (ssize_t) pbuf->len) {
-        log_warning("Discarding packet, incomplete write %zd but len=%u",
+        _mainloop->inc_incomplete_msg_error();
```
lucasdemarchi: ugh. no.  This counter should be inside the endpoint, not on mainloop

edersondisouza:
Updated to address @lucasdemarchi comment.

endpoint.h
```diff
@@ -86,6 +88,7 @@ class Endpoint : public Pollable {
     int read_msg(struct buffer *pbuf, int *target_system);
     virtual ssize_t _read_msg(uint8_t *buf, size_t len) = 0;
     bool _check_crc(const mavlink_msg_entry_t *msg_entry);
+    void inc_incomplete_msg_error() { _incomplete_msgs++; };
```
lucasdemarchi: this method can be removed

edersondisouza:
Updated to have one less method.

lucasdemarchi:
Applied with a patch on top for renames.

PR #68, opened 2017-04-12
Mult baud

edersondisouza:


endpoint.h
```diff
@@ -22,6 +22,9 @@
 #include "comm.h"
 #include "pollable.h"
 
+#define READ_MESSAGE_OK 1
+#define READ_MESSAGE_UNIDENTIFIED 2
```
lucasdemarchi: A good candidate to an enum

endpoint.cpp
```diff
@@ -32,10 +32,13 @@
 
 #include "log.h"
 #include "mainloop.h"
+#include "util.h"
 
 #define RX_BUF_MAX_SIZE (MAVLINK_MAX_PACKET_LEN * 4)
 #define TX_BUF_MAX_SIZE (8U * 1024U)
 
+#define UART_BAUD_RETRY 5
```
lucasdemarchi: add unit: UART_BAUD_RETRY_SEC

mainloop.cpp
```diff
@@ -59,6 +59,8 @@ int Mainloop::open()
         return -1;
     }
 
+    Endpoint::set_Mainloop(this);
```
lucasdemarchi: I'd prefer to be the other way around. There shouldn't be a "set_Mainloop()" method.

Rather add a static method `Mainloop::get_instance()`. Whoever needs it can call this method that will return the only Mainloop instance.

lucasdemarchi: And yes... this means it's probably a good idea to move the call to Mainloop::open() to happen earlier, so we only create endpoints when/if we have a valid Mainloop instance

edersondisouza: Ack

endpoint.h
edersondisouza: Do you mean just these two? Or all of the `read_msg` returns? Because currently, it may also return a whole range of 'invalid' values, starting on 0 to any negative `errno`.

lucasdemarchi: these 2

edersondisouza: I can do that, but isn't it weird to compare it with some old constant, like `-EINVAL` or `0`, and if not lesser than zero, with some enum?

edersondisouza:
Updated to address comments.

lucasdemarchi:
I pushed the commits except the last one.  Take a look on some fixes I did there as well, main one being the ability to inline the getter.

endpoint.cpp
```diff
@@ -346,35 +349,50 @@ void Endpoint::log_aggregate(unsigned int interval_sec)
...
-int UartEndpoint::open(const char *path, speed_t baudrate)
+UartEndpoint::~UartEndpoint()
+{
+    if (_change_baud_timeout) {
+        Mainloop::get_instance().del_timeout(_change_baud_timeout);
+        _change_baud_timeout = nullptr;
+    }
+}
+
+int UartEndpoint::open(const char *path, unsigned long baudrates[], size_t baud_size)
```
lucasdemarchi: instead of doing this, do the following:

* rename baudrate to "default_baudrate" - this is the baud we configure if the user never calls the method below
* add a set_speed() method

endpoint.cpp
```diff
@@ -346,35 +349,50 @@ void Endpoint::log_aggregate(unsigned int interval_sec)
...
     fd = ::open(path, O_RDWR|O_NONBLOCK|O_CLOEXEC|O_NOCTTY);
     if (fd < 0) {
         log_error_errno(errno, "Could not open %s (%m)", path);
         return -1;
     }
 
-    bzero(&tc, sizeof(tc));
-
-    if (ioctl(fd, TCGETS2, &tc) == -1) {
+    if (ioctl(fd, TCGETS2, &_tc) == -1) {
```
lucasdemarchi: nah... don't store the termios struct. Call TCGETS2 again when you need it (i.e. in the set_speed() method)

endpoint.cpp
```diff
@@ -384,7 +402,13 @@ int UartEndpoint::open(const char *path, speed_t baudrate)
         goto fail;
     }
 
-    log_info("Open UART [%d] %s:%u *", fd, path, baudrate);
+    log_info("Open UART [%d] %s:%lu *", fd, path, baudrates[_current_baud_idx]);
+
+    if (_bauds_size > 1) {
```
lucasdemarchi: move this to the `add_baudrate()` method

endpoint.cpp
```diff
@@ -394,6 +418,45 @@ int UartEndpoint::open(const char *path, speed_t baudrate)
     return -1;
 }
 
+bool UartEndpoint::_change_baud_cb(void *data)
```
lucasdemarchi: this callback is the one that handles the baud_idx and then simply call the `set_speed()` method

lucasdemarchi: And you need a log_info() here saying what baudrate you are trying

endpoint.cpp
```diff
@@ -394,6 +418,45 @@ int UartEndpoint::open(const char *path, speed_t baudrate)
...
+    fd = -1;
+    return false;
+}
+
+int UartEndpoint::read_msg(struct buffer *pbuf, int *target_sysid)
+{
+    int ret = Endpoint::read_msg(pbuf, target_sysid);
+
+    if (_change_baud_timeout != nullptr && ret == ReadOk) {
+        log_debug("Baudrate %lu responded, keeping it", _baudrates[_current_baud_idx]);
```
lucasdemarchi: log_info()

endpoint.cpp
edersondisouza: Hmmm - so it's caller (in the case, mainloop) task to keep retrying on different bauds? So another callback so main can set speed? Maybe it's too late on Friday and I lost my reasoning skills, but it appears strange.
And your PR has a `set_speed` method: should I rebase this on top of your pr?

lucasdemarchi: No.  set_speed() being public doesn't mean it needs to be called from outside.  You keep the timer here but but the timer callback just calls set_speed(), dealing with updating the `next_speed`

I'm waiting for that PR to be reviewed.... but yes, would be better to rebase.

edersondisouza:
Rebased and updated to address comments.

main.cpp
```diff
@@ -240,46 +240,87 @@ static int add_endpoint_address(const char *name, size_t name_len, const char *i
...
+        if (safe_atoul(tmp, &(*bauds)[i]) < 0) {
+            log_error("Invalid baud %s", tmp);
+            ret = -EINVAL;
+            goto atoul_error;
+        }
+        i++;
+        tmp += strlen(tmp) + 1;
+    }
+
+    free(nul_bauds_str);
```
lucasdemarchi: any reason why we are not using strtok? This looks complicated for something that will execute once.

main.cpp
```diff
@@ -240,46 +240,87 @@ static int add_endpoint_address(const char *name, size_t name_len, const char *i
...
         goto fail;
     }
 
-    if (baudrate != ULONG_MAX) {
-        conf->baud = baudrate;
-    }
-
-    if (conf->baud == ULONG_MAX) {
-        conf->baud = DEFAULT_BAUDRATE;
+    ret = add_bauds(&conf->bauds, &conf->bauds_size, bauds);
```
lucasdemarchi: if we are keeping bauds as an array rather than a list, it's a good candidate for std::vector.

main.cpp
```diff
@@ -552,11 +593,11 @@ static int parse_confs(ConfFile &conf)
...
     struct option_uart {
         char *device;
-        unsigned long baud;
+        char *bauds;
     };
     static const ConfFile::OptionsTable option_table_uart[] = {
-        {"baud",    false,  ConfFile::parse_ul,         OPTIONS_TABLE_STRUCT_FIELD(option_uart, baud)},
-        {"device",  true,   ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_uart, device)},
+        {"baud", false, ConfFile::parse_str_dup, OPTIONS_TABLE_STRUCT_FIELD(option_uart, bauds)},
+        {"device", true, ConfFile::parse_str_dup, OPTIONS_TABLE_STRUCT_FIELD(option_uart, device)},
```
lucasdemarchi: tables are exception to coding style. Please keep aligned.

main.cpp
edersondisouza: Fair...

main.cpp
edersondisouza: I need to let the C++ way flow =D

edersondisouza:
Updated to use `strtok` and `std::vector` instead of hand made split or C style vector (array + length).

fnoop:
This needs to be documented somewhere

PR #71, opened 2017-04-13
log: Add `auto` dialect, that will guess autopilot MAVLink dialect

edersondisouza:
This new MavlinkDialect option will identify flight stack MAVLink
dialect based on heartbeat.

This option is the new default, so people aren't expected to normally
deal with this setting.


edersondisouza:
Updated:
  - now with licenses
  - check for initialised log before anything else

PR #72, opened 2017-04-17
routing: Fix lots of 'message to uknown sysid' errors

edersondisouza:
mavlink2 messages may be trimmed - i.e., any trailing 0 byte is
discarded. However, mavlink-router wasn't taking that into consideration
when getting target_system from a message: it was reading past message
length.

This patch just assumes broadcast if target_system offset is past
message length: it was zero anyway.

### DIFF
--- a/endpoint.cpp
+++ b/endpoint.cpp
@@ -81,7 +81,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
     bool should_read_more = true;
     uint32_t msg_id;
     const mavlink_msg_entry_t *msg_entry;
-    uint8_t *payload, seq, sysid;
+    uint8_t *payload, seq, sysid, payload_len;
 
     if (fd < 0) {
         log_error("Trying to read invalid fd");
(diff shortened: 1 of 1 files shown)


endpoint.cpp
```diff
@@ -236,8 +238,11 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
     if (msg_entry == nullptr) {
         log_debug("No message entry for %u", msg_id);
     } else {
-        if (msg_entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM)
-            *target_sysid = payload[msg_entry->target_system_ofs];
+        if (msg_entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM) {
+            // if target_system is 0, it may have been trimmed out on mavlink2
```
lucasdemarchi: so the message actually has a target_system. Shouldn't you set *target_sysid to 0 in this case? Then it's clear -1 means "no target sysid" and 0 means broadcast.

edersondisouza: Yeah, it would be more semantically accurate.

zehortigoza: +1

zehortigoza:
LGTM just fix the typo on the commit message.

edersondisouza:
Updated to broadcast `0` and fix typo commit message.

PR #73, opened 2017-04-20
tests: Add unit test infra

edersondisouza:
Uses googletest [https://github.com/google/googletest] test files,
added as a submodule.

Makefile target 'test' changed to run both unit tests and current
routing test, also adapted to new infra.

Target 'test' will compile all files named `test*.cpp` on tests
directory - linking with googletest.

Right now, the only unit test is dummy test to show the infra,
naturaly, many improvements can come after this initial patch, like real
unit tests.


lucasdemarchi:
> Makefile target 'test' changed to run both unit tests and current
> routing test, also adapted to new infra.

target for compiling and running unit tests should be "check"

tests/run_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO check if it's worth redirecting output to some file
+
+RESULT=true
+printf "\n--- Running unit tests ---\n"
```
lucasdemarchi: ugh. Let's not reinvent the wheel on running unit tests.

https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/Makefile.am#n348
TESTS control all the tests to be run.

See "make check" in https://www.gnu.org/software/automake/manual/html_node/Standard-Targets.html

And: https://www.gnu.org/software/automake/manual/html_node/Tests.html

libell, bluez and systemd can also be used as a reference on how to integrate running testsuite

edersondisouza: Hmmm... Good to know that make handles those!

edersondisouza:
Update:
  - Using make features to run the tests
    - So 'make check' is the new target to run tests. Need to update CI.

edersondisouza:
CI tests updated.

PR #75, opened 2017-04-25
TCP improvements

lucasdemarchi:


endpoint.cpp
```diff
@@ -545,6 +545,7 @@ TcpEndpoint::TcpEndpoint()
 
 TcpEndpoint::~TcpEndpoint()
 {
+    close();
```
edersondisouza: I guess that `fd` is closed by `Pollable` destructor - but if this is for the logging, ok.

lucasdemarchi: You can't really call on the parent class a virtual method and expect the one on the child class to be executed.  Child class will already be gone for good when the destructor on the parent is executed.

lucasdemarchi: ok... now checking again, `fd` is from the parent class. So... yes. We could just log the message on the destructor.  However I think it makes sense to do inside the close() method since it would catch all cases in which the connection is actually closed.  Thoughts?

PR #77, opened 2017-04-27
Uart startup improvements

lucasdemarchi:


endpoint.cpp
```diff
@@ -371,14 +371,28 @@ int UartEndpoint::open(const char *path, speed_t baudrate)
...
     tc.c_oflag &= ~(OCRNL | ONLCR | ONLRET | ONOCR | OFILL | OPOST);
-    tc.c_lflag &= ~(ECHO | ECHOE | ECHOK | ECHOCTL | ECHOKE | ECHONL | ICANON | IEXTEN | ISIG | TOSTOP);
-    tc.c_cflag &= ~(CSIZE | PARENB | CBAUD | CRTSCTS);
+
+    tc.c_lflag &= ~(ECHO | ECHOE | ECHOK | ECHOCTL | ECHOKE | ECHONL | ICANON | IEXTEN | ISIG);
+
+    /* never send SIGTTOU*/
+    tc.c_lflag &= ~(TOSTOP);
+
+    /* disable flow control */
```
edersondisouza: Well - on PX4 mavlink module, there's this line https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_main.cpp#L824 that actually tries to enable flow control - couldn't this create some problems if we always disable it?

endpoint.cpp
```diff
@@ -347,7 +347,42 @@ void Endpoint::log_aggregate(unsigned int interval_sec)
...
+        log_error_errno(errno, "Could not set terminal attributes (%m)");
+        return -1;
+    }
+
+    if (ioctl(fd, TCFLSH, TCIOFLUSH) == -1) {
+        log_error_errno(errno, "Could not flush terminal (%m)");
+        return -1;
+    }
+
+    log_info("UART [%d] speed = %u", fd, baudrate);
```
edersondisouza: If the flush doesn't happen, terminal speed may still change, right? If so, we may want to log_info to which speed we are changing anyway - iff things work even if flush fails.
If last sentence is not true, ignore this comment.

endpoint.cpp
lucasdemarchi: we currently don't support flow control (and we didn't before...). We need to properly enable it

> 			PX4_WARN("hardware flow control not supported");

for the hardwares we currently support, PX4 would print this message, because flow control is disabled on the stm32 side

endpoint.cpp
lucasdemarchi: right. just pushed a fix.

mindThomas:
OBS. This PR does not seem to fix it completely.
At least it is not possible to just stop mavlink-routerd and start Mavros or the Ulog script (with the applied PR, https://github.com/intel-aero/meta-intel-aero/pull/134) without first having to call:
`stty -F /dev/ttyS1 min 1`

lucasdemarchi:
There was another PR merged yesterday that should fix it. Please test master branch

PR #78, opened 2017-05-02
Add new log implementation

lucasdemarchi:
@otaviobp @anselmolsm


new_log.h
```diff
@@ -0,0 +1,66 @@
...
+    static int close();
+
+    static Level get_max_level() _pure_ { return _max_level; }
+    static void set_max_level(Level level);
+
+#define debug(...) log_internal(Log::Level::DEBUG, __VA_ARGS__)
+#define info(...) log_internal(Log::Level::INFO, __VA_ARGS__)
+#define notice(...) log_internal(Log::Level::NOTICE, __VA_ARGS__)
+#define warning(...) log_internal(Log::Level::WARNING, __VA_ARGS__)
+#define error(...) log_internal(Log::Level::ERROR, __VA_ARGS__)
```
lucasdemarchi: humn... I actually dislikes this and am willing to go back to what it was before.  The problem is that you now call Log::debug(), but `debug()` is a macro in the global namespace without proper prefix.

lucasdemarchi: I did a template version that would work by means of std::forward... but then we lose the format attribute.

lucasdemarchi:
New version pushed.

main.cpp
```diff
@@ -722,15 +722,15 @@ int main(int argc, char *argv[])
 {
     Mainloop &mainloop = Mainloop::init();
 
-    log_open();
+    Log::open();
```
otaviobp: Do we need to have open/close functions? Can't we use static initializes for that? Or a singleton design pattern?

log.h
```diff
@@ -20,55 +20,46 @@
...
 
-#define log_full_errno(level, error, ...)                               \
-    ({                                                                  \
-         int _level = (level), _e = (error);                            \
-         (log_get_max_level() >= LOG_PRI(_level))                       \
-            ? log_internal(_level, _e, __FILE__, __LINE__, __VA_ARGS__) \
-            : -abs(_e);                                                 \
-        })
+    static void logv(Level level, const char *format, va_list ap);
+    static void log(Level level, const char *format, ...) _printf_format_(2, 3);
```
otaviobp: It would be nice, as we are now using a c++ class, to support logging objects, instead of only logging char* strings. I like the approach from std::cout or qLog(), that uses the operator << overriding for logging.
I would like to have something like:

log_debug() << MyObject;

instead of 

log_debug(MyObject.get_string()); //or something similar to that.

main.cpp
lucasdemarchi: We will need the open/close whenever we are not using only stderr for logging. For example when/if we integrate it with the journal.

log.h
lucasdemarchi: I think it's a matter of preference... I don't like the << operator since for my use cases the log is always a string not a function. Often logging is confused as normal output, or things that should go into dbg() (for which ifdef'ing by LOG_DEBUG would accomplish the same result).
If the implementation is reasonable, I wouldn't oppose to it, as long as it keep the properties of this lightweight logging. Feel free to provide an additional method for dealing with object logging.

otaviobp:
ok for me then.

PR #79, opened 2017-05-03
Fix uart reset

lucasdemarchi:
When resetting the tty instead of putting it in "raw" mode use "sane" mode
(see stty(1)) and also reset special characters to their defaults
(termios(3)).

### DIFF
--- a/xtermios.cpp
+++ b/xtermios.cpp
@@ -17,17 +17,45 @@
  */
 #include "xtermios.h"
 
+#include <string.h>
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Validated that it works on a usb-serial connected to Pixhawk... need to check on Aero/Joule where the problem is happening.

lucasdemarchi:
Also validated on a joule + pixhawk and on Aero RTF.

Issue #35, opened 2017-03-17
Add systemd service file


edersondisouza:

lucasdemarchi:
@avinash-palleti for mavlink-router, please add the systemd service file in this repository. Some references on how to do it:

https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/Makefile.am#n44
https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/configure.ac#n246
https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/src/bluetooth.service.in  (but our type is simple, not dbus

PR #80, opened 2017-05-08
Add Systemd service file

Rjasuja:

Issue: #35

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,9 +57,39 @@ AC_ARG_WITH([rootlibdir],
         [], [with_rootlibdir=$libdir])
 AC_SUBST([rootlibdir], [$with_rootlibdir])
 
+AC_ARG_WITH([systemdsystemunitdir],
+			AC_HELP_STRING([--with-systemdsystemunitdir=DIR],
+			[path to systemd system unit directory]),
+					[path_systemunitdir=${withval}])
+if (test "${enable_systemd}" != "no" && test -z "${path_systemunitdir}"); then
+	AC_MSG_CHECKING([systemd system unit dir])
+	path_systemunitdir="`$PKG_CONFIG --variable=systemdsystemunitdir systemd`"
+	if (test -z "${path_systemunitdir}"); then
+		AC_MSG_ERROR([systemd system unit directory is required])
+	fi
+	AC_MSG_RESULT([${path_systemunitdir}])
+fi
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -57,10 +57,40 @@ AC_ARG_WITH([rootlibdir],
         [], [with_rootlibdir=$libdir])
 AC_SUBST([rootlibdir], [$with_rootlibdir])
 
+AC_ARG_WITH([systemdsystemunitdir],
+			AC_HELP_STRING([--with-systemdsystemunitdir=DIR],
+			[path to systemd system unit directory]),
+					[path_systemunitdir=${withval}])
+if (test "${enable_systemd}" != "no" && test -z "${path_systemunitdir}"); then
+	AC_MSG_CHECKING([systemd system unit dir])
+	path_systemunitdir="`$PKG_CONFIG --variable=systemdsystemunitdir systemd`"
```
edersondisouza: If I run `./configure` on a clean mavlink-router (just after `./autogen.sh`, of course), I get this error

```
checking systemd system unit dir... ./configure: line 17114: --variable=systemdsystemunitdir: command not found
```
(both on ubuntu or arch)
I can only get this working if I disable systemd using `--disable-systemd`. My guess is that `$PKG_CONFIG` is not being set on my environments. Am I missing something?

I'm not well acquainted with autotools, but maybe macro `PKG_CHECK_VAR` can be of help? See https://autotools.io/pkgconfig/variables.html

avinash-palleti: @edersondisouza use --with-systemdsystemunitdir=<sysdir>  --with-systemduserunitdir=<userdir>

<sysdir>  is /lib/systemd/system
<userdir> is /usr/lib/systemd

edersondisouza: Oh, indeed there's another message about the missing flag - but not stating the flag itself, maybe that can be improved? Something like 'Please use ´--with-systemdsystemunitdir=´? 

Also, update README and configure example showed when one runs `./autogen.sh`?

edersondisouza: Nevertheless, this code **is not** working. It is trying to use `pkg-config` to get default values for those variables, but `PKG_CONFIG` is not set. You must use `PKG_PROG_PKG_CONFIG` to set it. See https://autotools.io/pkgconfig/pkg_check_modules.html

PR #84, opened 2017-05-10
Add Systemd service file

Rjasuja:

Issue: #35

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,6 +18,7 @@ BUILT_SOURCES = include/mavlink/ardupilotmega/mavlink.h
 
 clean-local:
 	rm -rf $(top_builddir)/include/mavlink
+	rm -f mavlink-router.service
 	rm -f $(top_builddir)/gtest.a
 	rm -f $(top_builddir)/tests/unit_test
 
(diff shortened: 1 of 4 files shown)


avinash-palleti:
@lucasdemarchi I verified it on my system. 

"./autogen.sh c " able to find all paths and no configure errors. Not sure why Travis build is failing.

lucasdemarchi:
```
checking systemd system unit dir... configure: error: systemd system unit directory is required
```

From https://travis-ci.org/01org/mavlink-router/builds/230678099?utm_source=github_status&utm_medium=notification

@avinash-palleti it's failing because now systemd is mandatory (unless disabled).  In your system it may succeeded since you have systemd installed, so it's able to retrieve the systemd system unit dir via pkgconfig.

lucasdemarchi:
This is a diff on top of your patch that should make it work:
https://hastebin.com/jozuvarini

I also reworded the readme a little bit, too.

lucasdemarchi:
Oh... and also removed the user unit dir since we aren't using it (nor I think we will ever do).
https://hastebin.com/sediqunoti

PR #86, opened 2017-05-10
Add Systemd service file

Rjasuja:

Issue: #35


Rjasuja:
@lucasdemarchi  please check this one with your diff

configure.ac
```diff
@@ -74,6 +91,9 @@ AC_CONFIG_FILES([
 ])
 
 #####################################################################
+AC_SUBST(SYSTEMD_SYSTEMUNITDIR, [${path_systemunitdir}])
+
+AC_SUBST(SYSTEMD_USERUNITDIR, [${path_userunitdir}])
```
edersondisouza: I guess that as we are not using user units, this line is unnecessary.

lucasdemarchi:
Fixed comments from @edersondisouza, added entry to .gitignore and applied. Thanks.

PR #91, opened 2017-05-22
conf files: Preference to command line options

edersondisouza:
Command line options should have preference over config files, as they
enable quick testing, for instance.

Work has been done on some relevant 'General' options that can be
overriden via command line options, like log verbosity and stats
reporting.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -49,6 +49,21 @@ static struct options opt = {
         .mavlink_dialect = Auto
 };
 
+static const struct option long_options[] = {
+    { "endpoints",              required_argument,  NULL,   'e' },
+    { "conf-file",              required_argument,  NULL,   'c' },
+    { "conf-dir" ,              required_argument,  NULL,   'd' },
+    { "report_msg_statistics",  no_argument,        NULL,   'r' },
+    { "tcp-port",               required_argument,  NULL,   't' },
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -613,10 +613,34 @@ static int parse_confs(ConfFile &conf)
...
+    // Merge general options giving preference to those from command line,
+    // when different from default
+    if (opt.tcp_port == ULONG_MAX)
+        opt.tcp_port = options.tcp_port;
+    if (!opt.report_msg_statistics)
+        opt.report_msg_statistics = options.report_msg_statistics;
+    if (opt.logs_dir == nullptr && options.logs_dir != nullptr)
+        opt.logs_dir = strdup(options.logs_dir);
+    if (opt.debug_log_level != (int)Log::Level::INFO)
+        opt.debug_log_level = options.debug_log_level;
```
lucasdemarchi: ugh... I don't want to do this for each and every option. You need to defer the command line parsing (probably with the exception of -c) to be done after the configuration files so it overrides them.

edersondisouza: Yeah, I was a bit lazy to separate options parsing... will do that, them...

edersondisouza:
Updated to perform command line parsing in two steps.

PR #92, opened 2017-06-02
Component id is back

edersondisouza:
mavros uses component_id on mavlink messages, so time to handle them again.

Note that this is a dangerous PR, review with care =D

As I've also updated tests, `make check` itself may be wrong, so, review with care²

### DIFF
--- a/endpoint.cpp
+++ b/endpoint.cpp
@@ -67,21 +67,24 @@ bool Endpoint::handle_canwrite()
 
 int Endpoint::handle_read()
 {
-    int target_sysid, r;
+    int target_sysid, target_compid, r;
+    uint8_t src_sysid, src_compid;
     struct buffer buf{};
 
-    while ((r = read_msg(&buf, &target_sysid)) > 0)
-        Mainloop::get_instance().route_msg(&buf, target_sysid, _system_id);
+    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid)) > 0)
+        Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid,
+                                           src_compid);
 
     return r;
 }
 
-int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
+int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compid,
+                       uint8_t *src_sysid, uint8_t *src_compid)
 {
     bool should_read_more = true;
     uint32_t msg_id;
     const mavlink_msg_entry_t *msg_entry;
-    uint8_t *payload, seq, sysid, payload_len;
+    uint8_t *payload, seq, payload_len;
 
     if (fd < 0) {
         log_error("Trying to read invalid fd");
@@ -114,7 +117,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
         if (r <= 0)
             return r;
 
-        log_debug("%s: Got %zd bytes", _name, r);
+        log_debug("%s: Got %zd bytes [%d]", _name, r, fd);
         rx_buf.len += r;
     }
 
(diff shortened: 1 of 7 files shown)


endpoint.cpp
```diff
@@ -271,6 +280,57 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
     return msg_entry != nullptr ? ReadOk : ReadUnkownMsg;
 }
 
+void Endpoint::_add_sys_comp_id(uint16_t sys_comp_id)
+{
+    _sys_comp_ids.insert(sys_comp_id);
+}
+
+bool Endpoint::has_sys_id(int sysid)
```
lucasdemarchi: sysid should be unsigned

endpoint.cpp
```diff
@@ -271,6 +280,57 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
...
+{
+    for (auto it = _sys_comp_ids.begin(); it != _sys_comp_ids.end(); it++) {
+        if (((*it >> 8) | (sysid & 0xff)) == sysid)
+            return true;
+    }
+    return false;
+}
+
+bool Endpoint::has_sys_comp_id(int sysid, int compid)
```
lucasdemarchi: sysid and compid -> unsigned

endpoint.cpp
```diff
@@ -271,6 +280,57 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid)
...
+{
+    return _sys_comp_ids.find(((sysid & 0xff) << 8) | (compid & 0xff)) != _sys_comp_ids.end();
+}
+
+bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid,
+                          uint8_t src_compid)
+{
+    if (Log::get_max_level() == Log::Level::DEBUG) {
```
lucasdemarchi: should be '>=', not '=='

endpoint.h
```diff
@@ -121,7 +126,7 @@ class Endpoint : public Pollable {
 
     const bool _crc_check_enabled;
     uint32_t _incomplete_msgs = 0;
-    uint8_t _system_id = 0;
+    std::set<uint16_t> _sys_comp_ids;
```
lucasdemarchi: this should be a vector, not a set.

lucasdemarchi:
In general it looks good to me. The logic seems correct, but I don't like the use of std::set because it's slow: use vector instead.

Please test this under `perf` and add a comment about the increased overhead if any.

zehortigoza:
What is the use case of ROS with component id? Why the current setup do not work?
I'm guessing that ROS have multiple system ids, right? If that is true we only need keep the multiple system id per endpoint that you implemented in here.

Also filtering by component id will cause some messages not be routed, example: QGC wants to send some message to vehicle with a specific component id but the vehicle never had send a message with this component id, it would just be dropped. From my understanding component id is something to be handled internally, like ports in a IP network.

edersondisouza:
@zehortigoza No, mavros actually uses the same sysid as flight stack, with a different component id. So, we have two endpoints with same sysid, but different component id.

Regarding second issue - not routing when component id is specified - I'm not sure how to approach it. Maybe we could handle unknown component id as broadcast component id?

lucasdemarchi:
> Regarding second issue - not routing when component id is specified - I'm not sure how to approach it. Maybe we could handle unknown component id as broadcast component id?

This is the reason why components should send a heartbeat, even if not at 1Hz

zehortigoza:
Maybe is better just add to the Aero documentation how to change the mavros system id.

http://wiki.ros.org/mavros#Usage
UDP: udp://[bind_host][:port]@[remote_host][:port][/?ids=sysid,compid]
ids from URL overrides ids given by parameters ~system_id and ~component_id.

lucasdemarchi:
@zehortigoza we already did that (it's in the bug report, to be moved to the docs actually). But we should use the principle of the least surprise and make mavros work without passing extra params. And... we were cheating not to use component id. For good or for bad, it's there as part of mavlink protocol and we should handle it.

edersondisouza:
Updated to address comments by @lucasdemarchi

lucasdemarchi:
Applied with a simple patch on top. thanks.

Issue #82, opened 2017-05-09
Separate mainloop from app


lucasdemarchi:
This grew way more than it should. Mainloop should be just mainloop, not handling "app specific tasks".  Move everything that's specific to mavlink-router elsewhere so mainloop is a small mainloop implementation rather than mainloop + app.

jbeyerstedt:
Will this be fixed by #242?

jkflying:
#242 is a starting point to at least be able to run the contents of the loop one iteration at a time. Once that's in it will be a lot easier to split the loop into smaller chunks, and unit tests can be written to make sure nothing breaks in the process.

PR #98, opened 2017-06-21
Rtps

edersondisouza:
Experimental RTPS implementation

This patch series provides an initial/experimental RTPS communication with PX4 flight stack. Currently based only on three uORB mesages, two outbound (from flight stack), `sensor_combined` and `vehicle_command_ack`, and one inbound (to flight stack), `vehicle_command`.

This new implementation is provided on a new binary, `rtpsd`. To allow some code sharing with existing `mavlink-routerd`, some code reorganisation was made.

Although only a initial version, it allows UART sharing with mavlink messages. `rtpsd`, besides publishing and subscribing on RTPS topics, also sends/receives mavlink messages to an UDP port, that can be connected to mavlink-router, for instance. Note that in this scenario, `rtpsd` is the one that connects the UART.

Two sample applications are also provided: one to publish commands to flight stack, another that simply subscribes on sensor combined topic.

Naturally, for this application to make sense, an rtps module must be running on the flight stack, providing at least those three uORB topics.


Makefile.am
```diff
@@ -38,6 +38,8 @@ AM_CPPFLAGS = \
 	-include $(top_builddir)/config.h \
 	-I$(top_builddir) \
 	-I$(top_srcdir) \
+	-I$(top_srcdir)/src/mavlink-router \
+	-I$(top_srcdir)/src/common \
```
lucasdemarchi: > One may suspect that `mainloop` and `pollable` classes belong to
> `src/common` - and they are right. However, currently, `mainloop` is the
> one class that perform the routing, so it will live on
> `src/mavlink-router` till it's converted to a pure mainloop.

I'd prefer to have this code move and the correspondent separation of mainloop from app in a separate PR (#82). And this PR to be done on top.  Code move is very easy to apply... this needs a thorough review.

lucasdemarchi: Ad it also has the benefit you don't have to explain it on the commit message :)

edersondisouza: Code move on a separated PR, with this one on top: easy.
Properly extract mainloop: not so...(way easy just explain on commit message =P )

Can I do just the first now? =D

lucasdemarchi: That's a long standing issue that I'd like to be fixed for good. I prefer fixing the issue first.

LorenzMeier:
@edersondisouza @lucasdemarchi Could we get this rebased and in? Firmware support is coming in as well.

avinash-palleti:
@lucasdemarchi I have rebased this PR and verified rtpsd. Please review: #115

PR #100, opened 2017-06-28
Add support for HW flow control

### LINKED ISSUE #89: Support for HW Flow Control
@lucasdemarchi

Will it be possible to add support for hardware flow control for serial links? Probably as a configuration parameter.

lucasdemarchi:
Fix #89 

@mhkabir can you give this a try?  I tested on a pixhawk and seems to be working fine.  You need to set `FlowControl = true`  in the config file


mhkabir:
Perfect, thank you. Will test in a bit.

Issue #97, opened 2017-06-13
Buffer overflow while logging


nkhoit:
Trying to log results in a buffer overflow immediately after starting the process.

mavlink-routerd -l . /dev/ttyTHS2:921600
Open UART [4] /dev/ttyTHS2 *
UART [4] speed = 921600
Open TCP [5] 0.0.0.0:5760 *
Logging target system_id=1 on 02021-2017-06-12_23-58-43.ulg
*** buffer overflow detected ***: mavlink-routerd terminated
Aborted

I'm running it on an Nvidia TX1 with Ubuntu 16.04.2 LTS (arm64) 
g++ --version
g++ (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

edersondisouza:
Hi @nkhoit,

Do you have a core dump file? Or could you please run `mavlink-routerd` on valgrind or gdb? So we could get more information on this issue

Thanks

nkhoit:
Valgrind:
valgrind ./mavlink-router/mavlink-routerd -l . /dev/ttyTHS2:921600
==20435== Memcheck, a memory error detector
==20435== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==20435== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==20435== Command: ./mavlink-router/mavlink-routerd -l . /dev/ttyTHS2:921600
==20435==
Open UART [4] /dev/ttyTHS2 *
UART [4] speed = 921600
Open TCP [5] 0.0.0.0:5760 *
Logging target system_id=1 on 00002-2017-06-13_20-50-29.ulg
==20435== Syscall param epoll_pwait(sigmask) points to unaddressable byte(s)
==20435==    at 0x4B10240: epoll_pwait (epoll_pwait.c:42)
==20435==    by 0x40E59F: Mainloop::loop() (mainloop.cpp:292)
==20435==    by 0x40296B: main (main.cpp:805)
==20435==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==20435==
**20435** *** memcpy_chk: buffer overflow detected ***: program terminated
==20435==    at 0x48479CC: ??? (in /usr/lib/valgrind/vgpreload_memcheck-arm64-linux.so)
==20435==
==20435== HEAP SUMMARY:
==20435==     in use at exit: 95,055 bytes in 17 blocks
[...]

PR #102, opened 2017-06-30
Fix buffer overflow in ulog and binlog

zehortigoza:
This should fix the buffer overflow reported here https://github.com/01org/mavlink-router/issues/97#issuecomment-312353368

### DIFF
--- a/src/mavlink-router/binlog.cpp
+++ b/src/mavlink-router/binlog.cpp
@@ -97,11 +97,20 @@ int BinLog::write_msg(const struct buffer *buffer)
         return buffer->len;
     }
 
+    const mavlink_msg_entry_t *msg_entry = mavlink_get_msg_entry(msg_id);
+    if (!msg_entry) {
+        return buffer->len;
+    }
+
+    if (payload_len > msg_entry->msg_len) {
+        payload_len = msg_entry->msg_len;
+    }
+
     uint8_t *payload;
 
     if (mavlink2) {
         payload = buffer->data + sizeof(struct mavlink_router_mavlink2_header);
-        trimmed_zeros = get_trimmed_zeros(buffer);
+        trimmed_zeros = get_trimmed_zeros(msg_entry, buffer);
     } else {
         payload = buffer->data + sizeof(struct mavlink_router_mavlink1_header);
         trimmed_zeros = 0;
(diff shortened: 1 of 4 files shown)


src/mavlink-router/ulog.cpp
```diff
@@ -168,6 +168,7 @@ int ULog::write_msg(const struct buffer *buffer)
         mavlink_msg_logging_ack_encode(_system_id, MAV_COMP_ID_ALL, &msg, &ack);
         _send_msg(&msg, _target_system_id);
         /* no break needed, message will be handled by MAVLINK_MSG_ID_LOGGING_DATA case */
+        _fall_through_;
```
lucasdemarchi: a simple `/* fall-through */` should solve this, no? You can even replace the "no break needed" part above

zehortigoza: It did not worked even if the "no break needed" part was deleted.

src/mavlink-router/binlog.cpp
```diff
@@ -97,11 +97,20 @@ int BinLog::write_msg(const struct buffer *buffer)
         return buffer->len;
     }
 
+    const mavlink_msg_entry_t *msg_entry = mavlink_get_msg_entry(msg_id);
```
lucasdemarchi: I propose the following change to the commit message:

```
MAVLink 2.0 messages can have extensions which allows new fields
to be added. This happened to mavlink_command_ack_t, gaining an
additional byte with the percentage of progress in case a command
takes a long time to finish.

So if MAVLink router was built with a MAVLink version without
this extension and the flight stack was built with, it would case
a buffer overflow on MAVLink router.

According to the spec, the rules for extensions are:

"(...) the fields after the extensions line are extended fields. The rules
for extended messages are:

* if sent as a MAVLink1 message then extended fields are not sent
* if received by an implementation that doesn't have the extended
  fields then the fields will not be seen
* if sent by an implementation that doesn't have the extended fields
  then the recipient will see zero values for the extended fields" 
```

So here what we do is to truncate the message to be considered the size of the known message rather than the payload length.

zehortigoza: done

lucasdemarchi:
Thank you, pull request merged! <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #105, opened 2017-07-01
Fix Log routing + minor fixes on Log

zehortigoza:
Fixes: https://github.com/01org/mavlink-router/issues/103

### DIFF
--- a/src/mavlink-router/autolog.cpp
+++ b/src/mavlink-router/autolog.cpp
@@ -66,13 +66,10 @@ void AutoLog::stop()
 {
     if (_logger)
         _logger->stop();
-
-    _system_id = 0;
 }
 
 bool AutoLog::start()
 {
-    _system_id = LOG_ENDPOINT_SYSTEM_ID;
     _add_sys_comp_id(LOG_ENDPOINT_SYSTEM_ID << 8);
 
     return true;
--- a/src/mavlink-router/autolog.cpp
+++ b/src/mavlink-router/autolog.cpp
@@ -31,20 +31,23 @@ int AutoLog::write_msg(const struct buffer *buffer)
     const bool mavlink2 = buffer->data[0] == MAVLINK_STX;
     uint32_t msg_id;
     uint8_t *payload;
+    uint8_t source_system_id;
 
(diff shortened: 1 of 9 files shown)


nkhoit:
Fixed #103 for me on my Ubuntu 16.04 x86-64 desktop.

src/mavlink-router/autolog.cpp
```diff
@@ -73,6 +73,7 @@ void AutoLog::stop()
 bool AutoLog::start()
 {
     _system_id = LOG_ENDPOINT_SYSTEM_ID;
+    _add_sys_comp_id(LOG_ENDPOINT_SYSTEM_ID << 8);
```
lucasdemarchi: do we need it here? It's only purpose is to "wait for heartbeat to decide which logger to start". It will receive a heartbeat from the flight stack since that's a broadcast.

In `write_msg()` when we call `_logger->start()` we will call LogEndpoint::start() which will do everything needed.

zehortigoza: yes, after it initialize the right Log backend it will forward every write_msg() call to the backend.

```
int AutoLog::write_msg(const struct buffer *buffer)
{
    if (_logger != nullptr) {
        return _logger->write_msg(buffer);
    }
    ...
```

So it need a system id so router can route messages that have log backends as target.

src/mavlink-router/autolog.cpp
```diff
@@ -80,3 +80,12 @@ bool AutoLog::start()
 
     return true;
 }
+
+void AutoLog::print_statistics()
```
lucasdemarchi: oh... I think this shows that the autolog functionality should not really be done the way it is right now.... it can lead to more places using the wrong "wrapper object" rather than the real one, particularly on methods implemented on the parent's class like this one.

lucasdemarchi: Anyway, it's ok for now.

lucasdemarchi:
Thank you, pull request merged! <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #109, opened 2017-08-01
Use gnu++11 to handle fallthrough warning

lucasdemarchi:
GCC 7 turned the fallthrough warning on by default. These are the ways
to turn them handle them on the cases it should fallthrough:

    switch (cond) {
    case 1:
        bar (1);
        __attribute__ ((fallthrough)); // C and C++03
    case 2:
        bar (2);
        [[gnu::fallthrough]]; // C++11 and C++14
    case 3:
        bar (3);
        [[fallthrough]]; // C++17 and above
    /* ... */
    }

More info: https://dzone.com/articles/implicit-fallthrough-in-gcc-7
And it does work with clang as well.

### DIFF
--- a/src/common/macro.h
+++ b/src/common/macro.h
@@ -25,16 +25,6 @@
 #define _pure_ __attribute__((pure))
 #define _packed_ __attribute__((packed))
 
-#ifdef __has_cpp_attribute
-#if __has_cpp_attribute(fallthrough)
-#define _fall_through_ [[fallthrough]]
-#endif
-#endif
-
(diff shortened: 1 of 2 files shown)


zehortigoza:
But this will not work in old versions of gcc.

lucasdemarchi:
It is gnu++11. If the compiler supports gnu++11 then it must work.

And we require gnu++11 to build the project

zehortigoza:
Ok, if this works in old versions looks good to me.

Issue #107, opened 2017-07-18
duplicated uorb topics / reflecting mavlink msgs


ChristophTobler:
When I run the optical flow app (https://github.com/zehortigoza/aero-optical-flow/tree/without-bmi) on the aero and publish the messages at 15Hz, I see them in PX4 (uorb top in mavlink_shell) at ~26Hz.
```
TOPIC NAME                    INST #SUB #MSG #LOST #QSIZE
optical_flow                     0    4   25    52 1
```
With `listener optical_flow 10`, I can see that some messages are duplicated (see timestamps in [output](http://pasted.co/325b34e6))

@zehortigoza FYI

lucasdemarchi:
Humn... Do you have any reason to believe it's a problem in mavlink-router?  I'd guess it's much more likely to be something in the aero-optical-flow repository.

ChristophTobler:
Could be, yes. I just verified that https://github.com/zehortigoza/aero-optical-flow/blob/without-bmi/src/main.cpp#L238 runs at ~15Hz. But @zehortigoza has better insight on that.

ChristophTobler:
@lucasdemarchi We're pretty certain that this is a mavlink-router issue. Two more indications:
* when running a mavlink shell and pressing enter it behaves like pressing enter twice
* run `mavlink status` in a mavlink shell and you will see that rx is identical to tx even though we are not sending anything. 

```
GCS heartbeat:	550067 us ago
	mavlink chan: #1
	type:		GENERIC LINK OR RADIO
	flow control:	ON
	rates:
	tx: 7.088 kB/s
	txerr: 0.000 kB/s
	rx: 7.088 kB/s
	rate mult: 1.000
	accepting commands: YES
	MAVLink version: 2
	transport protocol: serial (/dev/ttyS1 @460800)
```

Seems like mavlink-router is reflecting the messages.

lucasdemarchi:
I'm trying to reproduce this. I'm writing below how I'm testing, let me know if I am missing any step you are doing.

Connect to Aero over wifi. Make sure mavlink-router is running:

``` sh
ps -ef  | grep mavlink-router
```

csd is running as well by default but AFAIR it's not sending any mavlink packet on the released versions (and it didn't change anything here). Anyway, you can stop it:

``` sh
systemctl stop csd
```

On my host computer I start mavlink_shell:

``` sh
./Tools/mavlink_shell 0.0.0.0:14550
```

This is a replay of my shell:
[...]

PR #113, opened 2017-08-14
Prevent reflecting packets back on origin endpoints

NaterGator:
My fix for #107

### DIFF
--- a/src/mavlink-router/mainloop.cpp
+++ b/src/mavlink-router/mainloop.cpp
@@ -132,12 +132,12 @@ int Mainloop::write_msg(Endpoint *e, const struct buffer *buf)
 }
 
 void Mainloop::route_msg(struct buffer *buf, int target_sysid, int target_compid, int sender_sysid,
-                         int sender_compid)
+                         int sender_compid, Endpoint *src)
 {
     bool unknown = true;
 
     for (Endpoint **e = g_endpoints; *e != nullptr; e++) {
-        if ((*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid)) {
+        if (*e != src && (*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid)) {
             log_debug("Endpoint [%d] accepted message to %d/%d from %u/%u", (*e)->fd, target_sysid,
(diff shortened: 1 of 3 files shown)


src/mavlink-router/mainloop.cpp
```diff
@@ -132,12 +132,12 @@ int Mainloop::write_msg(Endpoint *e, const struct buffer *buf)
...
 
 void Mainloop::route_msg(struct buffer *buf, int target_sysid, int target_compid, int sender_sysid,
-                         int sender_compid)
+                         int sender_compid, Endpoint *src)
 {
     bool unknown = true;
 
     for (Endpoint **e = g_endpoints; *e != nullptr; e++) {
-        if ((*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid)) {
+        if (*e != src && (*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid)) {
```
edersondisouza: What if we have two different 'elements' connected on the same endpoint, each with its own sysid? I've seen this while dealing with ROS stuff - can't remember exactly the scenario now.

edersondisouza: Anyway, please check `accept_msg`  method on endpoint. If this change is deemed necessary anyway, I'd say that it's better keep this kind of logic on that method, than starting scattering this throughout the code.

NaterGator: Are you saying they are transmitting and listening on the same endpoint? What type of endpoint?

edersondisouza: TCP/UDP. As I said, I don't remember properly, so maybe I'm mistaken. I sure remember having different compid on the same endpoint - but not so sure about sysid behaviour at that time. Sorry for not being able to be specific...

NaterGator: That's ok, I'll move this check to accept_msg as requested and spend some more time running on it. Maybe it makes sense to add a flag / config option to enable or disable loopback?

lucasdemarchi: @NaterGator any news on this?

lucasdemarchi:
See #118

PR #114, opened 2017-08-14
samples: Improve px4-offboard-mode

zehortigoza:
This is a more complete example, doing takeoff, moving arround and landing.

**Only tested in simulator yet**

### DIFF
--- a/examples/px4-offboard-mode.cpp
+++ b/examples/px4-offboard-mode.cpp
@@ -17,18 +17,22 @@
  */
 
 /*
- * This example shows how to change to offboard mode in PX4.
+ * This example shows how to fly in offboard mode in PX4.
  *
  * PX4 will only change to offboard mode and keep it in this mode if it is
  * receiving one of this messages: mavlink_set_position_target_local_ned_t,
  * mavlink_set_actuator_control_target_t or mavlink_set_attitude_target_t
  * at least at 0.5 hz.
  *
- * At each 50msec this example will request to the vehicle to move
- * 0 meters in X, 0 meters in Y, 0 meters in Z and 0 rads in yaw
- * (position_set_send(0, 0, 0, 0)) from the actual position of the vehicle and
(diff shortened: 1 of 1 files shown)


examples/px4-offboard-mode.cpp
```diff
@@ -181,26 +330,183 @@ static void set_mode_send(enum px4_modes mode)
...
+    default:
+        printf("Unknown action action id=%u\n", t->action);
+    }
+}
+
 static void timeout_callback()
 {
-    position_set_send(0, 0, 0, 0);
+    static float x = 0, y = 0, z = 0, yaw = 0;
+    static uint8_t task_list_index = 0;
```
lucasdemarchi: can these be a static global struct?  static variables inside functions are really bad to keep state.

You are always incrementing task_list_index and never checking by sizeof(list).  I think you will access invalid memory, no?

lucasdemarchi:
Great stuff, thanks.

examples/px4-offboard-mode.cpp
zehortigoza: It do not happen because when the task list reach to END action it do not increment anymore but I added the code to check the index and the array size.

RohanKurane:
@zehortigoza 

I basically need to send and receive messages on all my local ubuntu machines (1 or more). My application needs to save the responses from the FC and send it messages so it can alter position and get its current position and so on. So I think I can use MAV_COMP_ID_ALL.  I assume I should be able to send and receive the MAVLink messages on all my systems using it.  Is that correct ?

I do not see a specific component id and system id for the HEARTBEAT msg in px4-offboard.cpp or heartbeat.cpp

zehortigoza:
I mean when you send a a HEARTBEAT, mavlink_msg_heartbeat_encode(SYSTEM_ID, MAV_COMP_ID_ALL, &msg, &heartbeat);

PR #115, opened 2017-08-17
RTPS changes

avinash-palleti:
Rebased PR #98


src/rtps/UART_node.cxx
```diff
@@ -0,0 +1,318 @@
...
+    if (packet_len > buflen) {
+        return -EMSGSIZE;
+    }
+
+    // Found a whole message, send it and remove from local _buf
+    memmove(buffer, _buf + i, packet_len);
+    memmove(_buf + i, _buf + i + packet_len, sizeof(_buf) - i - packet_len);
+    _buf_size -= packet_len;
+
+    return packet_len;
```
lucasdemarchi: This looks like a lot of repetition from code we have in the repo. And it's actually from earlier versions of it.
Besides that, I'm surprised by the copyright on this file since it looks so similar to what we had on our repo already.  Maybe license was wrongly copied and pasted?

src/rtps/UART_node.cxx
```diff
@@ -0,0 +1,318 @@
...
+
+    uint16_t crc = crc16(_buf + start + sizeof(struct Header), len);
+    uint16_t read_crc = ((uint16_t)header->crc_h << 8) | header->crc_l;
+    if (crc != read_crc)
+        log_debug("CRC mismatch %u - %u (%u) %d", crc, read_crc, len, header->topic_ID);
+    else
+        memmove(buffer, _buf + start + sizeof(struct Header), len);
+
+    memmove(_buf + start, _buf + start + sizeof(struct Header) + len,
+            sizeof(_buf) - start - sizeof(struct Header) - len);
```
lucasdemarchi: As I said, it's so similar that it has the same shortcomings of use a small buffers and using a memmove here :(.

src/rtps/main.cpp
```diff
@@ -0,0 +1,289 @@
...
+    // TODO use mavlink-router mainloop (which need to be extracted from it)
+    do {
+        if (_options.report_stats && !measuring) {
+            clock_gettime(0, &begin);
+            measuring = true;
+        }
+
+        // TODO check if it's possible to add RTPS subscribers on epoll
+        // or use a thread for them
+        int r = epoll_wait(epollfd, events, 8, USLEEP_TIME);
```
lucasdemarchi: why USLEEP_TIME? And worse.... why is it 500 usec? You will just wakeup a lot the mainloop to do nothing.

edersondisouza: No reason at all, but reminiscence of a code that strove for throughput. Point is, I have no idea of a good value here.

Makefile.am
```diff
@@ -157,10 +156,55 @@ mavlink_routerd_SOURCES = \
...
+
+if RTPS
+bin_PROGRAMS += rtpsd
+rtpsd_SOURCES = \
+	src/common/log.cpp \
+	src/common/log.h \
+	src/common/util.c \
+	src/common/util.h \
+	src/rtps/main.cpp \
+	src/rtps/RtpsTopics.cxx \
```
edersondisouza: Yep - but they are generated on PX4 tree.

src/rtps/UART_node.cxx
edersondisouza: Ha. This file is 'generated' [1] from PX4 tree when creating the RTPS bridge there. So, long story short, there was an UART_node there, and I've made changes to it - that's why it's similar, but has a different licence. It came from PX4 tree tools. 

[1] Actually is a static file there copied to where you want the RTPS bridge to be created.

src/rtps/UART_node.cxx
edersondisouza: =D

src/rtps/main.cpp
```diff
@@ -0,0 +1,289 @@
...
+    char topic_ID = 255;
+
+    struct timespec begin;
+    bool measuring = false;
+    uint8_t seq;
+
+    struct epoll_event events[8];
+    char data_buffer[BUFFER_SIZE] = {};
+
+    // TODO use mavlink-router mainloop (which need to be extracted from it)
```
edersondisouza: I'll give @avinash-palleti more details about this.

src/rtps/main.cpp
lucasdemarchi: You kill throughout when you keep waking to do nothing.  This timeout should be used to avoid locking the main loop when something really bad happens, not to wakeup periodically. And if you are not handling the case in which you wakeup and have no FD ready, there's no point to wakeup: it should rather be a blocking wait.

Makefile.am
lucasdemarchi: Then it's worse as compatibility will be tied to a specific commit on the px4 tree. The tool to generate these files should probably be extracted in a way that it can be used on both sides of the bridge.  And the tool should probably be improved (if it wasn't already) as I'm seeing a lot of verbose code generated here

src/rtps/UART_node.cxx
lucasdemarchi: Don't change generated files. Never. It will only tie your hands when you want to improve/update the code. Not even commiting the generated files (as I am requesting here) but doing it right from beginning avoids the temptation of doing the simple changes here and there that will chase you later.

Do it right so you do it once.

edersondisouza: Nope, it's not regenerated, I've changed the 'root' one.

Makefile.am
edersondisouza: Yep, generated code has lots of room to improvement.

src/rtps/main.cpp
edersondisouza: Maybe now it can be a blocking one. When I've wrote it, it was necessary to wake up so the subscribing of RTPS would happen - see TODO above.
I remember that they were approaching a threaded scheme for this - @avinash-palleti do you know if they implemented it?

Makefile.am
avinash-palleti: > The tool to generate these files should probably be extracted in a way that it can be used on both sides of the bridge.

Actually tool is generating code for both sides of bridge (client and agent). But we are not using micrortps_agent as is. We took couple of files from there and used in mavlink-router.
@edersondisouza any specific reason that we cant use micrortps_agent as is?

src/rtps/main.cpp
avinash-palleti: @edersondisouza AFAIK it is not yet implemented.

edersondisouza:
On Fri, 2017-08-18 at 03:19 +0000, avinash-palleti wrote:
> @avinash-palleti commented on this pull request.
> In Makefile.am:
> > @@ -157,10 +156,55 @@ mavlink_routerd_SOURCES = \
>  	src/common/xtermios.cpp \
>  	src/common/xtermios.h
>  
> +mavlink_routerd_CPPFLAGS = ${AM_CPPFLAGS}
> +mavlink_routerd_CPPFLAGS += \
> +	-I$(top_srcdir)/src/mavlink-router \
> +	-I$(top_builddir)/include/mavlink \
> +	-I$(top_builddir)/include/mavlink/ardupilotmega
> +
> +if RTPS
> +bin_PROGRAMS += rtpsd
> +rtpsd_SOURCES = \
> +	src/common/log.cpp \
> +	src/common/log.h \
> +	src/common/util.c \
> +	src/common/util.h \
> +	src/rtps/main.cpp \
> +	src/rtps/RtpsTopics.cxx \
> > The tool to generate these files should probably be extracted in a way that it can be used on both sides of the bridge.
> 
> Actually tool is generating code for both sides of bridge (client and agent). But we are not using micrortps_agent as is. We took couple of files from there and used in mavlink-router.
> @edersondisouza any specific reason that we cant use micrortps_agent as is?

Basically, to split RTPS from Mavlink packets. We send both via UART.

avinash-palleti:
Submitted another commit as there are changes in auto-generated code from PX4 side.
As @lucasdemarchi mentioned earlier, we need to figure out a way to fetch these auto generated instead of checking in. 

There is an issue on PX4 repo https://github.com/PX4/Firmware/issues/7728, once this is addressed there will be an independent repo which we can use to generate these files.

And also rtps_client (on PX4) side was receiving continuous messages because of #107. I took #113 fix and it worked.

anitha-suresh:
@lucasdemarchi @edersondisouza , waiting for your review.

lucasdemarchi:
Auto-generated code has no place in the tree... it only leads to commits like your last one to update the code when the tool/spec changes. Please rework so generated code can be generated at compile time, like the mavlink headers are.

avinash-palleti:
Working with PX4 team to get separate repo for generated code, till that is done let the PR be open for our reference.

lucasdemarchi:
Why a repo for generated code? What's wrong with generating  it in build-time, like we do for mavlink headers?

avinash-palleti:
generator script is in Firmware repo. You mean we will have our copy of that script in our repo to generate build time?

lucasdemarchi:
We could. Better than handling copies of the generated source.

lucasdemarchi:
No progress since a long time, let's close the stale PR.

PR #118, opened 2017-09-11
endpoint: don't send msg back to the same endpoint

lucasdemarchi:
If an endpoint is sending a packet through a channel and there are more
components IDs on the same endpoint, mavlink-router would send back any
message through the same channel.  It's never the expected behavior to
receive the message reflected back and there's no way to the other end
to deal with it. Exception would be of course check the destination
sysid/compid, but those may not be present.

This also changes the behavior if the same vehicle/component is
connected through multiple endpoints on mavlink-router, each of them
having more components. mavlink-router will now drop the packet do avoid
loops.

This is an alternative to the fix provided in
https://github.com/01org/mavlink-router/pull/113 and should fix the
issue in https://github.com/01org/mavlink-router/issues/107.

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -320,8 +320,9 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         }
     }
 
-    // This endpoint sent the message, and there's no other sys_comp_id: reject msg
-    if (has_sys_comp_id(src_sysid, src_compid) && _sys_comp_ids.size() == 1)
+    // This endpoint sent the message, we don't want to send it back over the
+    // same channel to avoid loops: reject
+    if (has_sys_comp_id(src_sysid, src_compid))
         return false;
 
     // Message is broadcast on sysid: accept msg


lucasdemarchi:
@NaterGator @edersondisouza @zehortigoza

src/mavlink-router/endpoint.cpp
```diff
@@ -320,8 +320,8 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         }
     }
 
-    // This endpoint sent the message, and there's no other sys_comp_id: reject msg
-    if (has_sys_comp_id(src_sysid, src_compid) && _sys_comp_ids.size() == 1)
+    // This endpoint sent the message
+    if (has_sys_comp_id(src_sysid, src_compid))
```
avinash-palleti: As you mentioned in description it will drop packet though there are other component IDs connected to same endpoint. @edersondisouza mentioned in #113 that he observed a scenario where different components are connected to same end point while working on ROS.

lucasdemarchi: It's ok for different components to be connected to the same endpoint. This is normal. What's not normal is that these components depend on mavlink-router to work.

Examples:

```
GCS -> Flight Controller   -> mavlink-router -> App
       Flight Controller -/
```

```
GCS -> Flight Controller   -> mavlink-router -> mavros -> App1
       Flight Controller -/                            \-> App2
```

In the first case, the flight stack on the flight controller needs to check if it should handle the message or if it should forward it. Both ArduPilot and PX4 already do that

On the second case, App1 and App2 should not be talking to each other via mavlink-router. They should either a) be communicating directing (alas they are ROS applications and should keep the mavlink serialization and de-serialization out of the game) or b) make mavros route the packet. There's no way for mavlink-router to do the right thing if it sends a (possibly broadcast message) over to the same endpoint that sent it.

zehortigoza:
LGTM

src/mavlink-router/endpoint.cpp
edersondisouza: I think it's fine, then.

src/mavlink-router/endpoint.cpp
```diff
@@ -320,8 +320,8 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         }
     }
 
-    // This endpoint sent the message, and there's no other sys_comp_id: reject msg
-    if (has_sys_comp_id(src_sysid, src_compid) && _sys_comp_ids.size() == 1)
+    // This endpoint sent the message
```
edersondisouza: Minor nitpick: I'd just state on the comment that desired outcome is to reject the message.

lucasdemarchi:
I expanded the commit message with the examples above and added a comment about rejecting the message.

PR #123, opened 2017-11-07
examples: fix leak on error path

lucasdemarchi:
We would leak ip when we `goto fail`.  Always allocate the small string
so we can always call free on it.

### DIFF
--- a/examples/px4-offboard-mode.cpp
+++ b/examples/px4-offboard-mode.cpp
@@ -649,7 +649,7 @@ static int setup_timeout()
 
 int main(int argc, char *argv[])
 {
-    char *ip = (char *)"127.0.0.1";
+    char *ip = strdup("127.0.0.1");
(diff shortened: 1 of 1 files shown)


examples/px4-offboard-mode.cpp
```diff
@@ -649,7 +649,7 @@ static int setup_timeout()
 
 int main(int argc, char *argv[])
 {
-    char *ip = (char *)"127.0.0.1";
+    char *ip = strdup("127.0.0.1");
     int port = 5760;
 
     if (argc > 1) {
```
zehortigoza: This leak memory every time user provides the IP:port.
I will squash the fix and merge.

lucasdemarchi: thanks

zehortigoza:
thanks

PR #130, opened 2018-04-17
Running -> Routing

shakthi-prashanth-m:
Replace typo **Running** to **Routing**


avinash-palleti:
@shakthi-prashanth-m I think it is not typo. 
Those are steps to build, install and run this utility. So the title for the section looks apt.

PR #136, opened 2018-08-08
Logging: set file to read-only & add logging mode option

bkueng:
This PR includes 2 changes:
- change the log file permissions to read-only after the file is closed. This can be used by other clients to detect when logging is finished
- add an option `LogMode` that can be set to `arm_until_disarm`. In that case logging is started when the vehicle is armed and stopped when disarmed. The default is `always`, which is the same behavior as before.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -104,8 +104,7 @@
 #Mavlink-router serves on this TCP port
 TcpServerPort=5790
 ReportStats=false
-#Which mavlink dialect is being used. Can be `common`[default] or `ardupilotmega`
-MavlinkDialect=ardupilotmega
+MavlinkDialect=auto
 
 [UdpEndpoint alfa]
 Mode = Eavesdropping
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -31,6 +31,12 @@
 #       Path to directory where to store flightstack log.
 #       No default value. If absent, no flightstack log will be stored.
 #
+#   LogMode
+#       One of <always>, <while_armed>
+#       Default: always, log from start until mavlink-router is stopped.
+#       If set to while_armed, a new log file is created whenever the vehicle is
+#       armed, and closed when disarmed.
+#
 #   DebugLogLevel
 #       One of <error>, <warning>, <info> or <debug>. Which debug log
 #       level is being used by mavlink-router, with <debug> being the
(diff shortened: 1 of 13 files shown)


examples/config.sample
```diff
@@ -31,6 +31,11 @@
 #       Path to directory where to store flightstack log.
 #       No default value. If absent, no flightstack log will be stored.
 #
+#   LogMode
+#       One of <always>, <arm_until_disarm>
```
lucasdemarchi: always / while-armed would be better names imo

examples/config.sample
```diff
@@ -31,6 +31,11 @@
...
+#       Default: always. If absent, log from start until mavlink-router is
+#       stopped.
```
lucasdemarchi: also document the other one, mentioning the log file will be open when the vehicle is armed and closed when disarmed, creating a new log file.

examples/config.sample
```diff
@@ -98,8 +103,7 @@
 #Mavlink-router serves on this TCP port
 TcpServerPort=5790
 ReportStats=false
-#Which mavlink dialect is being used. Can be `common`[default] or `ardupilotmega`
-MavlinkDialect=ardupilotmega
+MavlinkDialect=auto
```
lucasdemarchi: unrelated. Either a separate commit or at least mention in the commit message that while changing this you fixed the default value here.

src/mavlink-router/binlog.cpp
```diff
@@ -75,22 +75,31 @@ int BinLog::write_msg(const struct buffer *buffer)
...
         struct mavlink_router_mavlink1_header *msg
             = (struct mavlink_router_mavlink1_header *)buffer->data;
         msg_id = msg->msgid;
+        payload = buffer->data + sizeof(struct mavlink_router_mavlink1_header);
         payload_len = msg->payload_len;
+        source_system_id = msg->sysid;
     }
 
+    /* Check if we should start or stop logging */
+    _handle_auto_start_stop(msg_id, source_system_id, payload);
```
lucasdemarchi: I don't like how this is a special case and the routing to the logging endpoint is plugged in the middle of this function. What's the problem with letting it inside the logendpoint itself implementing the normal pattern like other endpoints with accept_msg() + write()?

bkueng: I agree it's not that nice. But from what I see `accept_msg` cannot filter by `source_system_id` which is what we want here.

bkueng:
Thanks for the review, I addressed all comments except for `_handle_auto_start_stop`.

lucasdemarchi:
while_armed should be while-armed, but this can be polished on top.

lucasdemarchi:
Applied, thanks.

PR #138, opened 2018-09-07
logging: use dash on config entry

lucasdemarchi:
We don't use underscores in config entries: a dash should rather be
used.


lucasdemarchi:
@bkueng ping

PR #141, opened 2018-09-11
Revert "endpoint: set RTS/DTR state"

### LINKED ISSUE #128: Mavlink router tries to configure DTR/RTS and fails
I am trying to get an RK3328 based board to run mavlink router and using the `/dev/ttyFIQ0` UART.

I have been able to test the UART with `cat` and `echo` to confirm data is send and received ok. This much is working. When I start mavlink router I get an error 

    Could not set DTR/RTS (Invalid argument)

and I cannot proceed.

Is there a way of starting this or does it require code changes to stop this from failing for my use case?

Many thanks,
Ed

lucasdemarchi:
This reverts commit 940f0b8caf3cf61879c3cdfebd66a78eb127bd70.

I think this was probably one artficact of putting the uart in raw mode
rather than sane mode. We don't care about RTS/DTR and none of the
flight stacks we support are setting those.

Fix #128, but needs testing

### DIFF
--- a/examples/receiver.py
+++ b/examples/receiver.py
@@ -17,7 +17,6 @@
 # limitations under the License.
 
 from __future__ import print_function
-from __future__ import print_function
 
 import pymavlink.mavutil as mavutil
 import sys
--- a/examples/sender.py
+++ b/examples/sender.py
@@ -17,7 +17,6 @@
 # limitations under the License.
 
 from __future__ import print_function
-from __future__ import print_function
 from threading import Thread
 from time import sleep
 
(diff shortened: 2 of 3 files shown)


lucasdemarchi:
And at best this should be under the flowcontrol function.

lucasdemarchi:
44ff6aa16c26574cfe55941e5cd55975e4562a53

I don't think we should even try to mess with DTR - not even clear it because we risk it not being supported - if we clear it, then we should ignore errors. No board that I know of uses it. I checked ardupilot, PX4, mavros and mavproxy. Nobody uses those.

The best explanation I can give about why I added it is the commits around that change. I should have recorded better commit messages.

lucasdemarchi:
This is not compatible with a fake uart (pts) reated by socat. I think we should go ahead and remove it

PR #142, opened 2018-09-14
logging: allow autopilot logging from an arbitrary system id

bkueng:
The expected system id for log streaming was hardcoded to 1, and thus
logging would not work when the autpilot is configured to have a different
sys id.

This patch changes the behavior to pick up the system id from the heartbeat
of the autopilot.

### DIFF
--- a/src/mavlink-router/autolog.cpp
+++ b/src/mavlink-router/autolog.cpp
@@ -32,6 +32,7 @@ int AutoLog::write_msg(const struct buffer *buffer)
     uint32_t msg_id;
     uint8_t *payload;
     uint8_t source_system_id;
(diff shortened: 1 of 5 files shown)


bkueng:
> Coding style is wrong. Otherwise LGTM

Fixed. Can we check that with CI?

lucasdemarchi:
> Fixed. Can we check that with CI?

We have tools/checkpatch for that. I remember in the past the github API didn't have a way to warn that something is wrong but without giving an error (without adding a bot to do that rather than using the CI API). Since there are false positives I never hooked that up.

It may be better now, dunno. Patches welcome for that.

Issue #39, opened 2017-03-17
Allow some Endpoints to listen to messages based on a filter


edersondisouza:
Filtering would be made on routing function - it would allow an endpoint to even listen all messages, even those not meant for it.
A practical example would be LogEndpoints: a filter would avoid then receiving any messages not meant for logging.

Jaeyoung-Lim:
Is there any progress on this?

lucasdemarchi:
This is a feature idea we had a long time ago. It would be very welcome, but AFAIK no one is working on ti.

dk7xe:
Filtering would be really appreaciated on message id.
e.g. with the implementation of UTM_GLOBAL_POSITION message https://github.com/mavlink/mavlink/pull/1009 we have a message to report vehicle positions to air traffic management / UTM systems.
Currently all mavlink messages are forwarded to the UTM system which causes a lot of overhead.
Therefore filtering only needed messages in such a case would reduce the traffic a lot.

Jaeyoung-Lim:
@dk7xe This has been implemented on #167

dk7xe:
@Jaeyoung-Lim thank you!

PR #143, opened 2018-09-18
Cleanup mavlink header parsing

lucasdemarchi:
The parsing of mavlink header was intended to be contained in a single function, when we are reading the message from the associated buffer. Due to the addition of logging a long time ago and continuous use of the header fields to properly route the packets to the appropriate endpoint we ended up adding the same parsing to the log endpoints, including the check for different struct on mavlink1 vs mavlink2 wire format.

This is a cleanup aiming to stop doing that and rather just adding the appropriate fields in `struct buffer`. This also makes it much closer to be able to implement #39 since now we keep the msgid and could check for it against a set of msg ids in the correspondent `accept_msg()`.

This still needs to go through proper test and bench test, but apparently it's working (hey, it passes the testsuite).

### DIFF
--- a/examples/heartbeat-print.cpp
+++ b/examples/heartbeat-print.cpp
@@ -37,7 +37,7 @@ static void exit_signal_handler(int signum)
 
 static void setup_signal_handlers()
 {
-    struct sigaction sa = { };
+    struct sigaction sa = {};
 
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
@@ -53,8 +53,7 @@ static void handle_new_message(const mavlink_message_t *msg)
         printf("HEARTBEAT:\n"
                "\tmavlink_version: %u\n"
                "\ttype: %u\n",
-               heartbeat.mavlink_version,
-               heartbeat.type);
+               heartbeat.mavlink_version, heartbeat.type);
     }
 }
 
@@ -115,7 +114,6 @@ int main(int argc, char *argv[])
                 handle_new_message(&msg);
             }
         }
-
     }
 
     free(ip);
--- a/src/autolog.cpp
+++ b/src/autolog.cpp
@@ -28,37 +28,18 @@ int AutoLog::write_msg(const struct buffer *buffer)
         return _logger->write_msg(buffer);
     }
 
-    const bool mavlink2 = buffer->data[0] == MAVLINK_STX;
-    uint32_t msg_id;
-    uint8_t *payload;
-    uint8_t source_system_id;
-    uint8_t source_component_id;
-
-    if (mavlink2) {
-        auto *hdr = (struct mavlink_router_mavlink2_header *)buffer->data;
-        msg_id = hdr->msgid;
-        payload = buffer->data + sizeof(*hdr);
-        source_system_id = hdr->sysid;
-        source_component_id = hdr->compid;
-    } else {
-        auto *hdr = (struct mavlink_router_mavlink1_header *)buffer->data;
-        msg_id = hdr->msgid;
-        payload = buffer->data + sizeof(*hdr);
-        source_system_id = hdr->sysid;
-        source_component_id = hdr->compid;
-    }
-
     /* set the expected system id to the first autopilot that we get a heartbeat from */
-    if (_target_system_id == -1 && msg_id == MAVLINK_MSG_ID_HEARTBEAT
-        && source_component_id == MAV_COMP_ID_AUTOPILOT1) {
-        _target_system_id = source_system_id;
+    if (_target_system_id == -1 && buffer->curr.msg_id == MAVLINK_MSG_ID_HEARTBEAT
+        && buffer->curr.src_compid == MAV_COMP_ID_AUTOPILOT1) {
+        _target_system_id = buffer->curr.src_sysid;
     }
 
-    if (msg_id != MAVLINK_MSG_ID_HEARTBEAT || source_system_id != _target_system_id) {
+    if (buffer->curr.msg_id != MAVLINK_MSG_ID_HEARTBEAT
+        || buffer->curr.src_sysid != _target_system_id) {
         return buffer->len;
     }
 
-    const mavlink_heartbeat_t *heartbeat = (mavlink_heartbeat_t *)payload;
+    const mavlink_heartbeat_t *heartbeat = (mavlink_heartbeat_t *)buffer->curr.payload;
 
     /* We check autopilot on heartbeat */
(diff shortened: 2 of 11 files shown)


lucasdemarchi:
humn... CI fails due to old gcc. I'll probably have to change the designated initializers

lucasdemarchi:
@bkueng @zehortigoza  ping

src/mavlink-router/endpoint.cpp
```diff
@@ -172,11 +172,11 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
...
 
         msg_id = hdr->msgid;
-        payload = rx_buf.data + sizeof(*hdr);
+        pbuf->payload = payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
-        *src_sysid = hdr->sysid;
-        *src_compid = hdr->compid;
-        payload_len = hdr->payload_len;
+        pbuf->src_sysid = *src_sysid = hdr->sysid;
+        pbuf->src_sysid = *src_compid = hdr->compid;
```
zehortigoza: src_compid

src/mavlink-router/endpoint.cpp
```diff
@@ -231,29 +231,29 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
     _stat.read.handled_bytes += expected_size;
 
     if (!_crc_check_enabled || msg_entry) {
-        _add_sys_comp_id(((uint16_t)*src_sysid << 8) | *src_compid);
+        _add_sys_comp_id(((uint16_t)pbuf->src_sysid << 8) | pbuf->src_compid);
```
zehortigoza: nitpick: you could leave the line deleted

src/mavlink-router/endpoint.cpp
```diff
@@ -191,11 +191,11 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
             return 0;
 
         msg_id = hdr->msgid;
-        payload = rx_buf.data + sizeof(*hdr);
+        pbuf->payload = payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
-        *src_sysid = hdr->sysid;
-        *src_compid = hdr->compid;
-        payload_len = hdr->payload_len;
+        pbuf->src_sysid = *src_sysid = hdr->sysid;
```
zehortigoza: I did not checked if you are doing it in the further patches(as I will review the other patches other I want to leave this comment here) but with this fields in buffer we could remove then from read_msg() and make this:
pbuf->src_sysid = *src_sysid = hdr->sysid; to this: pbuf->src_sysid = hdr->sysid;

src/mavlink-router/endpoint.cpp
```diff
@@ -68,19 +68,18 @@ bool Endpoint::handle_canwrite()
...
 {
-    int target_sysid, target_compid, r;
-    uint8_t src_sysid, src_compid;
+    int r;
     struct buffer buf{};
 
-    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid)) > 0)
-        Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid,
-                                           src_compid);
+    while ((r = read_msg(&buf)) > 0)
```
zehortigoza: aaah cool you are removing here and in the next patch

src/mavlink-router/endpoint.cpp
```diff
@@ -174,8 +173,8 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
         msg_id = hdr->msgid;
         pbuf->payload = payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
-        pbuf->src_sysid = *src_sysid = hdr->sysid;
-        pbuf->src_sysid = *src_compid = hdr->compid;
+        pbuf->src_sysid = hdr->sysid;
+        pbuf->src_sysid = hdr->compid;
```
zehortigoza: src_compid

src/mavlink-router/endpoint.cpp
```diff
@@ -278,8 +278,10 @@ int Endpoint::read_msg(struct buffer *pbuf)
     return msg_entry != nullptr ? ReadOk : ReadUnkownMsg;
 }
 
-void Endpoint::_add_sys_comp_id(uint16_t sys_comp_id)
+void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
 {
+    uint16_t sys_comp_id = sysid << 8 | compid;
```
zehortigoza: ++ for this change

zehortigoza:
I just need to review with more calm 'logendpoint: move checks from write_msg to accept_msg()' other patches looks good to me, with the s/src_sysid/src_compid pointed out.

src/mavlink-router/endpoint.cpp
lucasdemarchi: As you noticed, yes I'm doing in a following patch, otherwise there would be a single giant patch that's hard to review and validate

src/mavlink-router/endpoint.cpp
lucasdemarchi: Ugh, I wonder if it became wrong when I rebased or if it was already wrong... It should not have passed the testsuite like this

lucasdemarchi:
Yep, logging stopped working, I need to debug that. At least it triggered my fixes on #145. Thanks @bkueng

Jaeyoung-Lim:
Are there any updates on this?

src/mavlink-router/comm.h
```diff
@@ -27,4 +27,13 @@
...
     uint8_t *data;
+    uint8_t *payload;
+
+    /* Data relevant for the last parsed msg available on this buffer */
+    uint32_t msgid;
+    int target_sysid;
+    int target_compid;
+    uint8_t src_sysid;
+    uint8_t src_compid;
+    uint8_t payload_len;
```
lucasdemarchi, line 38: probably a good idea to move this to a inner `last` struct

lucasdemarchi (commented):
This needs to be re-implemented by taking a look at the ideas and try again

lucasdemarchi:
I'm dropping this commit: "logendpoint: move checks from write_msg to accept_msg()". The rest is re-implemented on current master branch.

jbeyerstedt:
Reducing the number of method arguments is always appreciated. Looks quite nice now.

I'm not quite happy with the name "buffer", but also have no better name for it right now.

PR #151, opened 2018-12-12
mainloop: handle epoll_wait error

bkueng:
epoll can fail for example when an UART FTDI cable is plugged out and the
associated device becomes invalid. In that case we remove the file
descriptor from the list.

We also quit mavlink-router, because in most cases the error is fatal,
and an external component (e.g. systemd) can restart it.
If this behavior is not desired by default, we can make it a config option.

The current behavior is that mavlink-router busy-loops in case the cable is plugged out.

### DIFF
--- a/src/mavlink-router/mainloop.cpp
+++ b/src/mavlink-router/mainloop.cpp
@@ -288,6 +288,13 @@ void Mainloop::loop()
                     mod_fd(p->fd, p, EPOLLIN);
                 }
             }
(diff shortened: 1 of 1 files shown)


bkueng:
@lucasdemarchi ping?

lucasdemarchi:
If this is tested with udp/tcp, LGTM. Feel free to merge it, I just gave you write access.

bkueng:
Thanks.

PR #152, opened 2018-12-20
add support for musl

JonasVautherin:
Following the Gentoo patch here: https://cgit.gentoo.org/proj/musl.git/tree/sys-libs/efivar/files/27-strndupa.patch?id=0d259ad72bb04c5a9bc1c83d3

This enables build on musl-based systems (e.g. Alpine Linux).

### DIFF
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -56,3 +56,14 @@ int mkdir_p(const char *path, int len, mode_t mode);
 #ifdef __cplusplus
 }
 #endif
+
+#ifndef strndupa
+#define strndupa(s, n) \
+       (__extension__ ({const char *__in = (s); \
+                        size_t __len = strnlen (__in, (n)); \
+                        char *__out = (char *) alloca (__len + 1); \
+                        __out[__len] = '\0'; \
+                        (char *) memcpy (__out, __in, __len);}))
+
+#include <asm/ioctls.h>
+#endif


src/common/util.h
```diff
@@ -56,3 +56,12 @@ int mkdir_p(const char *path, int len, mode_t mode);
...
 }
 #endif
+
+#ifndef strndupa
+#define strndupa(s, n) \
+       (__extension__ ({const char *__in = (s); \
+                        size_t __len = strnlen (__in, (n)) + 1; \
+                        char *__out = (char *) alloca (__len); \
+                        __out[__len-1] = '\0'; \
+                        (char *) memcpy (__out, __in, __len-1);}))
```
lucasdemarchi: I'd rather remove the `+ 1` and just add it into the place needed: the call to alloca.

src/mavlink-router/endpoint.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include <arpa/inet.h>
 #include <assert.h>
+#include <asm/ioctls.h>
```
lucasdemarchi: Why? If it's needed it should rather be inside the ifndef in util.h

JonasVautherin:
@lucasdemarchi: what do you think of it now? :blush:

PR #157, opened 2019-02-22
Fixed build for Alpine Linux

MatejFranceskin:
To make mavkink-router build on minimalistic Alpine linux distribution I had to:
- replace strndupa with strndup. stdndupa uses alloca and is not portable. Alloca() allocates memory within the current function's stack frame, so when replacing with strndup we have to call free manually. There were some places where alloca was incorrectly used
- If TCSETS2 or TCGETS2 are not defined then use TCSETS and TCGETS

### DIFF
--- a/src/common/conf_file.cpp
+++ b/src/common/conf_file.cpp
@@ -515,8 +515,10 @@ int ConfFile::parse_bool(const char *val, size_t val_len, void *storage, size_t 
         if (storage_len < sizeof(_type))                                        \
             return -ENOBUFS;                                                    \
                                                                                 \
-        str = strndupa(val, val_len);                                           \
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
> replace strndupa with strndup

We already have our own definition of strndupa of it's not defined. Why is it not working for you?

> TCSETS2 or TCGETS2 are not defined then use TCSETS and TCGETS

This breaks cases in which we want rounded nauseated like 500k / 1.5M. TC[GS]ET doesn't work with those.

MatejFranceskin:
Sorry - I see that strndupa problem was solved 2 months ago and I didn't use the latest from master.
I will close this.

PR #161, opened 2019-03-28
Implemented proper stopping of ulog streaming

RomanBapst:
Implemented stop timeout for ulog streaming. Previously it could happen that the message for stopping ulog streaming got lost and therefore the autopilot never stopped streaming the log file.
The next time the mavlink router requested ulog streaming it was already running and thus needed to be restarted due to the missing ulog header. This caused several empty log files to be created.

With these changes the mavlink router waits for the autopilot ACK before considering the stream to be stopped. In case the router is shut down with ulog streaming still running it will try to stop the stream for 5 seconds, if unsuccessful after that it will shut down regardless of the stream still running.

### DIFF
--- a/src/mavlink-router/autolog.h
+++ b/src/mavlink-router/autolog.h
@@ -43,6 +43,7 @@ protected:
     // These functions should never be called
     const char *_get_logfile_extension() override { return ""; };
     bool _start_timeout() override { return true; };
+    bool _stop_timeout() override { return true; };
 
 private:
     std::unique_ptr<LogEndpoint> _logger;
--- a/src/mavlink-router/binlog.cpp
+++ b/src/mavlink-router/binlog.cpp
@@ -39,6 +39,14 @@ bool BinLog::_start_timeout()
     return true;
 }
 
+bool BinLog::_stop_timeout()
+{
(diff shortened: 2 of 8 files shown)


lucasdemarchi:
@bkueng @RomanBapst any update here?

RomanBapst:
> This seems to be implemented backwards with call chains like ULog -> LogEndpoint -> ULog. That should not happen... it only makes it more complex and there's no reason not to have the timeout on binlog as well.

> ULog::stop() should do thing that are specific to ulog and call the parent method to do things that are independent of logging implementation. Setting up a timer is independent of logging method.

@lucasdemarchi I can move the code executed in _stop_timeout() into the ULog::stop() method.
That way the call chain will be ULog -> LogEndpoint

lucasdemarchi:
@RomanBapst  are you still interested in having this merged?

lucasdemarchi:
@RomanBapst  @bkueng  ping?

bkueng:
I'm generally interested to get our changes back in sync with upstream, but I currently don't have the time. I'll ask internally, but feel free to close this for now.

lucasdemarchi:
@bkueng  ok, thanks

PR #165, opened 2019-04-11
reducing poll latency with ioctl serial control

catch-twenty-two:
This pr fixes the latency on the serial port that is only noticeable when communicating at high speeds.

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -34,6 +34,8 @@
 #include <common/util.h>
 #include <common/xtermios.h>
 
+#include <linux/serial.h>
+
 #include "mainloop.h"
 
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Tested on HW.

lucasdemarchi:
Tested on *which* hardware is the important part missing here. I suppose it's an ftdi converter?

Please add such details to the commit message.

jkflying:
I tested this on FTDI on an Up Core board, and it noticeably reduced latency.

Also of interest, MAVROS uses this same technique if you give it the serial port directly:
https://github.com/mavlink/mavros/blob/7c8b06ba17c9912628cd7c64f1f238e989863418/libmavconn/src/serial.cpp#L100

catch-twenty-two:
Tested it with the pocket beagle serial port which doesn't have a USB->Serial connection works fine, also tested with my FDTI adaptor.

PR #167, opened 2019-05-03
Enable message filtering to reduce network traffic.

MatejFranceskin:
Add Filter field to endpoint configuration to limit the messages that are sent to particular endpoint and improve performance.
For example XYZ is interested only in messages 24,33,76,320,321,323.
This configuration filters all other messages so that they're not sent to XYZ and don't need to be parsed.

[UdpEndpoint XYZ]
Mode = Normal
Address = 127.0.0.1
Port = 14530
Filter = 24,33,76,320,321,323

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -74,7 +74,7 @@ int Endpoint::handle_read()
     uint32_t msg_id;
     struct buffer buf{};
 
-    if ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id)) > 0)
+    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id)) > 0)
         Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid,
                                            src_compid, msg_id);
 
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -70,20 +70,20 @@ int Endpoint::handle_read()
 {
     int target_sysid, target_compid, r;
     uint8_t src_sysid, src_compid;
+    uint32_t msg_id;
     struct buffer buf{};
 
-    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid)) > 0)
+    if ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id)) > 0)
         Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid,
-                                           src_compid);
+                                           src_compid, msg_id);
 
     return r;
 }
 
 int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compid,
-                       uint8_t *src_sysid, uint8_t *src_compid)
+                       uint8_t *src_sysid, uint8_t *src_compid, uint32_t *msg_id)
 {
     bool should_read_more = true;
-    uint32_t msg_id;
     const mavlink_msg_entry_t *msg_entry;
     uint8_t *payload, seq, payload_len;
 
@@ -171,7 +171,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
         if (rx_buf.len < sizeof(*hdr))
             return 0;
 
-        msg_id = hdr->msgid;
+        *msg_id = hdr->msgid;
         payload = rx_buf.data + sizeof(*hdr);
         seq = hdr->seq;
         *src_sysid = hdr->sysid;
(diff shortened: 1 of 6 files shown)


src/mavlink-router/endpoint.cpp
```diff
@@ -70,20 +70,20 @@ int Endpoint::handle_read()
 {
     int target_sysid, target_compid, r;
     uint8_t src_sysid, src_compid;
+    uint32_t msg_id;
     struct buffer buf{};
 
-    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid)) > 0)
+    if ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id)) > 0)
```
lucasdemarchi: should continue to be `while`, or put in a separate commit with justification

src/mavlink-router/endpoint.h
```diff
@@ -131,6 +134,9 @@ class Endpoint : public Pollable {
     const bool _crc_check_enabled;
     uint32_t _incomplete_msgs = 0;
     std::vector<uint16_t> _sys_comp_ids;
+
+private:
+    std::set<uint32_t> message_filter;
```
lucasdemarchi: I'm not a big fan of std::set. I'd rather use std::vector here. When we are filtering we will probably care for very few messages, so std::vector should even be faster for lookups.

src/mavlink-router/main.cpp
```diff
@@ -224,19 +224,24 @@ static int add_endpoint_address(const char *name, size_t name_len, const char *i
...
-        free(conf->address);
         conf->address = strdup(ip);
         if (!conf->address) {
             ret = -ENOMEM;
             goto fail;
         }
-    }
-
-    if (!conf->address) {
+    } else {
```
lucasdemarchi: all these changes in this function seem good, but should be in a separate commit as a cleanup. I think this function got reworked at some time and these checks remained, but don't make sense anymore.

lucasdemarchi: In this one, btw: `ip` is mandatory... so just remove the `if (ip)` and let the check for ENOMEM. It would be good to put a `assert_or_return(ip != nullptr, -EINVAL)` in the beginning of the function.

src/mavlink-router/ulog.cpp
```diff
@@ -191,7 +191,7 @@ int ULog::write_msg(const struct buffer *buffer)
         mavlink_msg_logging_ack_encode(LOG_ENDPOINT_SYSTEM_ID, MAV_COMP_ID_ALL, &msg, &ack);
         _send_msg(&msg, _target_system_id);
         /* message will be handled by MAVLINK_MSG_ID_LOGGING_DATA case */
-        [[gnu::fallthrough]];
+        /* fall through */
```
lucasdemarchi: ?

src/mavlink-router/main.cpp
```diff
@@ -667,11 +672,13 @@ static int parse_confs(ConfFile &conf)
         char *addr;
         bool eavesdropping;
         unsigned long port;
+        char *filter;
     };
     static const ConfFile::OptionsTable option_table_udp[] = {
         {"address", true,   ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_udp, addr)},
         {"mode",    true,   parse_mode,                 OPTIONS_TABLE_STRUCT_FIELD(option_udp, eavesdropping)},
         {"port",    false,  ConfFile::parse_ul,         OPTIONS_TABLE_STRUCT_FIELD(option_udp, port)},
+        {"filter",  false,  ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_udp, filter)},
```
lucasdemarchi: I guess it would be useful to have a `ConfigFile::parse_ul_vector` and make it do the parsing. Then you just std::move<> the parsed vector instead of strdup() the string.

lucasdemarchi: this can be left for later though, as it's more intrusive.

src/mavlink-router/endpoint.h
MatejFranceskin: Replaced std::set with std::vector.

src/mavlink-router/ulog.cpp
MatejFranceskin: This removes compile warning, source:
https://stackoverflow.com/questions/45349079/how-to-use-attribute-fallthrough-correctly-in-gcc

src/mavlink-router/endpoint.cpp
MatejFranceskin: Reverted - will create a new cleanup PR with justification.

src/mavlink-router/main.cpp
MatejFranceskin: Reverted - will create a new cleanup PR.

src/mavlink-router/main.cpp
MatejFranceskin: If you think that it can stay this way then I will not change it for now.

MatejFranceskin:
Thank you for your review. Please take a look at the changes that I made to address the issues.

dk7xe:
Thanx Matej! That really helps! Gerald

PR #168, opened 2019-05-03
PR cleanup

MatejFranceskin:
This PR contains 3 small improvements or code cleanups:
- Fall through compile warning: __attribute__ ((fallthrough)) was introduced in GCC 7. To maintain backward compatibility and clear the fall through warning for both Clang and GCC, you can use the /* fall through */
- main.cpp - since conf is calloc-ed at the beginning it contains only zeroes, so we don't have to check if it is null or even free null pointer.
- while/if read_msg - if first read doesn't read all the available data in the buffers then epoll raises event again, so the control will return to this code until everything is read. If we use while then reading of all file descriptors will be blocked until everything is read which might theoretically be indefinite.


lucasdemarchi:
Why did you close it?

MatejFranceskin:
Sorry - I made a mistake!

lucasdemarchi:
Please update to solve the conflicts

lucasdemarchi:
Please **reabase** do not merge master branch back

MatejFranceskin:
Rebased.

lucasdemarchi:
@MatejFranceskin still not the proper way. The end result needs to be clean, not with fixup commits. We can't have **known** broken commits that are later fixed up. Please take a look on interactive rebase and fix it the proper way.

MatejFranceskin:
I squashed the last 2 commit. It looks better now.

MatejFranceskin:
@lucasdemarchi I rebased it again.

lucasdemarchi:
I think it's pretty common to have more than 1 message. Not sure I like the change to process one at time round-tripping to the kernel on the epoll wait.

If this is in fact a problem, maybe we should have a max time we spend processing data for each fd?

lucasdemarchi:
@MatejFranceskin this would need at least a rebase. Are you still interested in merging this?

lucasdemarchi:
Closing for inactivity. Feel free to open a new one if this is desired.

PR #169, opened 2019-05-14
logendpoint: mark unfinished logs as read-only on startup

bkueng:
If the system or mavlink-router crashed, the last log was not marked as read-only, so we go through all logs on startup and check them.

### DIFF
--- a/src/mavlink-router/logendpoint.cpp
+++ b/src/mavlink-router/logendpoint.cpp
@@ -54,6 +54,35 @@ void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)
     _stat.read.handled_bytes += buffer.len;
 }
 
+void LogEndpoint::mark_unfinished_logs()
+{
+    DIR *dir = opendir(_logs_dir);
+
+    // Assume the directory does not exist if opendir failed
+    if (!dir) {
+        return;
+    }
+
+    struct dirent *ent;
+    uint32_t u;
+
+    while ((ent = readdir(dir)) != nullptr) {
+        if (sscanf(ent->d_name, "%u-", &u) == 1) {
+            char log_file[PATH_MAX];
+            struct stat file_stat;
+            if (snprintf(log_file, sizeof(log_file), "%s/%s", _logs_dir, ent->d_name)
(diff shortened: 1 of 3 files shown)


src/mavlink-router/logendpoint.cpp
```diff
@@ -54,6 +54,35 @@ void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)
...
+    uint32_t u;
+
+    while ((ent = readdir(dir)) != nullptr) {
+        if (sscanf(ent->d_name, "%u-", &u) == 1) {
+            char log_file[PATH_MAX];
+            struct stat file_stat;
+            if (snprintf(log_file, sizeof(log_file), "%s/%s", _logs_dir, ent->d_name)
+                < (int)sizeof(log_file)) {
+                if (!stat(log_file, &file_stat)) {
+                    if (!S_ISDIR(file_stat.st_mode) && (file_stat.st_mode & S_IWUSR)) {
```
lucasdemarchi: should rather check it's a file... and doing return/continue early style is preferred.

bkueng: Fixed.

bkueng:
@lucasdemarchi good to merge?

src/mavlink-router/logendpoint.cpp
```diff
@@ -54,6 +54,38 @@ void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)
...
+            continue;
+
+        if (stat(log_file, &file_stat))
+            continue;
+
+        if (S_ISREG(file_stat.st_mode) && (file_stat.st_mode & S_IWUSR)) {
+            log_info("File %s not read-only yet, marking as RO", ent->d_name);
+            chmod(log_file, S_IRUSR | S_IRGRP | S_IROTH);
+        }
+    }
```
lucasdemarchi: Missing a closedir(). Otherwise lgtm

bkueng: Fixed

lucasdemarchi: Feel free to merge when you want when comments are handled

PR #173, opened 2019-05-31
Add sleep mode for endpoints

dvornikov-aa:
Frequently LTE connection has a limited traffic volume. In this case it's very useful to drop all outcoming MAVLink messages on some endpoints while there are no clients on the other side.


lucasdemarchi:
Any updates here? we have quite a few conflicts

PR #174, opened 2019-05-31
Add UDP broadcast mode

dvornikov-aa:
UDP endpoint updates target IP address from the incoming MAVLink message. This PR adds a new broadcast mode that has a fixed target IP.

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -684,7 +684,7 @@ UdpEndpoint::UdpEndpoint()
     bzero(&sockaddr, sizeof(sockaddr));
 }
 
-int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
+int UdpEndpoint::open(const char *ip, unsigned long port, UdpEndpoint::UdpMode mode)
 {
     const int broadcast_val = 1;
     fd = socket(AF_INET, SOCK_DGRAM, 0);
@@ -697,7 +697,9 @@ int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
     sockaddr.sin_addr.s_addr = inet_addr(ip);
     sockaddr.sin_port = htons(port);
 
-    if (to_bind) {
+    _mode = mode;
+
+    if (_mode == UdpEndpoint::Eavesdropping) {
         if (bind(fd, (struct sockaddr *)&sockaddr, sizeof(sockaddr)) < 0) {
             log_error("Error binding socket (%m)");
             goto fail;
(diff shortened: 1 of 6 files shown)


lucasdemarchi:
The name seems a bit confusing since it may be thought of as the broadcast address. Maybe "Fixed" would be better? 

What's the use case of having it fixed?

src/mavlink-router/endpoint.cpp
```diff
@@ -731,8 +735,12 @@ int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
 ssize_t UdpEndpoint::_read_msg(uint8_t *buf, size_t len)
 {
     socklen_t addrlen = sizeof(sockaddr);
-    ssize_t r = ::recvfrom(fd, buf, len, 0,
-                           (struct sockaddr *)&sockaddr, &addrlen);
+    ssize_t r;
+    if (_mode == UdpEndpoint::Broadcast)
+        r = ::recvfrom(fd, buf, len, 0, NULL, NULL);
+    else
+        r = ::recvfrom(fd, buf, len, 0,
```
lucasdemarchi: This change is not needed. You can just ignore the values later

dvornikov-aa:
I thought "Broadcast" is more obvious (at least you see the purpose of this mode). Actually, I like "Fixed" more. For example, https://github.com/MonashUAS/cmavnode project names this mode "bcastlock", https://github.com/mavlink/mavros - "permanent broadcast" (https://github.com/mavlink/mavros/pull/1043).

This mode is useful when you want multiple vehicles to use one endpoint (multiple GCS with different MAVLink IDs, direct vehicle communication without routing).

src/mavlink-router/endpoint.cpp
dvornikov-aa: sockaddr contains a target IP address. We rewrite it on every incoming packet (recvfrom). To ignore this change we need to add an additional sockaddr structure in the endpoint class.

https://github.com/intel/mavlink-router/blob/da8607430c9e336e826d576c15f555c5f0fe033c/src/mavlink-router/endpoint.cpp#L761

JonasVautherin (commented):
I agree on the fact that setting a broadcast address in "Normal" mode will not always work as one would expect, because the messages are sent to the address of the last received message.

So if you set the target address as `192.168.1.255/24`, and you have two clients A (`192.168.1.10/24`) and B (`192.168.1.11/24`), both A and B will receive the first few messages. Whenever mavlink-router receives a message from A, it will replace the target address with `192.168.1.10/24` and B will stop receiving messages. Similarly when it receives a message from B. Because probably both A and B will continuously send messages, the target address will flip between both and the behavior will be unexpected.

This said, I don't understand the need for a "Broadcast" mode. Why not always using `recv()` instead of `recvfrom()` and never change the target address? This way the broadcast may just work, right? Or am I missing something?

jbeyerstedt:
I'll close this for now because of missing activity. Please continue the discussion about the functionality itself in #350 and see the notes on another implementation attempt there.
(Or just add a comment here, if closing was a mistake.)

PR #181, opened 2019-09-17
Enable MAVLink CRC checks on UDP/TCP

bkueng:
UDP and TCP disabled MAVLink CRC checks because those protocols have
built-in CRC already. However there are several failure cases when relying
only on these:
- For the UDP case, if a UDP packet is not aligned with MAVLink messages,
  UDP packet drops or reorderings lead to corrupt MAVLink messages.
  For example some implementations fill up a fixed-size UDP packet.
- A connected component might just forward bytes that come from another
  (lossy) link. In my case it was a shared memory implementation
  translating to UDP, and the shared memory implementation
  dropped/had corrupt bytes.

These cases can lead to corrupt log files and even busy-loops of mavlink
router within the log processing.

An alternative would be to just ensure CRC is checked in the log classes.

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -44,9 +44,8 @@
 
 #define UART_BAUD_RETRY_SEC 5
 
-Endpoint::Endpoint(const char *name, bool crc_check_enabled)
+Endpoint::Endpoint(const char *name)
     : _name{name}
-    , _crc_check_enabled{crc_check_enabled}
 {
     rx_buf.data = (uint8_t *) malloc(RX_BUF_MAX_SIZE);
     rx_buf.len = 0;
@@ -216,7 +215,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
     _stat.read.total++;
 
     msg_entry = mavlink_get_msg_entry(*msg_id);
-    if (_crc_check_enabled && msg_entry) {
+    if (msg_entry) {
         /*
          * It is accepting and forwarding unknown messages ids because
          * it can be a new MAVLink message implemented only in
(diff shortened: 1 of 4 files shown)


zehortigoza:
You are the first one to hit this problems, probably because of your setup, so please add a parameter to disable it and keep the CRC check disabled by default.

lucasdemarchi:
The idea behind not checking crc for udp/tcp is that you always have 1 mavlink message per packet or at least aligned.

Since we have a ring buffer to support serial already and handle partial messages, I don't mind enabling crc in general. The overhead should be negligible.

lucasdemarchi:
Please rebase this to remove the merge commit you have .

bkueng:
@lucasdemarchi hmm it looks like your update to master excluded the commit from https://github.com/intel/mavlink-router/pull/176. I rebased and kept it here.

lucasdemarchi:
While testing the CI I think I accidentally pushed to master and when I rolled back master to the previous state it went to the wrong ref. Sorry for the trouble

bkueng:
@zehortigoza do you still want the parameter? Personally I would like to avoid it, as I don't think it imposes noticeable overhead either, while on the other hand it's something difficult to debug and know when you would actually have to set the parameter. It's something that is likely to trigger problems only in rare cases.

src/mavlink-router/endpoint.cpp
```diff
@@ -44,9 +44,8 @@
 
 #define UART_BAUD_RETRY_SEC 5
 
-Endpoint::Endpoint(const char *name, bool crc_check_enabled)
-    : _name{name}
-    , _crc_check_enabled{crc_check_enabled}
+Endpoint::Endpoint(const char *name)
+    : _name {name}
```
lucasdemarchi: extra space in the initialization bracket is not according to the coding style. Here and everywhere else

lucasdemarchi:
I rebased, fixed the style issues and also added the relevant config to .clang-format in a separate commit

PR #183, opened 2019-09-21
Travis CI: Lint Python code for syntax errors and undefined names

cclauss:
15 files with Python syntax errors https://travis-ci.org/intel/mavlink-router/builds/587837498#L334

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ matrix:
             - python3-setuptools
 
 before_script:
-        - pip3 install --user future pymavlink
+        - pip3 install --user flake8 future pymavlink
+        - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         - CXX="g++-6"
         - CC="gcc-6"
 


lucasdemarchi:
Sorry for the long time. We switched to Github Actions, so this needs an update

.travis.yml
```diff
@@ -15,7 +15,8 @@ matrix:
             - python3-setuptools
 
 before_script:
-        - pip3 install --user future pymavlink
+        - pip3 install --user flake8 future pymavlink
+        - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
lucasdemarchi, line 19: could be a separate workflow for code style in Github Actions

cclauss: These test have nothing to do with code style.  Use psf/black to autofix code style.

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

lucasdemarchi: Ok, may make sense to have them in the same workflow. Thanks

PR #184, opened 2019-10-01
LogEndpoint: call aio_fsync() once per second

lucasdemarchi:
With the default kernel settings (/proc/sys/vm/dirty_expire_centisecs
= 3000), it is possible to lose up to 30s of logging data in events of
power loss.
This patch reduces worst-case loss to 1s by calling fsync() every second.


lucasdemarchi:
This is the restored commit from #176  that got lost

PR #187, opened 2019-11-12
Clean gtest

caleridas:
Rework test/gtest infrastructure conforming to usual automake mechanisms. Provide sample test in mainloop and fix variable concurrently accessed by signal handler.


lucasdemarchi:
> Rework test/gtest infrastructure conforming to usual automake mechanisms

But then you break platforms that don't provide gtest or that have a different version. gtest itself suggested to bundle the library rather than using the system-provided one since they do break API/abi.

> fix variable concurrently accessed by signal handler.

Please provide an example in which it breaks. What arch you are using that would cause tearing on a bool?

Also, please **rebase** (rather than several pulls) this PR on a clean master branch. 

Thanks

caleridas:
> > Rework test/gtest infrastructure conforming to usual automake mechanisms
> 
> But then you break platforms that don't provide gtest or that have a different version.

Nothing breaks relative to current state: The existing "unit test" verifies that 2+2=4, I think it is fair to say that very little is there to be broken to begin with. Also, the change is structured such that it gracefully handles not having gtest (it then simply does not build tests).

If gtest is not packaged for your system (which is increasingly uncommon as it has reached Debian stable already and that's trailing everyone else), then downloading and installing it is not hard. The reverse, namely dozens of individual packages all pulling and random and incompatible versions, just is not a solution either and makes system integration terrible. Just for a moment pretend that you would want a custom instrumented test harness version for all of the software...

> gtest itself suggested to bundle the library rather than using the system-provided one since they do break API/abi.

API is very stable, actually. ABI is not an issue unless linking dynamically (which nobody does). We may have originally suggested bundling, but by now they are actually actively providing pkg-config and build config files.

> > fix variable concurrently accessed by signal handler.
> 
> Please provide an example in which it breaks. What arch you are using that would cause tearing on a bool?

ISO 14882:2011, aka C++11 section 1.9. Basically the only thing you are allowed to touch in a signal handler are objects of type "atomic<some_types>" and  "volatile sig_atomic_t" for legacy reasons. volatile bool is *not* among the exceptions.

It is not about "tearing", concurrency and compiler optimizations are much more complicated than that. There is no actual failure, this is a pure "drive-by" fix. I can separate it into a commit of its own if you prefer.

> Also, please **rebase** (rather than several pulls) this PR on a clean master branch.

Sorry that first push was a bit messed, this is rebased cleanly now.

Thanks for comments!

lucasdemarchi:
> Nothing breaks relative to current state: The existing "unit test" verifies that 2+2=4, I think it is fair to say that very little is there to be broken to begin with. Also, the change is structured such that it gracefully handles not having gtest (it then simply does not build tests).

Fair enough.

> API is very stable, actually. ABI is not an issue unless linking dynamically (which nobody does). We may have originally suggested bundling, but by now they are actually actively providing pkg-config and build config files.

Ok, I can have you word on that and ping you when/if things don't work :)

> ISO 14882:2011, aka C++11 section 1.9. Basically the only thing you are allowed to touch in a signal handler are objects of type "atomic<some_types>" and "volatile sig_atomic_t" for legacy reasons. volatile bool is not among the exceptions.

Yet it works with compilers and archs we use this in... If this is broken I know a bunch of other projects you may want to patch.

> It is not about "tearing", concurrency and compiler optimizations are much more complicated than that. There is no actual failure, this is a pure "drive-by" fix.

With the memory order you passed, the only thing it could possibly fix is tearing as you explicitly said it doesn't care about the reorder ( by compiler or out of order execution).

> I can separate it into a commit of its own if you prefer.

I don't mind the change as long as long as it's in a separate commit and justification is given like citing the standard as you did here.

caleridas:
> > Nothing breaks relative to current state: The existing "unit test" verifies that 2+2=4, I think it is fair to say that very little is there to be broken to begin with. Also, the change is structured such that it gracefully handles not having gtest (it then simply does not build tests).
> 
> Fair enough.
> 
> > API is very stable, actually. ABI is not an issue unless linking dynamically (which nobody does). We may have originally suggested bundling, but by now they are actually actively providing pkg-config and build config files.
> 
> Ok, I can have your word on that and ping you when/if things don't work :)
> 
> > ISO 14882:2011, aka C++11 section 1.9. Basically the only thing you are allowed to touch in a signal handler are objects of type "atomic<some_types>" and "volatile sig_atomic_t" for legacy reasons. volatile bool is not among the exceptions.
> 
> Yet it works with compilers and archs we use this in...

"never failed so far" != "correct"

> If this is broken I know a bunch of other projects you may want to patch.

True, 9 out of 10 projects doing anything with signal handlers get it wrong in this or many other random ways.

> > It is not about "tearing", concurrency and compiler optimizations are much more complicated than that. There is no actual failure, this is a pure "drive-by" fix.
> 
> With the memory order you passed, the only thing it could possibly fix is tearing as you explicitly said it doesn't care about the reorder ( by compiler or out of order execution).

No, this fixes undefined behavior (standard says: "the value of any object not in either of these two categories that is modified by the handler becomes undefined"). 

The meaning of memory order is far more subtle than you seem to suggest (it establishes "inter-thread happens-before" to allow race-free access to other non-atomic data -- not of relevance for signal handler since 1. not a different thread and 2. no other shared data is accessed).

> > I can separate it into a commit of its own if you prefer.
> 
> I don't mind the change as long as long as it's in a separate commit and justification is given like citing the standard as you did here.

Good, will do, thanks again for feedback.

lucasdemarchi:
> No, this fixes undefined behavior (standard says: "the value of any object not in either of these two categories that is modified by the handler becomes undefined").

Undefined behavior by the standard doesn't invalidate what I said. It may not be according to the standard, but the assembly generated by the compilers we support for the architectures we support is correct. If you think otherwise I would ask you to take a look on arm and x86-64 manuals. Or a reproducer.  I agree it's not following the c++ spec, I disagree this means it has a bug.

> The meaning of memory order is far more subtle than you seem to suggest (it establishes "inter-thread happens-before" to allow race-free access to other non-atomic data -- not of relevance for signal handler since 1. not a different thread and 2. no other shared data is accessed).

You chose the relaxed model, which means you don't have any memory barriers and the compiler is also free to reorder the loads and stores around.

caleridas:
> > No, this fixes undefined behavior (standard says: "the value of any object not in either of these two categories that is modified by the handler becomes undefined").
> 
> Undefined behavior by the standard doesn't invalidate what I said. It may not be according to the standard, but the assembly generated by the compilers we support for the architectures we support is correct.

C++ is not defined by the assembly it generates but by an abstract machine following the rules layed out in the standard. Compilers are free to do whatever as long as it follows the rules, have changed behavior in the past in ways not expected by people arguing from the assembly side, and will continue to do so. I honestly do not see a point in arguing that (avoidable) undefined behavior is okay.

> If you think otherwise I would ask you to take a look on arm and x86-64 manuals. Or a reproducer. I agree it's not following the c++ spec, I disagree this means it has a bug.

As I happened to have implemented this part of the standard (also for ppc, sparc & others back then), the code generated is perfectly well known to me, but is not even of relevance. My definition of bug is "something that *can* go wrong" and a bug-free program is one where the absence of bugs is provable, while your definition of bug is closer to "something that has been *proven* to have gone wrong at least once already" and a bug-free program is one where the presence of bugs has not been proved yet.

> > The meaning of memory order is far more subtle than you seem to suggest (it establishes "inter-thread happens-before" to allow race-free access to other non-atomic data -- not of relevance for signal handler since 1. not a different thread and 2. no other shared data is accessed).
> 
> You chose the relaxed model, which means you don't have any memory barriers and the compiler is also free to reorder the loads and stores around.

The C++ memory model is not about memory barriers (which are an instruction-level construct) but about ordering relations.

I split this portion of into a commit of its own now, please have a look.

Thanks!

lucasdemarchi:
> C++ is not defined by the assembly it generates but by an abstract machine following the rules layed out in the standard. Compilers are free to do whatever as long as it follows the rules, have changed behavior in the past in ways not expected by people arguing from the assembly side, and will continue to do so. I honestly do not see a point in arguing that (avoidable) undefined behavior is okay.
> 

I think continuing to argue this is not much productive as we have different views and that is fine.
I already said the change is acceptable if split into another commit.

>  "something that has been proven to have gone wrong at least once already"

not really, but as I said, not much productive discussion

PR #196, opened 2020-02-12
Pr mainloop testable upstream

caleridas:
Refactor mainloop to make it as well as endpoint code runnable in unit tests. Add simple test case verifying that udp endpoints are set up as configured and that direct send of mavlink messages behaves as expected.


jkflying:
Being handled in #242

PR #202, opened 2020-04-02
Update mavlink

### LINKED ISSUE #201: pl update mavlink submodule
it is fantastic that version 2 just came out few days ago, sadly the mavlink submodule is 2 years old however ...
it would be great if it could be updated to the latest and freshest :)
thx

lucasdemarchi:
Fix #201. This needs proper testing.

### DIFF
--- a/src/mavlink-router/binlog.cpp
+++ b/src/mavlink-router/binlog.cpp
@@ -119,8 +119,8 @@ int BinLog::write_msg(const struct buffer *buffer)
         return buffer->len;
     }
 
-    if (payload_len > msg_entry->msg_len) {
-        payload_len = msg_entry->msg_len;
+    if (payload_len > msg_entry->max_msg_len) {
+        payload_len = msg_entry->max_msg_len;
     }
 
     if (mavlink2) {
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -408,10 +408,10 @@ uint8_t Endpoint::get_trimmed_zeros(const mavlink_msg_entry_t *msg_entry, const 
         return 0;
 
     /* Should never happen but if happens it will cause stack overflow */
-    if (msg->payload_len > msg_entry->msg_len)
+    if (msg->payload_len > msg_entry->max_msg_len)
         return 0;
 
-    return msg_entry->msg_len - msg->payload_len;
+    return msg_entry->max_msg_len - msg->payload_len;
 }
 
 void Endpoint::log_aggregate(unsigned int interval_sec)
--- a/src/mavlink-router/ulog.cpp
+++ b/src/mavlink-router/ulog.cpp
@@ -148,8 +148,8 @@ int ULog::write_msg(const struct buffer *buffer)
         return buffer->len;
     }
 
-    if (payload_len > msg_entry->msg_len) {
-        payload_len = msg_entry->msg_len;
+    if (payload_len > msg_entry->max_msg_len) {
+        payload_len = msg_entry->max_msg_len;
     }
 
     if (mavlink2) {


lucasdemarchi:
@olliw42 @bkueng

olliw42:
many thx, this was really fast response, much appreciated
I unfortunately can't yet test since I'm stuck with getting it to build on a rpi (did use maverick before, but need the new mavlink). I'll certainly try once I'm operational.

olliw42:
@lucasdemarchi is there anything specific you would want me to test (if I'm capable of)?

lucasdemarchi:
@olliw42, testing that flight stack logging with ardupilot still works would be great

olliw42:
thx, I'll try my best :)

olliw42:
ok, I don't think I'm helpful here

what I did:
- cloned the branch tip-update-mavlink to a separate folder
- in main.cpp changed the line puts(PACKAGE " version " VERSION); to puts(PACKAGE " version " VERSION "xxx");
- did all the config&build&install steps 
- checked that I run the new version by calling mavlink-routerd --version

in the various different tests I did, the new version behaved exactly like the old, not sure I would be supposed to see any difference if the above changes would not work.

as regards the logging: I don't get it to work for either the old nor the new version. It creates the .bin files but they always are of size 0.

To be more detailed:
I do the testing in various configurations, they all behave the same so I just describe for this: MP on PC - usb - pixhawk - uart/telem1 - RPi3 - usb-ttl adapter - STorM32 controller. On the RPi I have in addition heartbeat-print-tcp.py 0.0.0.0:5760 running. So three endpoints in total, and 4 mavlink components in total (and I do see all). I call mavlink-routed -d /home/pi/Documents/mavlink-router and in the specified directory I have a file my.config with this content:
[General]
   MavlinkDialect=ardupilotmega
   Log=/home/pi/Documents/mavlink-router/logs
   LogMode=always
[UartEndpoint first]
   Device=/dev/serial0
   Baud=115200
   FlowControl=false
[UartEndpoint second]
   Device=/dev/ttyUSB0
   Baud=115200
   FlowControl=false
This config file is obviously accepted by mavlink-routerd. It nicely routes. It also produces .bin files in the given log directory, but as said, their size is always 0. It seems that nothing is been written to the logs.

I should say that I haven't yet figured out the proper way to stop mavlink-routerd. I start it in a console and stop it with ^C or ^Z (which makes the tcp to hang and I have to kill it manually). Not sure if this rude termination makes zero-sized .bin files on unix.

lucasdemarchi:
@olliw42 have you set the LOG_DISARMED=1 and LOG_BACKEND_TYPE=3 params in ardupilot?

olliw42:
yes, this did the trick, I changed the settings as you suggested and the .bin is now indeed created and filled. I also can read the bin with MissionPlanner and it seems to contain valid data. So, if this was the test I should do I would say it is passed.

many thx for the lesson

:)

as a corollary I however have to say that this behavior doesn't make much sense to me. I mean, what is an option `LogMode=always` supposed to mean if the actual behavior doesn't depend on it but is governed by settings in the autopilot? I'm also fundamentally not happy with that the logging of the router depends on the autopilot, I want to have router logs in order to see what the router (and its attached components) is doing and not to see what the autopilot is doing. 

I can understand that there probably was a reason for why this was done at the time it was done, but when coming from outside without such background like me it's really all but not apparent.

Maybe this could be a suggestion: 

One could change the existing LogModes `always` and `while-armed` to do what they say independent on any setting in the flight controller, and add a third option which behaves according to the LOG settings in the autopilot as you described.

lucasdemarchi:
> One could change the existing LogModes `always` and `while-armed` to do what they say independent on any setting in the flight controller, and add a third option which behaves according to the LOG settings in the autopilot as you described.

no, we don't change params on users' back.  Btw, we would probably have to change to backend as well and that is very intrusive and would come as a surprise.

The history behind these value is: previously we always logged and the user was responsible to configure backend and when to log on flight stack. PX4 however didn't have the option to log only when armed and they found it better to rather implement it on mavlink-router.

So... this is a client configuration (the Logging is a "special" endpoint). However user is still supposed to know what he needs to configure on the flight stack.

olliw42:
I understand
I think a bit more documentation could help noobs to better understand what a option "always" is really about ... 
I've spend some time on trying to understand how the various settings on the two ends would behave, but have given up, too much effort. I'll better add my own option :)

Issue #203, opened 2020-04-02
compiling on raspberry pi isn't smooth


olliw42:
I'm trying to install mavlink-router on a raspberry pi3 with a recent fresh raspbian (Raspbian Buster, Version: February 2020, Release date: 2020-02-13, Kernel version: 4.19), and it doesn't go smoothly.

I'm a unix illiterate largely and don't really know what the consequences are of what I am doing, but here are the deviations from the docs I did so far:

- ./autogen.sh did complain about autoreconf, so I did a sudo apt-get install autoconf which did get me past this

- ./configure xxxxx does complain about "syntax error near unexpected token disable-static, LT_INIT(disable-static pic-only)", so I have to edit the file configure.ac and uncomment the line  LT_INIT([disable-static pic-only]) (before running ./autogen.sh). This at least brings me past that complain, and the configure terminates with this mavlink router 2 overview info (I'm not sure I should post all the output)

btw: it's not clear what options to really chose for a "common configuration", the info in the README is seemingly different to the info outputted by ./autogen.sh. So I tried 

./configure CFLAGS='-g -O2' CXXFLAGS='-g -O2' --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib --prefix=/usr

and 

./configure CFLAGS='-g -O2' CXXFLAGS='-g -O2' --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib64 --prefix=/usr

- make does the compile, but not the linking and stops with error

  CXXLD    mavlink-routerd
/usr/bin/ld: src/mavlink-router/logendpoint.o: in function `LogEndpoint::_fsync()':
/home/pi/Documents/Github/mavlink-router/src/mavlink-router/logendpoint.cpp:397: undefined reference to `aio_error64'
/usr/bin/ld: /home/pi/Documents/Github/mavlink-router/src/mavlink-router/logendpoint.cpp:404: undefined reference to `aio_fsync64'
/usr/bin/ld: src/mavlink-router/logendpoint.o: in function `LogEndpoint::LogEndpoint(char const*, char const*, LogMode, unsigned long, unsigned long)':
/home/pi/Documents/Github/mavlink-router/src/mavlink-router/logendpoint.cpp:62: undefined reference to `aio_init'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:880: mavlink-routerd] Error 1
make: *** [Makefile:696: all] Error 2

[...]

PR #206, opened 2020-04-03
Update to README.md

olliw42:
this PR aims to improve the install instructions in the README.md file, following the result coming out of issue #203.

COMMENT: 
In this PR it says that one should use --libdir=/usr/lib64 on a 64-bit system and --libdir=/usr/lib on a 32-bit system. Please tell if that is indeed correct.

The other additions/changes to the instructions work for me on a RPi3A with latest Raspbian and with starting from a fresh install.

In addition I made some changes to the text, which are NFC but I believe improve readability.


README.md
```diff
@@ -26,16 +26,28 @@ We currently depend on mavlink C library which is generated by the build
 system during compilation. The corresponding submodule should be fetched.
 
     $ git submodule update --init --recursive
+    
+The following libraries need also to be installed :
+
+    $ sudo apt install python-future
+    $ sudo apt install python3-future
+    $ sudo apt install libtool
+    $ sudo apt install autoconf
```
lucasdemarchi, line 35: My problem with package names like these is that they change name from version to version. And it is exclusive to one Linux distro. I myself don't use Ubuntu and don't have apt. So I think this could be acceptable by being explicit, starting with "In Ubuntu".

I'm surprised we need both python-future and python3-future.

There are also other dependencies that you probably already had installed, but I think build-essential in Ubuntu should bring everything needed (including gcc). So I think this is the final result:

```
$ sudo apt-get install python-future python3-future build-essential
```

README.md
```diff
@@ -26,16 +26,28 @@ We currently depend on mavlink C library which is generated by the build
...
 
 #### Build ####
 
 Build system follows the usual configure/build/install cycle. Configuration is needed
 to be done only once. A typical configuration is shown below:
+The build system follows the usual configure/build/install cycle. Configuration is needed
+to be done only once. A typical configuration for a 64-bit system is shown below:
```
lucasdemarchi, line 40: some copy and paste mistakes above? we have duplicate sentences now

README.md
```diff
@@ -26,16 +26,28 @@ We currently depend on mavlink C library which is generated by the build
...
 
     $ ./autogen.sh && ./configure CFLAGS='-g -O2' \
             --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib64 \
+        --sysconfdir=/etc --localstatedir=/var --libdir=/usr/lib64 \
```
lucasdemarchi, line 43: same here

README.md
lucasdemarchi: oh... you fixed in the following commit.

README.md
lucasdemarchi: this too

README.md
```diff
@@ -117,21 +114,21 @@ That would change `Endpoint bravo` baudrate to `115200`.
...
-Mavlink router can also collect flight stack log. It supports collecting
+Mavlink router can also collect flight stack logs. It supports collecting
 both PX4 and Ardupilot flight stacks logs. To start logging, set a
-directory to `Log` key on `General` section of config file (or use
+directory to the `Log` key in the `General` section in the config file (or use
 argument option `-l`).
-Currently, mavlink router needs to be informed which MAVLink dialect
-flight stack speaks, `common` or `ardupilotmega`. To define it, use
+Currently, mavlink router needs to be informed which MAVLink dialect the
+flight stack speaks, `common` or `ardupilotmega`. To define it, use the
```
lucasdemarchi, line 122: humn... this is actually outdated now. We have "auto", that should be doing the right thing.

olliw42: ah, I was wondering what auto exactly is, thx. (and I'm wondering how it actually would work). I'll update my knowledge and will revise.

README.md
olliw42: I've tried that line using a totally fresh install, and I could not get it to work (I tried it repeatedly with slightly different sequences). I always need to install both autoconf and libtool. When I install build-essential I get the info that the currently installed build-install is already up to date. So, things seem to be consistent.

lucasdemarchi: Ok, thanks. Maybe this changed in recent years and they don't include libtool/autoconf anymore. Thanks

olliw42: so, what should I do, consider this conversation as resolved? break the descriptions out into a separate RPi section? Some advice would be much appreciated

Issue #207, opened 2020-04-03
Routing not conform with MAVlink routing rules: Bug, feature, or what?


olliw42:
mavlink-router is not routing the messages according to the MAVLink routing rules: 

While a target system id is correctly respected, a target component id is not always correctly respected.

I believe to see that very clearly in my testings, and I think that this is also very obvious from the code:
https://github.com/intel/mavlink-router/blob/18891603c436dd360bb77589f4120507124856ad/src/mavlink-router/endpoint.cpp#L336-L349

So, the question is, why is that? Is there a (good?) reason for that behavior, or is it just because of "history", or is this indeed a bug?

I would argue it is clearly a bug, since it is not conform with the rules and produces (much) unwanted traffic, and that's why I'm actually not really pleased by this behavior, also because I fail to see a really good reason for it.

any clarification on this point would be much appreciated.

many thx, Olli

lucasdemarchi:
> mavlink-router is not routing the messages according to the MAVLink routing rules:

what rules?

> I believe to see that very clearly in my testings,

can you expand on this? It would be interesting to know what you are testing and how you are seeing a wrong behavior.

what mavlink-router does is to route packets from one endpoint to another. In order to do that it uses it knowledge of "is this system available on this endpoint?". If yes, then it writes the packet to that endpoint (with the exception of "is it the same endpoint?" to avoid loops - see commit ad367033fd75f1535c82d2e604c7e6be93e16ad5)

From the code you mentioned, what appears wrong is the comment: we actually route any packet to the endpoint as long as the systemid matches (or is a broadcast).

We can't add a match on the compid, because that would imply that every component would need to broadcast at least one packet (usually

olliw42:
> what rules?

https://mavlink.io/en/guide/routing.html

> can you expand on this? It would be interesting to know what you are testing and how you are seeing a wrong behavior.

Here a case demonstrating the violation of the MAVLink routing rules:


Setup:
-----

ArduPilot - RPi3 - STorM32(Gimbal&Camera)

on the RPi3 in addition to mavlink-router runs a py script, which connects to tcp:5760 and emits a heartbeat with 1/191

so we have 4 MAVLink components, and 3 endpoints: 
* ArduPilot: 1/1 on endpoint [4]
* Gimbal: 1/154 on endpoint [5]
* Camera: 1/105 on endpoint [5]
* py script: 1/191 on endpoint [8]


Case:
------

[...]

PR #208, opened 2020-04-03
Corrected MAVLink routing

olliw42:
resolves issue #207

The comment in L. 344 is IMHO saying what it should do, but the if statement which follows becomes true also then the message has a target comp id > 0 and thus is not broadcasting. 

this PR corrects this.

I've tested this on the system described in the above issue, and it seems to exactly do what I would think it should do.

:)

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -333,7 +333,7 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         return false;
     }
 
-    // Message is broadcast on sysid: accept msg
+    // Message is broadcast on sysid or sysid is non-existent: accept msg
     if (target_sysid == 0 || target_sysid == -1)
         return true;
 
@@ -341,8 +341,8 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     if (target_compid > 0 && has_sys_comp_id(target_sysid, target_compid))
         return true;
 
-    // This endpoint has the target of message (sysid, but compid is broadcast): accept
-    if (has_sys_id(target_sysid))
+    // This endpoint has the target of message (sysid, but compid is broadcast or non-existent): accept
+    if ((target_compid == 0 || target_compid == -1) && has_sys_id(target_sysid))
         return true;
 
     // Reject everything else


src/mavlink-router/endpoint.cpp
```diff
@@ -342,7 +342,7 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         return true;
 
     // This endpoint has the target of message (sysid, but compid is broadcast): accept
-    if (has_sys_id(target_sysid))
+    if ((target_compid == 0 || target_compid == -1) && has_sys_id(target_sysid))
```
lucasdemarchi, line 345: You need to update the comment above: "but compid is broadcast or non-existent

olliw42: frankly, you guys are weird ... it exactly follows the recommendation given somewhere else to follow style and conventions, and here it exactly follows the convention set in L.336 just slightly in the above ...
and indeed, =0 and not existing have the same effect and could and should be both called "broadcast"
anyway, I guess you were kindly asking me to then also change L.336, so, I'll do this ;)

btw, I'm wondering if one shouldn't better use (target_sysid <= 0) and (target_compid <= 0) in L.337 and L.345, less clumsy and the compiler might also do less code, I would like it better. What do you say?

lucasdemarchi: May I ask you to stop with sentences like "frankly, you guys are weird"? It doesn't help in getting things merged/fixed.

lucasdemarchi:
@olliw42 I updated this PR squashing the commits and being a bit more verbose on the commit message.

As for breaking previous behavior (and my argument for being cautious), see that this change breaks the script executed by CI: https://travis-ci.org/github/intel/mavlink-router/builds/671341216#L898

Looks like tests/routing_test.py would need an update, too.

bkueng:
This looks correct, and at least in theory should not break things.

src/mavlink-router/endpoint.cpp
olliw42: May I ask you to stop commanding folks when they have perfectly blended in and followed what was some elses mistake. It doesn't help getting things fixed. Politeness is a mutual thing, not unidirectional.

lucasdemarchi: @olliw42 it was a simple code review as usual, nothing else. If you disagree, you are perfectly entitled to say "but not having compid is the same as being a broadcast". Perhaps being too succinct in my comment gave the impression I was commanding something, sorry.

olliw42:
thanks, sir!!

PR #216, opened 2020-05-31
Remove GNU-specific aio_init() function for musl compatibility

JonasVautherin:
`aio_init()` is GNU-specific, and not compatible with musl. Is there a reason for having those lines (i.e. has it been proven that this is useful), or can they be removed?

They are all that's needed to be fixed to support musl again.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,8 @@ LT_INIT([disable-static pic-only])
 # Function and structure checks
 #####################################################################
 
+AC_CHECK_DECLS([aio_init], [], [], [#include <aio.h>])
+
 AC_MSG_CHECKING([whether _Static_assert() is supported])
 AC_COMPILE_IFELSE(
 	[AC_LANG_SOURCE([[_Static_assert(1, "Test");]])],
--- a/src/mavlink-router/logendpoint.cpp
+++ b/src/mavlink-router/logendpoint.cpp
@@ -55,11 +55,13 @@ LogEndpoint::LogEndpoint(const char *name, const char *logs_dir, LogMode mode,
     _add_sys_comp_id(LOG_ENDPOINT_SYSTEM_ID << 8);
     _fsync_cb.aio_fildes = -1;
 
+#if HAVE_DECL_AIO_INIT
     aioinit aio_init_data {};
     aio_init_data.aio_threads = 1;
     aio_init_data.aio_num = 1;
     aio_init_data.aio_idle_time = 3; // make sure to keep the thread running
     aio_init(&aio_init_data);
+#endif
 }
 
 void LogEndpoint::_send_msg(const mavlink_message_t *msg, int target_sysid)


lucasdemarchi:
aio is currently used for asynchronously sync the fs. We can't make it sync as it would block the mainloop which is not acceptable - See commit in which it was implemented - it may take a long time.

If not using aio, then we'd need a worker thread, i.e. basically implementing the same thing but on our side rather than relying on libc.

JonasVautherin:
I don't think there is a need to remove aio completely. As you can see, I did not remove the rest of the aio calls, e.g. [`aio_fsync`](https://github.com/intel/mavlink-router/blob/fcfcba4213900080752b6b9397b1ac66f90ef7b1/src/mavlink-router/logendpoint.cpp#L398). 

This particular function, `aio_init`, which here sets `aio_threads`, `aio_num` and `aio_idle_time` does not exist in musl. I am not an expert in libc, but when I asked musl people, they said that unless this `aio_init` call had been proven to significantly improve performance, it was probably not needed.

Because it is GNU-specific (as per [man aio_init](https://www.man7.org/linux/man-pages/man3/aio_init.3.html)), I would think that musl does not need it. So if we knew that glibc does need those particular init values, we could even go for an `#ifdef` here.

Hence my question: do we know why this call of `aio_init()` made it here, or is it more the result of reading an example that did it while implementing the `aio_fsync` thing?

Does that make any sense at all, or am I completely off? :smile:

lucasdemarchi:
Oh... now I understand your question. I think the main reason is for the call is because the default on glibc with 20 worker threads is waaaay overkill for what we are doing. We know we will only ever need 1 thread for any "do this in the background thing", so we set  `aio_threads` to 1 .

JonasVautherin:
I see. Would it be fine with an `#ifdef` for musl, then?

lucasdemarchi:
Rather than cluttering the source code with ifdef, just add it to the missing.h header and leave the ifdef there.

JonasVautherin:
Sorry I did not get that. What's the `missing.h` header? My understanding was that I should move `aio_init()` to this `missing.h` header, but I can't seem to find any file with this name :thinking:.

lucasdemarchi:
ugh... working on too many projects... I was referring to https://github.com/lucasdemarchi/kmod/blob/master/shared/missing.h thinking we had the same infra here, too.
If you think you can add it to mavlink-router, it would be great. Otherwise, just add it to the only place needed and we can improve later. Thanks.

JonasVautherin:
Here is a proposition with the `#ifdef` directly in the code. So your suggestion would be to define `aio_init` in `missing.h` if `HAVE_AIO_INIT` is not defined, and make it a noop.

I'll try to make that later, but I think this should do it for now.

configure.ac
```diff
@@ -48,6 +48,13 @@ AC_COMPILE_IFELSE(
 	 AC_MSG_RESULT([yes])],
 	[AC_MSG_RESULT([no])])
 
+AC_MSG_CHECKING([whether aio_init is supported])
+AC_COMPILE_IFELSE(
+        [AC_LANG_PROGRAM([[#include <aio.h>]], [[aio_init(NULL);]])],
+        [AC_MSG_RESULT([no])],
+        [AC_DEFINE([HAVE_AIO_INIT], [1], [Define if aio_init is available]),
```
JonasVautherin, line 55: I'm confused now. If I invert the "yes" and the "no" statements, then it fails on both ArchLinux (glibc) and Alpine (musl). But reading the [documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Generating-Sources.html), it should create a file that includes `<aio.h>` and try to run `aio_init(NULL);`. So it should go the other way round, right? :confused:

lucasdemarchi:
@JonasVautherin I changed to a simple check for declaration rather than compiling a test program. Also note that we should not use `#ifdef` since it will be defined in all cases, either to 0 or 1. Now build is passing on both glibc and musl. Thanks!

JonasVautherin:
Oh, very good to know! Thanks a lot! :blush:

PR #217, opened 2020-05-31
added dockerfile and example docker compose file

orshefi:
Hi guys,

Thought nowadays it's important to add basic dockerfile for that important utility.

It's not close to cover all use cases, my focus was to enable my use case which is to route uart device to UDP port, feel free to modify to your needs.

Also, I added a little docker-compose example to show a use case I find important

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:16.04
+
+ARG UART_DEVICE
+ARG UART_BAUDRATE
+ARG UDP_OUT_IP
+ARG UDP_OUT_PORT
+
+ARG DEBIAN_FRONTEND=noninteractive
+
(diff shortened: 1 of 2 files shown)


lucasdemarchi:
Any reason to require running it inside a container rather than directly on host?

orshefi:
Sure,

Docker allows easy deployment, I don't have to worry for dependencies to be installed on the host or configuration files.

The Image I built if fully packed with the dependencies relevant to mavlink-router and all I have to do is to include it in my k8s configuration to make it work.

I have no affinity to any host, this is the easiest & most efficient way to scale a working solution in my opinion.

lucasdemarchi:
Merged another PR adding 2 example docker containers

PR #218, opened 2020-06-03
Move to github workflows and add alpine (musl)

JonasVautherin:
This depends on #216. I just think that if the codebase already supports musl (and with #216, it would), then it's worth having CI for it. At the same time I propose to move the CI to github workflows.

Of course, the musl build is failing for now, because #216 has not been merged.


.github/workflows/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    - 'master'
+    tags:
+    - 'v*'
+  pull_request:
+    branches:
+    - '*'
+
+jobs:
+  ubuntu18:
+    name: ubuntu 18.04
```
lucasdemarchi, line 15: could we make this "latest ubuntu/fedora/whatever" rather than 18.04?

.github/workflows/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-18.04
+    env:
+      CXX: g++-6
+      CC: gcc-6
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: recursive
+      - name: install tools
+        run: sudo apt-get install -y g++-6 gcc-6 python3-pip python3-setuptools && pip3 install future pymavlink
```
lucasdemarchi, line 25: and without giving a specific gcc version...

.github/workflows/main.yml
JonasVautherin: I am usually not in favor of "latest", because that may break when it changes. I could fix it to 20.04, it's just that I wanted to reproduce the current travis build, which I believe is on 18.04.

Your call: 20.04 or should I try "latest"?

.github/workflows/main.yml
JonasVautherin: Again, that's what came from Travis, it was actually more work to keep it this way than to remove this constraint :-). I'll update that.

.github/workflows/main.yml
lucasdemarchi: if there is a "latest" that always point to the latest version, better. That way we don't need to keep updating it.

I like to live on bleeding edge. If it breaks, we should fix because sooner or later someone will be using that version.

.github/workflows/main.yml
lucasdemarchi: I think this is leftover from when gcc 5.x was the default and we wanted at least 6... not sure. Thanks!

PR #219, opened 2020-06-05
fix reuse of r loop variable for p->handle_read

GerardHale:
r the number of polled events is changed if there is a readable event EPOLLIN 
Changed to a local temporary variable rd.


src/mavlink-router/mainloop.cpp
```diff
@@ -282,8 +282,8 @@ void Mainloop::loop()
             Pollable *p = static_cast<Pollable *>(events[i].data.ptr);
 
             if (events[i].events & EPOLLIN) {
-                r = p->handle_read();
-                if (r < 0 && !p->is_valid()) {
+              int rd = p->handle_read();
+                if (rd < 0 && !p->is_valid()) {
```
lucasdemarchi, line 286: good catch, thanks

Issue #204, opened 2020-04-02
IPv6 support?


GoMartin300:
Is it possible to forward Mavlink traffic to a IPv6 endpoint?
I have it working now where mavrouter sends traffic to a public IPv4 (University Server), but would prefer not having to reserve an IPv4 and use IPv6 instead.

jbeyerstedt:
Just as a quick teaser: In a project I attend, we'll need to send to an IPv6 multicast address and receive multicast messages as well. So I'm currently working in 2 consecutive pull requests:
- General IPv6 support in mavlink-router
- Adding a new endpoint, which listens on a local port and sends to a pre-defined IP address (multicast or broadcast).

jbeyerstedt:
@GoMartin300 Could this be closed? (I forgot to add a "fixes #xyz" in the commit message.)

GoMartin300:
Judging by the documentation you provided I would say yes, this did not end up being a requirement for my project but I will definitely take advantage of this progress on the next one!

PR #225, opened 2020-09-02
Enable IPv6 Addresses and add Input Validation for IPs

jbeyerstedt:
This PR will enable the use of IPv6 addresses in all endpoints and the command line options. The support for IPv6 is enabled by default, but can be disabled during the project configuration with `--disable-ipv6` (just like the systemd support).

To be able to properly differentiate between IPv4 and IPv6 addresses, this PR will also add input validation for IP addresses. IPv6 addresses must always be enclosed in square brackets, like the usage in URLs. Otherwise the IP:Port notation can't be verified.
Because the config file does use two different keys for IP and port, the angled brackets are not needed technically. But using also them there makes the usage consistent throughout the application and will make the input validation and differentiation between IPv6 and IPv4 easier.

All unicast and multicast IPv6 IP addresses should be supported for sending and basically all addresses for receiving (eavesdropping mode). Link-local unicast and multicast worked just fine in a test between to devices.

This feature was also requested in #204.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ be disabled with --disable-systemd. The default systemd system directory
 is taken via pkg-config. To use another directory update the above
 path, use --with-systemdsystemunitdir.
 
+By default mavlink-router is capable of using IPv6 addresses. In a system without an IPv6
+capable kernel, this can be disabled with --disable-ipv6.
+
 Installation location can be changed using the --prefix option while configuring.
 
 Build:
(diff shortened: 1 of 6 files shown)


jbeyerstedt:
As already noted in the existing issue, I'll add a special endpoint on top of this modification to enabled connecting multiple mavlink-router instances receiving multicast from each other and sending multicast to each other.
Because in my understanding a UDP endpoint can receive multicast packets, but will only send data back to one "client".

dk7xe:
Great! Thank you @jbeyerstedt 
That's a big step forward for using mavlink-router in future UTM communication scenarios.

dk7xe:
@LorenzMeier fyi

lucasdemarchi:
LGTM, thanks

PR #231, opened 2020-10-29
add meson build

coderkalyan:
Adds support for the meson build system.

### DIFF
--- a/config.h.meson.in
+++ b/config.h.meson.in
@@ -1,0 +1,4 @@
+#pragma once
+
+#define VERSION "@version@"
+#define PACKAGE "@package@"
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,0 +1,18 @@
+arm_authorizer = executable('arm-authorizer',
+  ['arm-authorizer.cpp'],
+  include_directories: [mavlink_inc, src_inc],
+  install: true,
+)
+
+heartbeat_print = executable('heartbeat-print',
+  ['heartbeat-print.cpp'],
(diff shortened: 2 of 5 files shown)


lucasdemarchi:
Cool. I tried some time ago but got stuck in a "bug" in meson where it was not possible to include the generated headers from mavlink. Has anything changed recently in meson that allows that?  See branch https://github.com/intel/mavlink-router/commits/meson for what I had.

lucasdemarchi:
This was the issue: https://github.com/mesonbuild/meson/issues/1855

lucasdemarchi:
@coderkalyan any update here ?

coderkalyan:
@lucasdemarchi sorry for the late reply. I encountered similar issues to your attempt, so when I created this PR I circumvented the problem by running a command through meson instead of a generator/custom target. It builds but I know this isn't ideal so I'm working on trying to fix the issue

lucasdemarchi:
@coderkalyan  but I don't see how it solves the problem. If we build multiple times, change the mavlink headers, it will have problems autobuilding, unless we call the specific commands agains.

I'd love to switch to meson, but it seems they are not very friendly to generating headers like we need.

coderkalyan:
@lucasdemarchi Understood, I'll try to dig into this more or ask on the meson support channel/issue tracker.

lucasdemarchi:
@coderkalyan See #291

coderkalyan:
Alright, makes sense. I'll update this when I get a chance.

lucasdemarchi:
See #294

PR #234, opened 2020-11-03
log: use base_mode rather than system_status

### LINKED ISSUE #226: Stopping sending bin logs after failsafe
Mavlink router is stopping sending bin logs over to a companion computer when a failsafe is happening.
using a cube orange on 4.0.3 ArduCopter

lucasdemarchi:
system_status is not a good field to get the "armed" state from: the
flight stack may change the system_status if for example a failsafe
condition is hit.  Derive the "armed" check from the base_mode field
that has a flag exactly with this meaning.

Thanks to Dan Moss for finding the issue.

Fix #226.


lucasdemarchi:
@Dan-Moss @bkueng

bkueng (approved):

Dan-Moss:
LGTM, can this be merged now?

PR #235, opened 2020-11-03
Add a fixed mode to UdpEndpoint

jbeyerstedt:
For Drone-to-Drone cooperative awareness a bi-directional broadcast (or IPv6 multicast) communication is needed, e.g. for the [UTM_GLOBAL_POSITION](https://mavlink.io/en/messages/common.html#UTM_GLOBAL_POSITION) message. The current modes (eavesdropping and normal) are not suitable, because we want to receive packets (which could be done with eavesdropping), but also already send messages before another party has "connected".

This pull-request adds another mode to the UdpEndpoint, where a binding and a target address can be set. If the target address is an IPv6 multicast group, packets addressed to this group will be captured (as well as unicast packets) and outgoing messages will be sent to this group. The binding address is needed to make sure, that outgoing packets are sent through the right interface. IPv4 broadcast also works fine, multicast would need a manual step to add the multicast address to the network interface for receiving packets. A unicast link is also possible.

To make this work, the routing of MAVLink messages had to be adapted. Previously incoming packets were presented to all endpoints, even the one where the message has been received from. This worked, because mavlink-router kept track of the connected systemIDs and used them for loop prevention. With UTM messages, which are all addressed to the MAVLink broadcast systemID, this is not possible any more. Therefore the routing was changes to just present incoming messages to all endpoints besides the one, which received the message in the first place.

Since #225 was merged, please let me know, if I should rebase this branch before merging. Sorry for opening this PR so late, currently I'm busy preparing a demo based on this. At least it's already bench-tested now, and will be flight-tested in the next few weeks.
Some code was inspired by #174 and this PR might also solve that issue.

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -776,6 +776,11 @@ int UdpEndpoint::open(const char *ip, unsigned long port, UdpEndpoint::UdpMode m
         char *ip_str = strdup(&ip[1]);
         ip_str[strlen(ip_str) - 1] = '\0';
 
+        /* remove omittable zeros from IPv6 address */
+        sockaddr_in6 ip_addr;
+        inet_pton(AF_INET6, ip_str, &ip_addr.sin6_addr);
+        inet_ntop(AF_INET6, &(ip_addr.sin6_addr), ip_str, strlen(ip));
+
         /* do some more input validation for multicast mode */
         if (_mode == UdpEndpoint::UdpMode::Multicast) {
             if (nullptr == target_ip) {
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -448,8 +448,8 @@ bool Endpoint::is_ipv6(const char *ip)
 
 bool Endpoint::ipv6_is_linklocal(const char *ip)
 {
-    /* link-local addresses start with fe80 */
-    if (strncmp(ip, "fe80", 4) == 0) {
+    /* link-local addresses start with fe80, ULA addresses are in fc::/7 range */
+    if (strncmp(ip, "fe80", 4) == 0 || strncmp(ip, "fc", 2) == 0 || strncmp(ip, "fd", 2) == 0) {
         return true;
     }
     return false;
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -735,8 +735,9 @@ UdpEndpoint::UdpEndpoint()
 #endif
 }
 
-int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
+int UdpEndpoint::open(const char *ip, unsigned long port, UdpEndpoint::UdpMode mode)
 {
+    this->_mode = mode;
     const int broadcast_val = 1;
 
 #ifdef ENABLE_IPV6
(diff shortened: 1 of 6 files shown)


JonasVautherin (commented):
In my understanding:

* Normal mode says: "I, mavlink-router, want to send UDP packets to `Address`, from any port (chosen by the OS), until I get a packet from _somebody_. Everytime I receive a packet, I send the next messages to the sender of that packet."
* Eavesdropper mode says: "I, mavlink-router, want to listen for UDP packets on `Address`, until I get a packet from _somebody_. Everytime I receive a packet, I send the next messages to the sender of that packet"
* Multicast mode says: "I, mavlink-router, want to listen for UDP packets on `Address`, and send UDP packets to `TargetAddress`. Everytime I receive a packet, I change `Address` to the value of the sender of that packet, which is weird, but not used".

My personal feeling is that modes should be removed in favor of `Address` and `TargetAddress` (or some other name, MAVROS calls them "bind host" and "remote host"). That way one would either set only one, or set both, and that would define the mode. It would make the whole code more readable because there would not be a `sockaddr_out` that is unused in 2 out of the 3 modes, and `sockaddr_in` that is mostly unused in the third one.

Then the could would detect if `TargetAddress` is a broadcast or multicast address and act accordingly. Backward compatibility would be affected for eavesdropping endpoints indeed, but one way would be to throw an error while parsing, saying that `Address` should become `TargetAddress` for eavesdropping endpoints. For normal endpoints, the mode should just be dropped.

I understand how creating a new mode avoids backward compatibility issues, but I would argue that it makes both the code and the config file more complicated (we can now define a node with a wrong combination of `Address`, `TargetAddress` and `Mode`).

Does that make sense?

src/mavlink-router/endpoint.cpp
```diff
@@ -323,11 +323,6 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
         }
     }
 
-    // This endpoint sent the message, we don't want to send it back over the
-    // same channel to avoid loops: reject
-    if (has_sys_comp_id(src_sysid, src_compid))
-        return false;
-
```
JonasVautherin, line 330: I think I missed it in the new code: what prevents a loop from happening now?

jbeyerstedt: It's in the [mainloop.cpp line 146ff.](https://github.com/intel/mavlink-router/pull/235/files#diff-9e3876545c86b9c8b8f34a888a09d5529acba3e9d937d6f1cadb00e8af3567b2R146). I extended the `Mainloop::route_msg()` method to get the endpoint instance, which received the message. When looping through all endpoint to offer them the message, I leave this one out.
It should at least have the same behavior as the comment of the deleted code you cited.

jbeyerstedt: I just figured out, that this filtering is still needed when using redundant data links. So I added that part to the code again.
(But: This might lead to the communication coming to a halt in some cases. I'm currently working on a separate fix for that.)

jbeyerstedt:
You're right about the config file, but I didn't want to do two things at once in this PR. Please comment on #250 for a that topic.

jbeyerstedt:
Lukas highly discourages making breaking changes in the configuration file, so we need to stick with the "mode" flag for now. But I'm not quite happy with the name "multicast" for this mode. Has anyone a better idea?

jbeyerstedt:
I'm not sure any more, if it's actually a good idea to allow sending data to IPv6 multicast groups (or the legacy IP equivalent). MAVLink isn't meant to be used in "big" networks (because system IDs need to be conflict free).

A quickly rebased version of this can be found in https://github.com/consider-it/mavlink-router/tree/feature/broadcast (untested and definitely incomplete), if anybody wants to build on top of this.

jbeyerstedt:
The two commits with fixes to the IPv6 support were already merged in other PRs. I will not work on the actual functionality of this PR any more. Please refer to issue #350 for further discussion and restart with the rebased version in https://github.com/consider-it/mavlink-router/tree/feature/broadcast, if needed.

PR #238, opened 2020-11-12
Add possibility to filter by source component ID

jbeyerstedt:
For a quite special use case, we'll need to filter outgoing messages by source component ID. This commit adds a corresponding config parameter to all endpoints. While implementing this feature, I also realized, that the current message ID filter is only present at the UDP endpoint. To keep everything consistent, I added the message ID filter to all endpoint types.

Our use case is the following:
Multiple telemetry receivers would be distributed on the ground and one or multiple drones fly in the coverage area. All of these receivers should forward the received messages to a central instance of mavlink-router. But transmission from the central router to the base stations should also work. All of them should send out the message, because it is not known which base station can reach the drone.
Because two or more base stations might receive messages from one drone, this will definitely lead to duplicated messages. But this is not the issue. The issue this PR should work around is, that the central mavlink-router instance would forward messages received by one base station to all other connected base stations. Therefore we want to limit the transmission to the base stations to the component ID of ground control software.

A cleaner way would be to have the possibility to group endpoints to a link with multiple failover links and prohibit routing between the endpoints in one group, and/or to enhance the known systems list with some kind of timeout, which cleans up systems from the list. But this would add way more complexity to the system with many possible side-effects.

Another use case of the component ID filtering would be to have multiple systems on the drone connected via mavlink-router but only send messages from the FMU via a (bandwidth-limited) telemetry link.

### DIFF
--- a/src/common/conf_file.cpp
+++ b/src/common/conf_file.cpp
@@ -542,6 +542,30 @@ int ConfFile::parse_uint8_vector(const char *val, size_t val_len, void *storage,
     return 0;
 }
 
+int ConfFile::parse_uint32_vector(const char *val, size_t val_len, void *storage,
+                                  size_t storage_len)
+{
+    assert(val);
+    assert(storage);
+    assert(val_len);
+
+    std::vector<uint32_t> *target;
+    if (storage_len < sizeof(*target)) {
+        return -ENOBUFS;
+    }
+
+    char *filter_string = strndupa(val, val_len);
+    target = static_cast<std::vector<uint32_t> *>(storage);
+
+    char *token = strtok(filter_string, ",");
+    while (token != nullptr) {
+        target->push_back(atol(token)); // we need 32 bit unsigned int, but atoi is signed
+        token = strtok(nullptr, ",");
+    }
+
+    return 0;
+}
+
 int ConfFile::parse_bool(const char *val, size_t val_len, void *storage, size_t storage_len)
 {
     int ival, ret;
--- a/src/common/conf_file.cpp
+++ b/src/common/conf_file.cpp
@@ -504,17 +504,18 @@ int ConfFile::parse_str_buf(const char *val, size_t val_len, void *storage, size
 
 int ConfFile::parse_stdstring(const char *val, size_t val_len, void *storage, size_t storage_len)
 {
-    auto *ptr = (std::string *)storage;
-
     assert(val);
     assert(storage);
     assert(val_len);
 
-    if (storage_len < sizeof(char *)) {
+    std::string *target;
+    if (storage_len < sizeof(*target)) {
         return -ENOBUFS;
     }
 
-    ptr->assign(val, val_len);
+    target = static_cast<std::string *>(storage);
+
+    target->assign(val, val_len);
     return 0;
 }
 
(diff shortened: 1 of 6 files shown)


jbeyerstedt:
I renamed the config option to be more clear about what it should do.

src/common/conf_file.cpp
```diff
@@ -540,6 +540,29 @@ int ConfFile::parse_uint8_vector(const char *val, size_t val_len, void *storage,
...
+    if (storage_len < sizeof(bool)) {
+        return -ENOBUFS;
+    }
+
+    char *filter_string = strndupa(val, val_len);
+    auto *target = (std::vector<uint32_t> *)storage;
+
+    char *token = strtok(filter_string, ",");
+    while (token != nullptr) {
+        target->push_back(atoi(token)); // int should be enough even on 32 bit systems
```
lucasdemarchi, line 559: you are parsing uint32_t here, but atoi returns signed int... there isn't enough room `INT32_MAX + 1` for that, so you should use atol, or strtou

```
$ $ cat /tmp/c.cpp 
#include <stdlib.h>
#include <stdio.h>
#include <inttypes.h>

{
	uint32_t u;

	u = atoi(argv[1]);
	printf("%" PRIu32 "\n", u);

	return 0;
}

$ g++ -m32 -o /tmp/c /tmp/c.cpp
$ /tmp/c 2147483648
2147483647

```

src/common/conf_file.h
```diff
@@ -125,6 +125,8 @@ class ConfFile {
     static int parse_stdstring(const char *val, size_t val_len, void *storage, size_t storage_len);
     static int parse_uint8_vector(const char *val, size_t val_len, void *storage,
                                   size_t storage_len);
+    static int parse_uint32_vector(const char *val, size_t val_len, void *storage,
+                                   size_t storage_len);
```
lucasdemarchi, line 129: please split the addition of parse_uint32_vector() to a separate commit

src/endpoint.h
```diff
@@ -51,15 +51,15 @@ struct UdpEndpointConfig {
...
+    std::vector<uint32_t> allow_msg_id_out;
 };
 
 struct TcpEndpointConfig {
     std::string name;
     std::string address;
     unsigned long port;
     int retry_timeout{5};
-    std::vector<uint8_t> allow_msg_id_out;
+    std::vector<uint32_t> allow_msg_id_out;
```
lucasdemarchi, line 64: the fix in this commit to change uint8_t -> uint32_t is good

lucasdemarchi (commented):
Only checked the first commit yet, with the fix

src/endpoint.h
```diff
@@ -194,7 +199,7 @@ class UartEndpoint : public Endpoint {
 
     bool setup(UartEndpointConfig config); ///< open UART device and apply config
 
-    static const ConfFile::OptionsTable option_table[4];
+    static const ConfFile::OptionsTable option_table[5];
```
lucasdemarchi, line 202: eventually we need to get rid of this number here

lucasdemarchi:
Second commit LGTM.

src/common/conf_file.cpp
jbeyerstedt: It should be fine in that place because the MAVLink system ID is just 24 bit, but the method should be capable or uint32, if it is called that way.

src/endpoint.h
jbeyerstedt: Using constexpr doesn't work (at least before C++17) and just removing the number breaks any sizeof() call. So I don't like that number either, but haven't found a solution yet.

src/common/conf_file.h
jbeyerstedt: done

src/common/conf_file.cpp
jbeyerstedt: changed

src/common/conf_file.cpp
```diff
@@ -540,6 +540,29 @@ int ConfFile::parse_uint8_vector(const char *val, size_t val_len, void *storage,
...
 }
 
+int ConfFile::parse_uint32_vector(const char *val, size_t val_len, void *storage,
+                                  size_t storage_len)
+{
+    assert(val);
+    assert(storage);
+    assert(val_len);
+
+    if (storage_len < sizeof(bool)) {
```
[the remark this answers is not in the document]
jbeyerstedt: Done in a new, separated commit.

PR #242, opened 2020-11-27
Dynamic endpoints, mainloop refactoring, coalescing

jkflying:
This brings upstream a number of features we've developed at Auterion.

* Dynamic endpoints, which can be added or removed at runtime via a named pipe
* UDP packet coalescing, to reduce packet overhead for certain transport types
* Refactoring of the mainloop, to allow it to be run one iteration at a time to enable unit tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -147,14 +147,26 @@ mainloop_test_SOURCES = \
 	src/common/log.h \
 	src/common/util.c \
 	src/common/util.h \
+	src/common/xtermios.cpp \
+	src/common/xtermios.h \
+	src/mavlink-router/autolog.cpp \
+	src/mavlink-router/autolog.h \
+	src/mavlink-router/binlog.cpp \
+	src/mavlink-router/binlog.h \
+	src/mavlink-router/comm.h \
+	src/mavlink-router/endpoint.h \
 	src/mavlink-router/endpoint.cpp \
 	src/mavlink-router/endpoint.h \
+	src/mavlink-router/logendpoint.cpp \
+	src/mavlink-router/logendpoint.h \
 	src/mavlink-router/mainloop_test.cpp \
 	src/mavlink-router/mainloop.cpp \
 	src/mavlink-router/pollable.cpp \
 	src/mavlink-router/pollable.h \
 	src/mavlink-router/timeout.cpp \
-	src/mavlink-router/timeout.h
+	src/mavlink-router/timeout.h \
+	src/mavlink-router/ulog.h \
+	src/mavlink-router/ulog.cpp
 mainloop_test_LDADD = $(GTEST_LIBS)
 
 # ------------------------------------------------------------------------------
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -329,8 +329,8 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     if (has_sys_comp_id(src_sysid, src_compid))
         return false;
 
-    if (msg_id != UINT32_MAX && 
-        _message_filter.size() > 0 && 
+    if (msg_id != UINT32_MAX &&
+        _message_filter.size() > 0 &&
         std::find(_message_filter.begin(), _message_filter.end(), msg_id) == _message_filter.end()) {
 
         // if filter is defined and message is not in the set then discard it
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -353,6 +353,22 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     return false;
 }
 
+void Endpoint::postprocess_msg(int target_sysid, int target_compid, uint8_t src_sysid,
+                               uint8_t src_compid, uint32_t msg_id)
+{
+    // avoid unused variable messages
+    (void)target_sysid;
+    (void)target_compid;
+    (void)src_sysid;
+    (void)src_compid;
+
+    if (msg_id != UINT32_MAX && _message_nodelay.size() > 0 &&
+        std::find(_message_nodelay.begin(), _message_nodelay.end(), msg_id) != _message_nodelay.end()) {
+        // if filter is defined and message is not in the set then discard it
+        flush_pending_msgs();
+    }
+}
+
 bool Endpoint::_check_crc(const mavlink_msg_entry_t *msg_entry)
 {
     const bool mavlink2 = rx_buf.data[0] == MAVLINK_STX;
@@ -759,12 +775,31 @@ int UartEndpoint::add_speeds(std::vector<unsigned long> bauds)
 }
 
 UdpEndpoint::UdpEndpoint()
-    : Endpoint{"UDP"}
+    : Endpoint{"UDP"},
+    _write_scheduled(false),
+    _max_packet_size(0),
+    _max_timeout_ms(0)
 {
     bzero(&sockaddr, sizeof(sockaddr));
 #ifdef ENABLE_IPV6
     bzero(&sockaddr6, sizeof(sockaddr6));
 #endif
+    _write_schedule_timer = Mainloop::get_instance().add_timeout(
+        _max_timeout_ms, [this](void*)
+            {
+                flush_pending_msgs();
+                return true;
+            },
+        this);
+}
+
+
+UdpEndpoint::~UdpEndpoint()
+{
+    if (_write_schedule_timer) {
+        Mainloop::get_instance().del_timeout(_write_schedule_timer);
+        _write_schedule_timer = nullptr;
+    }
 }
 
 int UdpEndpoint::open(const char *ip, unsigned long port, bool to_bind)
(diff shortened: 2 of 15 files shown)


src/mavlink-router/endpoint.cpp
```diff
@@ -965,13 +1071,8 @@ int TcpEndpoint::accept(int listener_fd)
...
 int TcpEndpoint::open(const char *ip, unsigned long port)
 {
-    if (!_ip || strcmp(ip, _ip)) {
-        free(_ip);
-        _ip = strdup(ip);
-        _port = port;
-    }
-
-    assert_or_return(_ip, -ENOMEM);
+    _ip = ip;
```
jkflying, line 1074: Double check this

src/mavlink-router/mainloop.cpp
```diff
@@ -63,6 +66,8 @@ Mainloop &Mainloop::init()
     return _instance;
 }
 
+static const char* pipe_path = "/tmp/mavlink_router_pipe";
```
lucasdemarchi: /tmp is usually not the proper place for a named pipe. /var/run/mavlink-router would probably be better.

src/mavlink-router/mainloop.cpp
```diff
@@ -151,6 +167,15 @@ void Mainloop::route_msg(struct buffer *buf, int target_sysid, int target_compid
...
     }
 
+    for (auto i: dynamic_endpoints) {
+        if (i.second->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid, msg_id)) {
+            log_debug("Endpoint [%d] accepted message to %d/%d from %u/%u", i.second->fd, target_sysid,
+                      target_compid, sender_sysid, sender_compid);
+            write_msg(i.second, buf);
+            unknown = false;
+        }
+    }
```
lucasdemarchi, line 164: these 3 loops should probably be consolidated later

src/mavlink-router/mainloop.cpp
```diff
@@ -273,9 +298,12 @@ void Mainloop::loop()
...
         if (r < 0 && errno == EINTR)
             continue;
-
         for (i = 0; i < r; i++) {
-            if (events[i].data.ptr == &g_tcp_fd) {
+            if (events[i].data.ptr == &pipefd) {
+                _handle_pipe();
+                continue;
+            }
+            else if (events[i].data.ptr == &g_tcp_fd) {
```
lucasdemarchi: coding style... else should be in the same line

src/mavlink-router/mainloop.cpp
```diff
@@ -80,6 +85,17 @@ int Mainloop::open()
         return -1;
     }
 
+    if (mkfifo(pipe_path, 0777) == -1) {
+        log_error("%m");
+    }
+
+    pipefd = ::open(pipe_path, O_RDWR);
```
lucasdemarchi: O_NONBLOCK?

src/mavlink-router/mainloop.cpp
```diff
@@ -358,6 +388,23 @@ static bool _print_statistics_timeout_cb(void *data)
...
 }
 
+bool Mainloop::_add_dynamic_endpoint(std::string name, Endpoint *endpoint)
+{
+    if (!endpoint) {
+        return false;
+    }
+    log_debug("Adding dynamic endpoint: %s", name.c_str());
+    auto ep = dynamic_endpoints.find(name);
+    if (ep != dynamic_endpoints.end()) {
```
lucasdemarchi: what's the use case of replacing an endpoint by name?

lucasdemarchi (changes requested):
Not much fan of defining a protocol on named pipes since it's easy to break other apps, but this seems simple enough.

lucasdemarchi (changes requested):
Also, could you expand the commit message? What is the build fix?

lucasdemarchi:
I started reviewing the commits, but it seems they are not split correctly. Could this be rebased with the split per commit fixed/redone?

jkflying:
Hi Lucas, thanks for the review so far, I'll incorporate the suggestions. The commits are a bit messy because I'm trying to keep to the original ordering / attribution. I'll see if I can clean things up a little to make it easier to review by commit though. There will also be a few follow up PRs.

lucasdemarchi:
@jkflying  thanks, the way it is split is pretty hard to review each commit.

Another thing that I was thinking... why is the named pipe important?

- There is already tcp server, so each client that connects and is accepted, has its own "dynamic" endpoint. The difference in the end would be the communication done over tcp vs udp. However from beginning it was thought to have this server for connections on *localhost*, in which the diff between tcp and upd won't matter
- Even if it matters, then a unix socket would be preferred to allow multiple clients without concurrency issues. It's also the usual way (besides bus approaches as dbus or the like) to communicate with daemons. We could later extend it later to do things like getting stats, etc

jkflying:
> The difference in the end would be the communication done over tcp vs udp

This is the issue, over lossy links with variable bandwidth having TCP isn't a viable solution for 'dynamic' endpoints due to the degradation characteristics.

lucasdemarchi:
> This is the issue, over lossy links with variable bandwidth having TCP isn't a viable solution for 'dynamic' endpoints due to the degradation characteristics.

but you are still controlling it to open a udp on localhost. If you want the dynamic endpoint feature to external clients, then either having unix socket (still requiring local access), or extend the TCP server to accept commands to create/destroy endpoints would be preferred.

jkflying:
Ah, I believe we are talking past each other. You mean using TCP instead of a pipe for the control of the endpoints, yes? That does sound like a good idea, the issue is I don't have the time to do that right now :-/ Can we leave that for a future extension/migration path?

I think it should be fairly easy on the other side to change pipe writes to `curl` commands as well (thinking of internal migration path here, we have several subsystems that use this pipe now). There would also need to be some whitelisting on the TCP side to make sure configs that don't want off-system reconfiguration can ignore remote connections.

src/mavlink-router/mainloop.cpp
jkflying: So, the reason it is in /tmp/ is so that for development purposes you don't need to run it as root or do any special setup. Thoughts? I guess this points more strongly towards using a TCP port...

src/mavlink-router/endpoint.cpp
jkflying: _ip is now a std::string so this is good

jkflying:
@lucasdemarchi I've cleaned up / squashed / reordered the commits here, so it should be easier to review now.

Makefile.am
```diff
@@ -147,14 +147,26 @@ mainloop_test_SOURCES = \
...
 	src/common/util.c \
 	src/common/util.h \
+	src/common/xtermios.cpp \
+	src/common/xtermios.h \
+	src/mavlink-router/autolog.cpp \
+	src/mavlink-router/autolog.h \
+	src/mavlink-router/binlog.cpp \
+	src/mavlink-router/binlog.h \
+	src/mavlink-router/comm.h \
+	src/mavlink-router/endpoint.h \
```
jbeyerstedt, line 157: This line is added twice

jbeyerstedt (changes requested):
This generally looks good to me, but I'm missing documentation on:
- the protocol to add dynamically add UDP endpoints
- the coalescence parameters (add some words in the example config, as it's done for everything else)

P.S. We need a better way to document the parameters than the example config. But that's another topic.

src/mavlink-router/mainloop.cpp
lucasdemarchi: the setup is one time only: create the dir with the appropriate perms. We can also use configure to set it to another place for convenience for developers... Example:  ./configure  --rundir=/tmp

lucasdemarchi: Also, as I said in previous version of this patch, we should not use a pipe.. it should rather be implemented with a unix/tcp socket.

src/mavlink-router/mainloop.cpp
```diff
@@ -564,12 +634,28 @@ Timeout *Mainloop::add_timeout(uint32_t timeout_msec, std::function<bool(void*)>
...
+    if (!t) {
+        return;
+    }
+
+    struct itimerspec ts;
+    ts.it_interval.tv_sec = timeout_msec / MSEC_PER_SEC;
+    ts.it_interval.tv_nsec = (timeout_msec % MSEC_PER_SEC) * NSEC_PER_MSEC;
+    ts.it_value.tv_sec = ts.it_interval.tv_sec;
+    ts.it_value.tv_nsec = ts.it_interval.tv_nsec;
+    timerfd_settime(t->fd, 0, &ts, nullptr);
```
lucasdemarchi, line 652: This should be a method of Timeout, not the mainloop. add_timeout() is the only method that is capable of creating a Timeout object since it also adds it to the mainloop, but setting a timeout should be perfectly possible from the Timeout object itself.

lucasdemarchi (changes requested):

lucasdemarchi:
@jkflying are you still interested in this?  Not sure what to do since it accumulated several conflicts. I myself can't review it if there are multiple changes bundled together.

For controlling and adding dynamic endpoints my main feedback was (and is) that a pipe is not acceptable: it should be an unix socket.

lucasdemarchi:
Closing for inactivity. Feel free to open a new one if this is desired.

PR #253, opened 2021-03-04
github actions: Add artifacts and release binaries 

patrickelectric:
Allow easy access of the binaries for development, test and also for official releases

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,14 +1,6 @@
 name: Build and Test
 
-on:
-  push:
-    branches:
-    - 'master'
-    tags:
-    - 'v*'
-  pull_request:
-    branches:
-    - '*'
+on: [push, pull_request]
 
 jobs:
   ubuntu:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,20 @@ jobs:
         run: make -j
       - name: test
         run: make check PYTHON=python3
+      - name: Rename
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -1,14 +1,6 @@
...
-on:
-  push:
-    branches:
-    - 'master'
-    tags:
-    - 'v*'
-  pull_request:
-    branches:
-    - '*'
+on: [push, pull_request]
```
lucasdemarchi, line 3: but then you would upload any binary on pushes to branches != master. This doesn't seem right

patrickelectric: Artifact upload will happen for all cases, artifacts are something that CI will produce but will only be available inside the CI UI: Actions -> specific build -> artifact will be there.

For github releases, it'll depends of what kind of push was done, if it's tag, an official binary will be available, otherwise a binary with "prerelease" mark will be created.
`if: ${{ github.event_name == 'push' }}`, `prerelease: ${{ !startsWith(github.ref, 'refs/tags/') }}`
I can remove prerelease binaries if that's something problematic.

patrickelectric: ping @lucasdemarchi

lucasdemarchi: artifacts take space and are eventually blocked. I don't know how much space it is, but I hit the max in another internal project. I'd prefer that we only upload artifacts for master branch and for pull requests, but not for random branches

patrickelectric: @lucasdemarchi done.

PR #262, opened 2021-04-20
Update mavlink

JonasVautherin:
Just updating the mavlink submodule, which is getting dated and drops e.g. gimbal v2 messages :blush:


JonasVautherin:
Whoops, somehow I missed something in my branch. Sorry for the noise

Issue #252, opened 2021-02-12
Closed connection to TCP port crashes mavlink-router


jbeyerstedt:
I was trying to use pymavlink to analyze a MAVLink message. To get the message, I tried to connect to the TCP port (`TcpServerPort` parameter) of mavlink-router.

But when my python script terminates, mavlink-router will stop working, because it crashed after the TCP connection was lost. As far as I understood, the TCP server should be able to even handle multiple clients, so one client disconnecting shouldn't lead to mavlink-router terminating, right?

The verbose output looks like this (output shortened):
```
[...]

TCP connection [9] accepted
Accepted TCP connection on [9]
UDP [5] got 30 bytes
Endpoint [4] got message 74 to -1/-1 from 23/1
	Known components:
		23/1
Endpoint [6] got message 74 to -1/-1 from 23/1
	Known components:
		255/190
Endpoint [6] accepted message 74 to -1/-1 from 23/1
UDP [6] wrote 30 bytes
Endpoint [9] got message 74 to -1/-1 from 23/1
	Known components:
Endpoint [9] accepted message 74 to -1/-1 from 23/1
TCP [9] wrote 30 bytes

[...]

UDP [4] got 13 bytes
Endpoint [5] got message 42 to -1/-1 from 23/1
	Known components:
		23/1
Endpoint [6] got message 42 to -1/-1 from 23/1
	Known components:
		255/190
Endpoint [6] accepted message 42 to -1/-1 from 23/1
UDP [6] wrote 13 bytes
Endpoint [9] got message 42 to -1/-1 from 23/1
	Known components:
Endpoint [9] accepted message 42 to -1/-1 from 23/1
TCP [9] wrote 13 bytes
UDP [5] got 13 bytes
poll error for fd 9, closing it
TCP Connection [9] closed
```

The second part of this issue is, that the return code is zero, so the systemd unit does not re-start mavlink-router after the crash.

lucasdemarchi:
> But when my python script terminates, mavlink-router will stop working, because it crashed after the TCP connection was lost. As far as I understood, the TCP server should be able to even handle multiple clients, so one client disconnecting shouldn't lead to mavlink-router terminating, right?

right, the bug sounds legit



> The second part of this issue is, that the return code is zero, so the systemd unit does not re-start mavlink-router after the crash.

yep, another bug

windelbouwman:
@jbeyerstedt I can confirm this issue exists. Do you have some minimal example script to trigger this bug?

[...]

PR #264, opened 2021-04-29
Add notion of critical endpoint. Quit on error on critical endpoints.

windelbouwman:
Related to issue #252 .

This proposal adds a boolean to indicate if a connection is critical or not. When a serial port is unplugged, mavlink-router should quit, but when a tcp connection breaks down, this is okay.

### DIFF
--- a/src/mavlink-router/mainloop.cpp
+++ b/src/mavlink-router/mainloop.cpp
@@ -310,16 +310,12 @@ int Mainloop::loop()
             if (events[i].events & EPOLLERR) {
                 log_error("poll error for fd %i", p->fd);
 
-                /*
-                 * TCP Pollables can be added/removed dynamically and
-                 * is_valid() is set when it becomes not available anymore.
-                 * Otherwise it's a unkonwn error or another bus has a
-                 * unexpected disconnect (e.g. when removing a usb-serial
-                 * device), so make it fatal: external components may restart
-                 * mavlink-router
-                 */
-                if (p->is_valid())
+                if (p->is_critical()) {
+                    log_error("Critical fd %i got error, exiting", p->fd);
                     request_exit(EXIT_FAILURE);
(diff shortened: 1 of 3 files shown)


src/mavlink-router/mainloop.cpp
```diff
@@ -310,16 +310,12 @@ int Mainloop::loop()
...
 
-                /*
-                 * TCP Pollables can be added/removed dynamically and
-                 * is_valid() is set when it becomes not available anymore.
-                 * Otherwise it's a unkonwn error or another bus has a
-                 * unexpected disconnect (e.g. when removing a usb-serial
-                 * device), so make it fatal: external components may restart
-                 * mavlink-router
-                 */
-                if (p->is_valid())
```
lucasdemarchi, line 321: the idea here was to reuse is_valid() because TCP endpoint is the only one to override this method and set it to  false if it got an error.

I would understand this solution if we were to move the EPOLLERR to be handled before EPOLLOUT/EPOLLIN. However since we are leaving it here, how is this different from what we currently have?

windelbouwman: Yes, I understood that `is_valid` was re-used. And, as you said, EPOLLOUT/EPOLLIN in handled before EPOLLERR, so I agree with your reasoning that when a socket is closed, the following would be expected:
- EPOLLIN should mark the fd as invalid
- EPOLLERR sees the invalid fd, and skips request exit

However, this is not what happens. During my testing, I discovered different behavior. Sometimes, EPOLLERR is handled without EPOLLIN. EPOLLIN happens the next iteration of the mainloop. So sometimes a closed TCP connection triggers EPOLLIN+EPOLLERR, sometimes only EPOLLERR. I do not know why this happens, but for this reason is_valid cannot be reused.

PR #266, opened 2021-05-09
Disable Nagle Algorithm on TCP sockets to reduce transport latency

soof-golan:
To reduce the latency introduced by TCP socket in real time systems it is recommended to disable aggregation on the transport level.
While aggregation is on the TCP endpoint Introduced latencies of up to 50 milliseconds to some of the mavlink message.
While aggregation is off the TCP endpoint Introduced latencies of 2-3 milliseconds consistently.

Further Reading:
https://en.wikipedia.org/wiki/Nagle%27s_algorithm

### DIFF
--- a/src/mavlink-router/endpoint.cpp
+++ b/src/mavlink-router/endpoint.cpp
@@ -23,6 +23,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <net/if.h>
+#include <netinet/tcp.h>
 #include <ifaddrs.h>
 #include <netdb.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
This looks good, but I'm not sure if we should enable this unconditionally. Not every endpoint need the telemetry info in real time and would rather profit from the increased throughput for leaving it as it was.  I will leave this open a few days for comments on whether we need/want a setting in the config for that.

soof-golan:
Since MAVLINK is usually considered low throughput compared to the common TCP link bandwidth, I'm not sure it would create a considerable overhead that should call for a config option.
If you'll decide a config is the way to go, is there a best practice on how to reference the config file?

soof-golan:
@lucasdemarchi Is there anything delaying the PR I should be aware of?
Any fixes you think are required?

PR #268, opened 2021-05-25
configure: Add git-describe information in version

patrickelectric:
Necessary information for people that are using versions
that are not a major release.

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>


patrickelectric:
ping @lucasdemarchi

configure.ac
```diff
@@ -1,6 +1,6 @@
 AC_PREREQ(2.64)
 AC_INIT([mavlink-router],
-	[2],
+	[2 (m4_esyscmd_s([git describe --always]))],
```
lucasdemarchi, line 3: This doesn't work unless you are doing a clean build. It also breaks builds on released tarballs that are not on a git repo. The usual way to do this is actually to add a .c that is always recompiled. Then you can regenerate the .c (or a .h by the build system). I don't have an example handy using autotools, but the idea is the same as we do in ardupilot (see how the ap_version.h is generated).

patrickelectric: Let me know what you think about the update.
It creates a variable in the config.h file for the build string information.

lucasdemarchi: This still doesn't work. The problem is you are doing all this at the **configure** phase rather than the **build** phase. In other words, you need a rule in Makefile.am, not something in configure.ac to update this header. config.h is produced by configure script to reflect the **config** options.

See how it doesn't work:

```console
$ git fetch origin refs/pull/268/head
$ git checkout FETCH_HEAD
$ git clean -fxd
$ ./autogen.sh c
$ make
$ strings mavlink-routerd | grep "Version"
mavlink-routerVersion v2-43-g7a049fd from 2021-06-15
$ git describe HEAD
v2-43-g7a049fd
```

So far so good. But:

```console
$ touch a && git add a && git commit -m tmp
$ make
$ strings mavlink-routerd | grep "Version"
mavlink-routerVersion v2-43-g7a049fd from 2021-06-15
$ git describe HEAD
v2-44-g2fe822c
```

Even if we change something that could affect the final binary, it doesn't work:

```console
$ echo "" >> src/mavlink-router/main.cpp && git commit  -a -m tmp 
$ make
$ strings mavlink-routerd | grep "Version"
mavlink-routerVersion v2-43-g7a049fd from 2021-06-15
$ git describe HEAD
v2-45-g38f1dff
```

PR #270, opened 2021-06-06
Add .deb package build to Github Actions

bbworld1:
This commit adds support for building a .deb archive for installation on Debian-derived systems.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,11 +57,81 @@ jobs:
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
           overwrite: true
       - uses: svenstaro/upload-release-action@v2
-        name: Upload binaries to release
+        name: Upload .debs to release
         if: ${{ startsWith(github.ref, 'refs/tags/') }}
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: mavlink-router_amd64.deb
+          tag: ${{ github.ref }}
+          prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+          overwrite: true
+
+  ubuntu-crossbuild:
+    name: ubuntu-${{ matrix.name }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - name: armhf
+            arch_name: linux-armv7
+          - name: arm64
+            arch_name: linux-arm64
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: recursive
+      - name: setup dockcross
+        run: docker run --rm dockcross/${{ matrix.arch_name }} > ./dockcross; chmod +x ./dockcross
+      - name: configure
+        run: ./dockcross /bin/bash -c "./autogen.sh c --with-systemdsystemunitdir=/usr/lib/systemd/system --host=arm"
+      - name: build
+        run: ./dockcross /bin/bash -c "make -j"
+      - name: test
+        run: ./dockcross /bin/bash -c "make check PYTHON=python3"
+      - name: Rename
+        run: cp mavlink-routerd mavlink-routerd-libc-${{ matrix.name }}
+      - name: package-deb
+        run: |
+          mkdir -p .debpkg/usr/bin
+          cp mavlink-routerd .debpkg/usr/bin/mavlink-routerd
+          mkdir -p .debpkg/DEBIAN
+          export VERSION=$([[ "${{ github.ref }}" == refs/tags/* ]] && echo $(echo ${{ github.ref }} | sed s#"refs/tags/v"##) || echo "0.0.0")
+          export PKGSIZE=$(du mavlink-routerd)
+          echo "Building mavlink-router version: ${VERSION}"
+          cat > .debpkg/DEBIAN/control << EOL
+          Package: mavlink-router
+          Version: ${VERSION}
+          Section: custom
+          Priority: optional
+          Architecture: ${{ matrix.name }}
+          Essential: no
+          Installed-Size: ${PKGSIZE}
+          Maintainer: Dronecode Project
+          Description: Route mavlink packets between endpoints
+          EOL
+          dpkg-deb --build .debpkg
+          mv .debpkg.deb mavlink-router_${{ matrix.name }}.deb
+      - uses: actions/upload-artifact@master
+        with:
+          path: mavlink-routerd-libc-${{ matrix.name }}
+      - uses: actions/upload-artifact@master
+        with:
+          path: mavlink-router_${{ matrix.name }}.deb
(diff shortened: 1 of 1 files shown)


auturgy:
Is AMD64 the best target arch? Most mavlink-router deployments are embedded, suggesting arm/64 might better support users.

bbworld1:
AMD64 was just what was already built in the pipeline, so I just packaged that. ARM64 I think does require some cross compiling (all Github Actions instances are amd64), but it is a good idea to implement. I'll push a change soon to support building ARM packages.

bbworld1:
@auturgy The pipeline now builds packages for the amd64, arm64 and armhf platforms. See here: https://github.com/bbworld1/mavlink-router/releases/tag/v2.5

auturgy:
Great, thanks.

.github/workflows/main.yml
```diff
@@ -20,9 +20,33 @@ jobs:
         run: make check PYTHON=python3
       - name: Rename
         run: cp mavlink-routerd mavlink-routerd-libc-x86-64
+      - name: package-deb
+        run: |
+          mkdir -p .debpkg/usr/bin
+          cp mavlink-routerd .debpkg/usr/bin/mavlink-routerd
```
lucasdemarchi, line 26: we should rather be using `make DESTDIR=$PWD/.debpkg/ install`

.github/workflows/main.yml
```diff
@@ -20,9 +20,33 @@ jobs:
...
+          export PKGSIZE=$(du mavlink-routerd)
+          echo "Building mavlink-router version: ${VERSION}"
+          cat > .debpkg/DEBIAN/control << EOL
+          Package: mavlink-router
+          Version: ${VERSION}
+          Section: custom
+          Priority: optional
+          Architecture: amd64
+          Essential: no
+          Installed-Size: ${PKGSIZE}
```
lucasdemarchi, line 38: is this really needed?

.github/workflows/main.yml
```diff
@@ -20,9 +20,33 @@ jobs:
...
+          Maintainer: Dronecode Project
```
lucasdemarchi, line 39: mavlink-router is not part of dronecode, nor dronecode controls. "mavlink-router developers" should be sufficient here

.github/workflows/main.yml
```diff
@@ -32,6 +56,85 @@ jobs:
...
+      - name: build
+        run: ./dockcross /bin/bash -c "make -j"
+      - name: test
+        run: ./dockcross /bin/bash -c "make check PYTHON=python3"
+      - name: Rename
+        run: cp mavlink-routerd mavlink-routerd-libc-${{ matrix.name }}
+      - name: package-deb
+        run: |
+          mkdir -p .debpkg/usr/bin
+          cp mavlink-routerd .debpkg/usr/bin/mavlink-routerd
```
lucasdemarchi, line 96: ditto, please use "make install"

lucasdemarchi (commented):
thanks for working on this. I left some comments.

Another problem is that although it is said to be "generic .deb", it's not really that: it's built on ubuntu-latest (whatever that is in github actions) and it's valid on ubuntu-latest only. There is no guarantee it will work in another version or another distro that uses .deb packages. Maybe we should be more explicit on what we are building

.github/workflows/main.yml
bbworld1: I assume architecture is definitely a needed field. The other fields, if I recall correctly, are also recommended by Debian.

lucasdemarchi: My comment was in the line just above, "Installed-Size". This should be calculated by the debian tooling, shouldn't it?

bbworld1: It doesn't appear to be calculated from what I can gather. Omitting it and then building leads to apt reporting an installed size estimate of 0B.

bbworld1:
Thanks for the feedback. I've pushed changes to address these points.
- The Maintainer has been changed to MAVLink-Router Developers. Apologies for the mistake.
- make install is now used to install to debpkg location.
- I've renamed the built debs to ubuntu-${DISTRO}_${ARCH}, to be more specific about which packages are intended for which distros. I've also added support for building both Ubuntu 20.04 and Ubuntu 18.04 packages, and simplified the workflow by combining the previously separate amd64 build and the arm builds into one matrix.

Please let me know if there's any more issues.

PS: I'm not fully sure what to do with the binary artifacts, so I just kept the ones built on 20.04. If we want to do something else please let me know.

lucasdemarchi:
@bbworld1 that is really nice, thanks for putting this together. As someone not familiar with dockcross, I'm not sure I understand the isolation of the toolchain from the host distro. Suppose we have an external dependency, how is the cross-compilation going to work without having that dep in the container that is building mavlink-router?

bbworld1:
Apologies for the late response. As far as using traditional Debian/Ubuntu library packages (lib<x>-dev), yes, that would require having the dependency in the container. Submodule dependencies etc. however should just work; they will need to be compiled from source, however. I actually came across dockcross in another project - MAVSDK - which depends on gRPC; their solution is to use CMake to pull in a remote copy of the source and just build it. If we ever needed to pull in dependencies for mavlink-router a similar approach could likely be followed. The consequence is the library generally has to be statically linked in the end, but I don't see this being a major problem for a tool like mavlink-router which has minimal dependencies and is largely standalone in the first place.

Also, I've just realized that since the binaries are built inside dockcross, they link against dockcross libraries (which are based on Debian 10). The generated binaries for 18.04 and 20.04 are the same, so there's not really much point in building separate ones for each.

What do you think about exporting one statically-linked binary (so it won't depend on dynamic libraries) and packaging one generic deb for each architecture?

mrpollo:
Whenever this is merged can we also have a GitHub Action to push the Debian packages to the repository registry? It will make it super easy for anyone to find them

lucasdemarchi:
@bbworld1 I'm against doing a statically-built binary. If we have more dependencies we may look at one of the following:
1)  install each distro package manually so we link against the correct library
2) adopt another packaging solution like appimage or flatpak
3) gather the dependencies manually

I ended up with (3) in https://github.com/arduPilot/dema-rc by using a script to find the dependencies. But I did that due to the minimal amount of space needed and not having a proper distro as base.

I don't want to hold this. Right now we don't really have additional dependencies, so I'm good with it. But we may need to find another solution in the future.

lucasdemarchi:
> Whenever this is merged can we also have a GitHub Action to push the Debian packages to the repository registry? It will make it super easy for anyone to find them

We'd need to create a ppa/overlay-repo somewhere for that to work. Proper packaging would actually be to submit the project to the specific distros asking for inclusion.

mrpollo:
@lucasdemarchi you can publish them to this repository package registry it's free for OSS projects 

https://github.com/features/packages

PR #271, opened 2021-06-29
github: Add ARM-v6 builds

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,3 +143,33 @@ jobs:
           tag: ${{ github.ref }}
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
           overwrite: true
+
+  arm-v6:
+    name: ARM v6 binary
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -143,3 +143,33 @@ jobs:
           tag: ${{ github.ref }}
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
           overwrite: true
+
+  arm-v6:
+    name: ARM v6 binary
```
lucasdemarchi, line 148: Just above we have:
```
run: docker run --rm dockcross/${{ matrix.arch_name }} > ./dockcross; chmod +x ./dockcross
```

And apparently it supports armv6:
https://github.com/dockcross/dockcross/tree/master/linux-armv6

Can we add it to the matrix of distros instead of duplicating this?

lucasdemarchi: @bbworld1 any idea?

patrickelectric: Oh, I was not rebased over the latest master, so I was not aware that armv7 builds are already available, going to close the PR.

PR #276, opened 2021-07-29
Fix warnings related to USB latency setting

lucasdemarchi:


lucasdemarchi:
This was introduced in #165. Adding some people here: @catch-twenty-two, @jkflying, @LorenzMeier

PR #279, opened 2021-09-04
README: expand package list for other distros

lucasdemarchi:
Add some common distros: Debian, Ubuntu, Fedora and Archlinux.


jbeyerstedt (approved):
Looks good to me (but haven't had the time to test on a fresh installation)

lucasdemarchi:
I tested all of these on docker (also reason why I have to install systemd).

PR #294, opened 2021-11-01
build: switch to meson

lucasdemarchi:
Add a minimal meson configuration that can build the current source
code.


lucasdemarchi:
This supersedes #231

lucasdemarchi:
Humn.. not sure what to do with those dockcross failures. We need to add some cross-file.txt to work with meson, but then there is the question about the additional libraries. The (docker) rootfs is not for the foreign arch, so it isn't clear how the additional (foreign) libraries would be installed... Either figure out if the base distro for those cross toolchains support multi-arch or.... get rid of those from CI. I wonder if they are actually being used since we upload the binaries without any library. And the binaries are dynamically linked. So it can only work if whoever is using those binraries are using the same distro as what is used by dockcross.

lucasdemarchi:
@patrickelectric  are you using these binaries? Can we migrate to something else?

lucasdemarchi:
@bbworld1 ^ too

bbworld1:
I'm not using those binaries at the moment, so I'm actually OK with migrating to something else. Would like to figure out some other method of cleanly compiling and releasing them other than dockcross, that's just what I found originally.

lucasdemarchi:
@bbworld1 thanks. I think the role dockcross is serving could be accomplished by a simple "run on container + install cross toolchain", which is achievable with debian/ubuntu. And with those we would at least make clear where these packages are expected to install.

lucasdemarchi:
Ok, after a lot of experiments I think I settled on a solution:

1. We use the cross-toolchain from the host. Developers can also use any other cross-toolchain they please. This is working for me on both ubuntu and archlinux
2. For dependencies we may have we use a rootfs to serve as the sysroot for the toolchain. There are several methods for creating such rootfs: mkosi, buildroot, yocto, debootstrap, etc and the "container tools" like docker, podman. Given that on CI the container integration is very easy, I tried that. I was initially thinking that the rootfs creation that way could delay the build a lot, but that is not what I see after implementing it: developers need to do it only once and on CI it's pretty fast too.
3. Since the rootfs is in the foreign architecture, any dependency we have can be easily installed there (with the help of qemu to run the binaries in the container).

PR #296, opened 2021-11-12
Run clang-format on everything, add new CI workflow to check

jbeyerstedt:
This PR introduces a few new settings in the clang-format file, some of those need at least clang-format 9:
- AlignConsecutiveDeclarations, AlignConsecutiveAssignments, AlignConsecutiveMacros: Specify as already used in most of the files, **even though I would like to enable all of them**.
- IndentPPDirectives: Was used this way in some files, and not in others -> **TBD**
- FixNamespaceComments: Enabling this is always a good idea to keep things in sync.

Please add your comments, whether consecutive assignments and declarations should also be aligned and how to set the indentation of preprocessor directives (nested #ifdef).
```cpp
# AlignConsecutiveDeclarations off
int foo = 5;
double foobar = 7;

# AlignConsecutiveDeclarations on
int    foo    = 5;
double foobar = 7;
```

All files were formatted again with clang-format 13. Most of the changes do not result from the changes made to the configuration, but just apply the existing rules.

**Notes on compatibility**:  
AlignConsecutiveMacros was introduced in clang-format 9 (but I really like it), IndentPPDirectives in version 6. Debian stable (bullseye) comes with version 11, old stable (buster) might be an issue using clang 7 by default. Ubuntu is fine from version 20.04 (focal) and newer, but 18.04 (bionic) seems to have clang-format-9 as well, just as a separate package.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -135,11 +135,13 @@ int Endpoint::handle_read()
     int target_sysid, target_compid, r;
     uint8_t src_sysid, src_compid;
     uint32_t msg_id;
-    struct buffer buf{};
+    struct buffer buf {
+    };
 
-    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id)) > 0)
-        Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid,
-                                           src_compid, msg_id);
+    while ((r = read_msg(&buf, &target_sysid, &target_compid, &src_sysid, &src_compid, &msg_id))
+           > 0)
+        Mainloop::get_instance().route_msg(&buf, target_sysid, target_compid, src_sysid, src_compid,
+                                           msg_id);
 
     return r;
 }
@@ -387,9 +389,9 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     if (has_sys_comp_id(src_sysid, src_compid))
         return false;
 
-    if (msg_id != UINT32_MAX &&
-        _message_filter.size() > 0 &&
-        std::find(_message_filter.begin(), _message_filter.end(), msg_id) == _message_filter.end()) {
+    if (msg_id != UINT32_MAX && _message_filter.size() > 0
+        && std::find(_message_filter.begin(), _message_filter.end(), msg_id)
+            == _message_filter.end()) {
 
         // if filter is defined and message is not in the set then discard it
         return false;
--- a/src/logendpoint.cpp
+++ b/src/logendpoint.cpp
@@ -56,7 +56,7 @@ LogEndpoint::LogEndpoint(const char *name, const char *logs_dir, LogMode mode,
     _fsync_cb.aio_fildes = -1;
 
 #if HAVE_DECL_AIO_INIT
-    aioinit aio_init_data {};
+    aioinit aio_init_data{};
     aio_init_data.aio_threads = 1;
     aio_init_data.aio_num = 1;
     aio_init_data.aio_idle_time = 3; // make sure to keep the thread running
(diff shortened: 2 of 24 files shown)


src/endpoint.cpp
```diff
@@ -387,9 +389,9 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     if (has_sys_comp_id(src_sysid, src_compid))
         return false;
 
-    if (msg_id != UINT32_MAX &&
-        _message_filter.size() > 0 &&
-        std::find(_message_filter.begin(), _message_filter.end(), msg_id) == _message_filter.end()) {
```
lucasdemarchi, line 392: Is there anyway to stop clang-format to change those (other than with a comment on top, that doesn't scale)? Having conditions in single lines sometimes make it easier to read.... Not always, but in this case I very much prefer the way it was, even if it becomes a long line.

src/logendpoint.cpp
```diff
@@ -327,8 +327,9 @@ void LogEndpoint::stop()
 
     // change file permissions to read-only to mark them as finished
     char log_file[PATH_MAX];
-    if (snprintf(log_file, sizeof(log_file), "%s/%s", _logs_dir, _filename) < (int)sizeof(log_file)) {
-        chmod(log_file, S_IRUSR|S_IRGRP|S_IROTH);
+    if (snprintf(log_file, sizeof(log_file), "%s/%s", _logs_dir, _filename)
+        < (int)sizeof(log_file)) {
```
lucasdemarchi, line 331: ditto.... this `< ...` in a single line look hard to read

src/main.cpp
```diff
@@ -57,47 +57,47 @@ static struct options opt = {
...
-    { "conf-file",              required_argument,  NULL,   'c' },
-    { "conf-dir" ,              required_argument,  NULL,   'd' },
-    { "report_msg_statistics",  no_argument,        NULL,   'r' },
-    { "tcp-port",               required_argument,  NULL,   't' },
-    { "tcp-endpoint",           required_argument,  NULL,   'p' },
-    { "log",                    required_argument,  NULL,   'l' },
-    { "debug-log-level",        required_argument,  NULL,   'g' },
-    { "verbose",                no_argument,        NULL,   'v' },
-    { "version",                no_argument,        NULL,   'V' },
-    { }
```
lucasdemarchi, line 71: this one probably deserves a comment on top to stop clang-format to make it horrible.

src/main.cpp
```diff
@@ -57,47 +57,47 @@ static struct options opt = {
...
+                                             {"verbose", no_argument, NULL, 'v'},
+                                             {"version", no_argument, NULL, 'V'},
+                                             {}};
+
+static const char *short_options = "he:rt:c:d:l:p:g:vV";
+
+static void help(FILE *fp)
+{
+    fprintf(
+        fp,
```
lucasdemarchi, line 77: don't really understand the logic of moving fp to the next line... what config is causing this?

src/endpoint.cpp
jbeyerstedt: I'll take a look at it again, but I guess, that it won't be possible. You can always enclose the lines with `// clang-format off` and `// clang-format on`, but that wasn't the question.

src/main.cpp
```diff
@@ -715,9 +712,10 @@ static int parse_confs(ConfFile &conf)
         bool flowcontrol;
     };
     static const ConfFile::OptionsTable option_table_uart[] = {
-        {"baud",        false,  ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_uart, bauds)},
-        {"device",      true,   ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_uart, device)},
-        {"FlowControl", false,  ConfFile::parse_bool,       OPTIONS_TABLE_STRUCT_FIELD(option_uart, flowcontrol)},
```
lucasdemarchi, line 720: another table, another place clang-format does a bad job

lucasdemarchi (commented):
Overall looks good to me. I left some comments in places that I think clang-format is trying to change too much. Thanks.

src/endpoint.cpp
lucasdemarchi: for some places, like the tables, the comment make sense. They will always be tables. For places in the middle of the code, those comments do more harm than good: we either a) find a config we like or b) accept the fact it's better to be consistent than to have everything perfect or c) do not enforce it in some places

PR #298, opened 2021-11-15
Dockerfile: add example dockerfile with alpine

khancyr:
This add a minimal dockerfile as example

This is really handy to use on something like [BalenaOS](https://www.balena.io/) that is an OS with a docker orchestrator.


khancyr:
on side note, on the CI you add `libaio-dev` but it doesn't seems needed. Adding it doesn't change `Header <aio.h> has symbol "aio_init"` to YES

lucasdemarchi:
I had removed that... probably a leftover from the multiple edits.

I don't really like this Dockerfile in the root of the project. Probably because I don't like applications packaged as containers like this. Probably better to move it to examples/  since it's expected that people will not it as is, but modify it for their own use.

khancyr:
moved to example

PR #302, opened 2021-11-26
Docs: Restructure readme and be more precise on UDP modes

jbeyerstedt:
I reworked the "Running" chapter to (hopefully) reduce confusion about:
- master links (which don't exist config- and functionality-wise)
- UDP server and client mode (which is the actual difference between what was previously defined as master link)
- the mapping between CLI options and config file endpoint modes

This PR also changes the chapters from "Running", "Conf file" and "Flight Stack Logging" to "Running" and "Details" in preparation for a more extensive description of mavlink-router's behavior.

It might be controversial, but I also re-formatted the text lines to be one sentence per line instead of hard-wrapped blocks. This will be worse to read in a text editor without soft-wrapping, but I will improve git diffs, because a small change in one sentence will not change all other lines in the paragraph.


lucasdemarchi:
> It might be controversial, but I also re-formatted the text lines to be one sentence per line instead of hard-wrapped blocks. This will be worse to read in a text editor without soft-wrapping, but I will improve git diffs, because a small change in one sentence will not change all other lines in the paragraph.

I think it actually makes it worse, because then we have to rely on the word-diff to figure out when a few words change.

README.md
```diff
@@ -77,102 +73,103 @@ Or in order to install to another root directory:
...
-An example conf file can be found in [examples/config.sample](examples/config.sample)
+Please see the output of `mavlink-routerd --help` for the full list of command line options.
+The most important facts are:
 
-#### Conf dirs ####
+- The TCP server is enabled by default
+- TCP and UDP endpoints can be added multiple times
+- UDP endpoints added with the `-e` option are started in `normal` mode (sending data to the specified address and port)
+- The last parameter (without a key) can either be one UART device or a UDP connection.
+This UDP endpoint will be started in `server` mode (waiting for an incoming connection)!
```
lucasdemarchi, line 104: we probably can say this is just for compatibility with previous behavior. Because without a concept of "master", the additional arg is not really needed

README.md
```diff
@@ -77,102 +73,103 @@ Or in order to install to another root directory:
...
+    $ mavlink-routerd -e 192.168.7.1:14550 -e 127.0.0.1:14550  0.0.0.0:24550
 
-That would change `Endpoint bravo` baudrate to `115200`.
+Additionally, mavlink-router also listens on port 5760 for TCP connections by default.
+Any client connecting to that port will be able to send and receive MAVLink data.
 
-### Flight stack logging ###
+IPv6 addresses must be enclosed in square brackets like this: `[::1]`.
+The port number can be specified in the same way, as with IPv4 then: `[::1]:14550`.
+Both unicast and multicast addresses should be handles properly and the interface for a link-local address is auto-detected.
```
lucasdemarchi, line 123: typo in "handles"?

README.md
```diff
@@ -77,102 +73,103 @@ Or in order to install to another root directory:
...
+### Endpoint Configuration
+
+A description of the different endpoint operating modes and their configuration parameters can be found in the [examples/config.sample](examples/config.sample)
+
+### Flight Stack Logging
+
+Mavlink router can also collect flight stack logs.
+It supports collecting both PX4 and Ardupilot flight stacks logs.
+To start logging, set a directory to the `Log` key in the `General` section in the config file (or use argument option `-l`).
+Currently, mavlink router needs to be informed which MAVLink dialect the flight stack speaks, `common` or `ardupilotmega`.
```
lucasdemarchi, line 139: this is not longer true... there is an "auto" option.

lucasdemarchi (changes requested):
Overall LGTM. I left some comments and would prefer to keep the line wrap as it makes it easier to read and the git diff is simpler when a few words change. Thanks.

README.md
jbeyerstedt: But it's still a different behavior for the UDP endpoints (server vs normal mode), right?

README.md
jbeyerstedt: Sentence reworked

README.md
jbeyerstedt: yes, fixed now.

README.md
lucasdemarchi: indeed... maybe that's why it was left there.

lucasdemarchi (approved):

lucasdemarchi:
Thanks, applied.

PR #303, opened 2021-11-30
Update/ patch python scripts for python 3, fix routing test script

jbeyerstedt:
I had to clean up some code first before fixing things.

The routing test script was failing because the the sender was targeting component 1, but the response was sent by component 0. It would have worked the other way around, though (because of the component broadcast rule). But this also didn't work before, because the source component ID was not set in the `mavutil.mavlink_connection()`.

For easier modification and less magic constants, I introduced a variable to set the number of sent messages as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,12 +23,12 @@ jobs:
         if: ${{ matrix.arch == 'native' }}
         run: |
           sudo apt-get install --yes ninja-build python3-pip
-          pip3 install meson
+          pip3 install meson pymavlink
       - name: install dependencies (cross)
         if: ${{ matrix.arch != 'native' }}
         run: |
           sudo apt-get install --yes ninja-build gcc-${{ matrix.arch }} g++-${{ matrix.arch }} podman python3-pip
-          pip3 install meson
+          pip3 install meson pymavlink
 
       - name: prepare sysroot
         if: ${{ matrix.arch != 'native' }}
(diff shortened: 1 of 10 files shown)


lucasdemarchi:
Nice!! Thanks for fixing that. Maybe we can add it to run automatically on CI too?

jbeyerstedt:
Just added the script as another step in the build tasks. (After 3 attempts of getting the typos and dependencies right.)

The old script used the mavlink-router binary from the current directory, so I changed it from `./mavlink-routerd` to `./build/src/mavlink-routerd`. This works fine in the standard setup (as noted in the readme), but fails in the CI, because it builds inside `./build-${{ matrix.name }}`.

We could either change the test script to use the installed binary (which is not so useful while developing) or introduce a CLI parameter to routing_test.py to provide a custom path to the binary. Or do you have another solution in mind?

jbeyerstedt:
Added the new parameter for now. Seems to be the easiest solution.

jbeyerstedt:
Running the cross compiled builds inside the CI doesn't work somehow (`emu-aarch64-static: Could not open '/lib/ld-linux-aarch64.so.1': No such file or directory`), but that's another issue. The script run is limited to native now.

lucasdemarchi:
Running native is fine. Thanks

lucasdemarchi:
For running the cross-compiled one in CI you'd need to run it inside the cross docker environment. Because although the static qemu is emulating the target distro, the libraries it links to are not copied and available on the host

PR #304, opened 2021-11-30
Add documentation of the message ID filter, rename for future compatibility

jbeyerstedt:
The `Filter` parameter was named quite badly, because it doesn't say what is actually being filtered. In preparation for some issues and outstanding PRs requested more filtering opportunities, this PR renames the filter parameter to "outgoing message ID allow list".
Yes, this is a breaking change of the configuration, but since it wasn't documented and fully implemented, I think this is fine and improves 

This PR also contains:
- Extending the message ID filter to all endpoints
- fix the "messages to unknown endpoints" counter being increased by filtered messages


jbeyerstedt:
Just rebased on master again.

src/mainloop.cpp
```diff
@@ -154,24 +154,48 @@ void Mainloop::route_msg(struct buffer *buf, int target_sysid, int target_compid
     bool unknown = true;
 
     for (Endpoint **e = g_endpoints; *e != nullptr; e++) {
-        if ((*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid, msg_id)) {
+        switch (
+            (*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid, msg_id)) {
```
lucasdemarchi, line 158:
```
AcceptState state = (*e)->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid, msg_id);
switch (state) {
```

would probably be a little more readable and avoid funny autoformat

src/mainloop.cpp
```diff
@@ -154,24 +154,48 @@ void Mainloop::route_msg(struct buffer *buf, int target_sysid, int target_compid
...
+            // fall through
+        default:
+            break; // do nothing (will count as unknown)
         }
     }
 
     for (struct endpoint_entry *e = g_tcp_endpoints; e; e = e->next) {
-        if (e->endpoint->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid,
-                                    msg_id)) {
+        switch (e->endpoint->accept_msg(target_sysid, target_compid, sender_sysid, sender_compid,
```
lucasdemarchi, line 178: ditto.  I think I had a patch to avoid this double loop and code duplication, but it's fine for now

lucasdemarchi (changes requested):

src/main.cpp
```diff
@@ -723,13 +723,14 @@ static int parse_confs(ConfFile &conf)
...
-        char *filter;
+        char *msgIdFilter;
     };
     static const ConfFile::OptionsTable option_table_udp[] = {
         {"address", true, ConfFile::parse_str_dup, OPTIONS_TABLE_STRUCT_FIELD(option_udp, addr)},
         {"mode", true, parse_mode, OPTIONS_TABLE_STRUCT_FIELD(option_udp, server)},
         {"port", false, ConfFile::parse_ul, OPTIONS_TABLE_STRUCT_FIELD(option_udp, port)},
-        {"filter", false, ConfFile::parse_str_dup, OPTIONS_TABLE_STRUCT_FIELD(option_udp, filter)},
+        {"AllowMsgIdOut", false, ConfFile::parse_str_dup,
+         OPTIONS_TABLE_STRUCT_FIELD(option_udp, msgIdFilter)},
```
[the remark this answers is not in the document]
jbeyerstedt: Having both lines write to the msgIdFilter, the newer option will overwrite the old one, if both are present in the config file. Seems to be a good solution since unknown config keys are ignored. So a config with both `Filter` and `AllowMsgIdOut` will work in any version.

src/endpoint.cpp
```diff
@@ -106,6 +106,11 @@ static unsigned int ipv6_get_scope_id(const char *ip)
     return scope;
 }
 
+template <typename type> static bool vector_contains(std::vector<type> vect, type elem)
+{
+    return std::find(vect.begin(), vect.end(), elem) != vect.end();
+}
+
```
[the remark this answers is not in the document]
jbeyerstedt: You're right. Util.c needs to be C++ then, but since it was the only C source file in the whole project, this shouldn't be an issue.

src/mainloop.cpp
jbeyerstedt: Still results in a somewhat funny autoformat, but might be easier during debugging when the variable has a name.

src/endpoint.cpp
```diff
@@ -389,11 +394,9 @@ bool Endpoint::accept_msg(int target_sysid, int target_compid, uint8_t src_sysid
     if (has_sys_comp_id(src_sysid, src_compid))
         return false;
 
+    // If filter is defined and message is not in the set: discard it
     if (msg_id != UINT32_MAX && _allowed_msg_ids.size() > 0
-        && std::find(_allowed_msg_ids.begin(), _allowed_msg_ids.end(), msg_id)
-            == _allowed_msg_ids.end()) {
-
-        // if filter is defined and message is not in the set then discard it
+        && !vector_contains<uint32_t>(_allowed_msg_ids, msg_id)) {
```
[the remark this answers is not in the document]
jbeyerstedt: I wasn't aware of that, but it works with no specific template parameter value

src/mainloop.cpp
jbeyerstedt: I will take a look at the config flow through the code as my next task (therefore all the cleanup before doing the actual changes). So the double loop might get resolved during that change as well.

jbeyerstedt:
Reworked the commits and force-pushed to incorporate the requested changes.

lucasdemarchi (approved):

Issue #284, opened 2021-10-08
Specification of MAVLink Router's Intended Behavior


jbeyerstedt:
As seen in #283 and some other questions risen by users, not everybody is on the same page how MAVLink Router should actually behave. So I quickly started to write down how I imagine the functionality goal: https://gist.github.com/jbeyerstedt/c6df961612d4f544eb307fa168b68062

Actually there are just minor (but important) differences to the current behavior at the UDP endpoint and how the message filters should move forward. Everything else should match how it's currently implemented.

Open questions are:
- What should the `eavesdropping` mode be, actually? Just dumping out all messages going through the router (no filtering) while ignoring incoming messages on that endpoint? Or basically the same as the normal UDP endpoint but just having all the message filters disabled (allowing receiving messages)?
- Which message filters are needed/ should be implemented? Currently there's just an outgoing allow-list for message IDs. I needed an outgoing allow-list for component IDs (see #238) once. So:
  - Are outgoing filters enough?
  - Are allow-lists enough or do we need block-lists, too?
  - By which fields should we filter?

lucasdemarchi:
@patrickelectric  humn... UDP server mode is what pymavlink calls "udpin". Do you know if mavproxy/mavutil supports multiple clients? I don't think so, as when using mavutil we always have to wait for a heartbeat to read the src port. And I think pymavlink "locks" onto it forever.

In theory it could be possible to learn the src ports and send to all of them when sending. Problem here is that unlike tcp, we will have no idea when/if a client disconnected. We could auto-kill endpoints that were dynamically created after X seconds of inactivity. That might work

jbeyerstedt:
Regarding multiple clients on one UDP "server" endpoint: As noted by Lucas, detecting when a client has disconnected isn't supported by UDP. And the implementation of a server-mode UdpEndpoint would need to be completely different from the client-mode UdpEndpoint -- not creating an endpoint instance right away but adding a new one for each client (like TcpServerPort).
So I would propose to keep the UdpEndpoint's functionality as `udpin`/ `udpout` and add the UDP server functionality the same way as `TcpServerPort`.

jbeyerstedt:
Original topic is closed by #327. Please move your wish for a UDP server to a new issue, it it's still up to date.

Issue #110, opened 2017-08-04
Mavlink router does not fall back to broadcast after the first connection times out


LorenzMeier:
I'll try to get better docs to http://mavlink.io, there is not enough or clear documentation on the discovery scheme.

If you connect with a GCS, disconnect, then reconnect with a different GCS mavlink router does not fall back to broadcast (at least according to our testing). It would be great to add this, as its consistent with our discovery scheme.

It should also stop sending heartbeats to broadcast once it is connected to a host to (I have not verified if that is the case right now or not, this is merely a general remark).

lucasdemarchi:
> If you connect with a GCS, disconnect, then reconnect with a different GCS mavlink router does not fall back to broadcast (at least according to our testing).

Yep. I'll add the label bug here. I always intended to implement something like that but lacked information about it.

What's the timeout? How to handle the case in which the same GCS should be used? I.e. the drone simply went too far for a long time and returned. The client side (gcs or others) may not be listening on broadcast anymore).

Or if one want to try a PR should be straightforward.

jbeyerstedt:
This issue is very old, so mavlink router might have changed in the meantime. Because I don't really understand what the issue is. (Especially, if "broadcast" means IP broadcast or mavlink message broadcast.)

An UDP client mode endpoint will just send messages to a port and does not know (and care), if anybody is listening. So this works perfectly fine sending data to the default UDP port of QGC. As soon as the GCS has sent a message back to mavlink-router on that socket (like parameter requests), it'll get these targeted answers, as well (and not just mavlink broadcast messages).

The same should be true for UDP connections from the GCS to mavlink-router (server mode UDP endpoint). Obviously mavlink-router is only able to send data through the socket as soon as a client has connected (sent data) once. So the GCS has to start with it's heartbeat.  
A TCP connection to the `TcpServerPort` works even better.

What am I missing?

lucasdemarchi:
@jbeyerstedt  the problem is that once we lock in to a specific ip:port (due to have received a packet from it), we never release it falling back to "sending packets to broadcast address, waiting for someone to connect". And by "connect" here it is basically "did we receive a message in endpoint in the last X seconds?".  This also assumes all endpoints send heartbeat.

lucasdemarchi:
Documentation added since this issue was created: http://mavlink.io/en/services/heartbeat.html

jbeyerstedt:
[...]

PR #305, opened 2021-12-01
Assorted fixes

jbeyerstedt:
As discussed in #284, this project needs a specification of the routing behavior. This PR adds a description of the current behavior to the readme as well as some more unit tests of the endpoints (no full routing unit test though).

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -87,13 +87,13 @@ static bool is_ipv6(const char *ip)
 
 static bool ipv6_is_linklocal(const char *ip)
 {
-    /* link-local addresses start with fe80 */
-    return strncmp(ip, "fe80", 4) == 0;
+    /* link-local addresses start with fe80, ULA addresses are in fc::/7 range */
+    return (strncmp(ip, "fe80", 4) == 0 || strncmp(ip, "fc", 2) == 0 || strncmp(ip, "fd", 2) == 0);
 }
 
 static bool ipv6_is_multicast(const char *ip)
 {
-    /* link-local addresses start with ff0x */
+    /* multicast addresses start with ff0x (most of the time) */
     return strncmp(ip, "ff0", 3) == 0;
 }
 
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -941,6 +941,11 @@ int UdpEndpoint::open_ipv6(const char *ip, unsigned long port, UdpEndpointConfig
     char *ip_str = strdup(&ip[1]);
     ip_str[strlen(ip_str) - 1] = '\0';
 
+    /* remove omittable zeros from IPv6 address */
+    sockaddr_in6 ip_addr;
+    inet_pton(AF_INET6, ip_str, &ip_addr.sin6_addr);
+    inet_ntop(AF_INET6, &(ip_addr.sin6_addr), ip_str, strlen(ip));
+
     sockaddr6.sin6_family = AF_INET6;
     sockaddr6.sin6_port = htons(port);
 
@@ -951,7 +956,9 @@ int UdpEndpoint::open_ipv6(const char *ip, unsigned long port, UdpEndpointConfig
         struct ipv6_mreq group;
         inet_pton(AF_INET6, ip_str, &group.ipv6mr_multiaddr);
         if (setsockopt(fd, IPPROTO_IPV6, IPV6_ADD_MEMBERSHIP, &group, sizeof(group)) < 0) {
-            log_error("Error setting IPv6 multicast socket options for %s:%lu (%m)", ip, port);
+            log_error("Error setting IPv6 multicast socket options for [%s]:%lu (%m)",
+                      ip_str,
+                      port);
             goto fail;
         }
     } else {
@@ -965,7 +972,7 @@ int UdpEndpoint::open_ipv6(const char *ip, unsigned long port, UdpEndpointConfig
 
     if (mode == UdpEndpointConfig::Mode::Server) {
         if (bind(fd, (struct sockaddr *)&sockaddr6, sizeof(sockaddr6)) < 0) {
-            log_error("Error binding IPv6 socket for %s:%lu (%m)", ip, port);
+            log_error("Error binding IPv6 socket for [%s]:%lu (%m)", ip_str, port);
             goto fail;
         }
         sockaddr6.sin6_port = 0;
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -1221,15 +1221,15 @@ bool UdpEndpoint::validate_config(const UdpEndpointConfig &config)
         return false;
     }
 
-    if (config.mode == UdpEndpointConfig::Mode::Server && config.port == ULONG_MAX) {
-        log_error("UdpEndpoint %s: Port is required in Server mode", config.name.c_str());
-        return false;
-    }
-
     if (config.port == 0 || config.port == ULONG_MAX) {
         log_error("UdpEndpoint %s: Invalid or unset UDP port %lu",
                   config.name.c_str(),
                   config.port);
+        return false;
+    }
+
+    if (config.mode != UdpEndpointConfig::Mode::Client
+        && config.mode != UdpEndpointConfig::Mode::Server) {
         return false;
     }
 
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -79,7 +79,7 @@ const ConfFile::OptionsTable TcpEndpoint::option_table[] = {
 };
 // clang-format on
 
-static bool is_ipv6(const char *ip)
+static bool ip_str_is_ipv6(const char *ip)
 {
     /* Square brackets always exist on IPv6 addresses b/c of input validation */
     return strchr(ip, '[') != nullptr;
@@ -1024,10 +1024,10 @@ bool UdpEndpoint::open(const char *ip, unsigned long port, UdpEndpointConfig::Mo
 {
     const int broadcast_val = 1;
 
-    this->ipv6 = is_ipv6(ip);
+    this->is_ipv6 = ip_str_is_ipv6(ip);
 
-    // setup the special ipv6/ipv4 part
-    if (this->ipv6) {
+    // setup the special IPv6/IPv4 part
+    if (this->is_ipv6) {
         open_ipv6(ip, port, mode);
     } else {
         open_ipv4(ip, port, mode);
@@ -1077,7 +1077,7 @@ bool UdpEndpoint::_nomessage_timeout_cb(void *data)
     Mainloop::get_instance().mod_timeout(nomessage_timeout, 0);
     bool change = false;
 
-    if (this->ipv6) {
+    if (this->is_ipv6) {
         change = memcmp(&sockaddr6, &config_sock.v6, sizeof(sockaddr6)) != 0;
         sockaddr6 = config_sock.v6;
     } else {
@@ -1098,7 +1098,7 @@ ssize_t UdpEndpoint::_read_msg(uint8_t *buf, size_t len)
     struct sockaddr *sock;
     ssize_t r = 0;
 
-    if (this->ipv6) {
+    if (this->is_ipv6) {
         addrlen = sizeof(sockaddr6);
         sock = (struct sockaddr *)&sockaddr6;
     } else {
@@ -1138,7 +1138,7 @@ int UdpEndpoint::write_msg(const struct buffer *pbuf)
     }
 
     bool sock_connected = false;
-    if (this->ipv6) {
+    if (this->is_ipv6) {
         addrlen = sizeof(sockaddr6);
         sock = (struct sockaddr *)&sockaddr6;
         sock_connected = sockaddr6.sin6_port != 0;
@@ -1256,7 +1256,7 @@ bool TcpEndpoint::setup(TcpEndpointConfig conf)
 
     this->_ip = conf.address;
     this->_port = conf.port;
-    this->retry_timeout = conf.retry_timeout;
+    this->_retry_timeout = conf.retry_timeout;
 
     for (auto msg_id : conf.allow_msg_id_out) {
         this->filter_add_allowed_msg_id(msg_id);
@@ -1266,8 +1266,8 @@ bool TcpEndpoint::setup(TcpEndpointConfig conf)
         log_warning("Could not open %s:%ld, re-trying every %d sec",
                     conf.address.c_str(),
                     conf.port,
-                    retry_timeout);
-        if (this->retry_timeout > 0) {
+                    this->_retry_timeout);
+        if (this->_retry_timeout > 0) {
             _schedule_reconnect();
         }
         return true;
@@ -1287,7 +1287,10 @@ int TcpEndpoint::accept(int listener_fd)
     struct sockaddr *sock;
     socklen_t addrlen;
 
-    if (this->ipv6) {
+    this->_retry_timeout = 0; // disable reconnect for incoming TCP server connections
+    this->is_ipv6 = false;    // TCP server is IPv4 only for now
+
+    if (this->is_ipv6) {
         addrlen = sizeof(sockaddr6);
         sock = (struct sockaddr *)&sockaddr6;
     } else {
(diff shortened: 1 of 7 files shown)


jbeyerstedt:
Note to myself:
- Add an explicit sentence, that the config file is not required
- Document current behavior when using broadcast or multicast IP addresses (see #110 )

README.md
```diff
@@ -1,17 +1,10 @@
...
-mavlink-router mainly routes mavlink packets from one endpoint to the other
-endpoints without differentiating what they are. However the user is still able
-to differentiate them by targeting just a subset of the packets by using
-message filters.
-
-TCP endpoints are added automatically if the TCP server is enabled, allowing
-clients to simply connect to mavlink-router without changing its configuration.
+MAVLink Router is an application to distribute [MAVLink](https://mavlink.io/en/)
+messages between multiple endpoints (connections). It actually behaves more
+like a network switch distributing packets to a single port or multiple ports
```
lucasdemarchi, line 5: I'd say it's more like a network router than a network switch/bridge.  Hence the name :)

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
-
-### Endpoint Configuration
-
-A description of the different endpoint operating modes and their configuration
-parameters can be found in the [examples/config.sample](examples/config.sample).
+To understand how MAVLink Router forwards messages between endpoints, it
+important to know, that a MAVLink message has to have a sender address, but
+only some messages have a target address (e.g. parameter requests). The sender
+and target address consist of a system and component ID each, so one UAV can
+have a flight controller and companion computer with individual component IDs,
```
lucasdemarchi, line 145: the companion computer itself usually does not have a mavlink sys/comp id. Individual programs running on it do

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
+but sharing the same system ID. In the target address, a component ID or 0 is
```
lucasdemarchi, line 146: I guess you meant `a component ID of 0`?

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
+MAVLink Router supports three basic types of endpoints: UART, UDP link and TCP
+client. Additionally, it'll act as a TCP server for dynamic clients (if not
+explicitly deactivated).
+
+Endpoint types (see [examples/config.sample](examples/config.sample) for the
+config file format):
+
+  - UART: For telemetry radios or other serial links
+    * Configuration: UART device path/name and baudrate
+    * Behavior: Data is received and sent right away
```
lucasdemarchi, line 160: we still "packetize" what we receive, buffering it until we have a complete mavlink message.

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
+  - UDP:
+    * Configuration: Mode (client or server), IP address and port; both unicast
+      and multicast addresses are possible
+    * Behavior in client mode: Data is sent right away, but can only be
+      received on established socket connection
+    * Behavior in server mode: Data is received right away, but can only be
```
lucasdemarchi, line 166: I think the distinction is more on "is the port configure for sending or receiving?". Then when data can be received. Or we may want to be clearer on what "[sent|received| right away" means. Here it means "on program startup".

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
+config file format):
+
+  - UART: For telemetry radios or other serial links
+    * Configuration: UART device path/name and baudrate
+    * Behavior: Data is received and sent right away
+  - UDP:
+    * Configuration: Mode (client or server), IP address and port; both unicast
+      and multicast addresses are possible
+    * Behavior in client mode: Data is sent right away, but can only be
+      received on established socket connection
```
lucasdemarchi, line 165: what does "established socket connection" mean? There is no connection...

README.md
```diff
@@ -145,12 +138,80 @@ for a link-local address is auto-detected.
...
+      during runtime.)
+
+### Message Routing
+In general, each message received on one endpoint is delivered to all other
+endpoints. But there are some rules to reduce the number of "uninteresting"
+messages being delivered to a system.
+
+Routing rules:
+
+  - A message received on one endpoint is delivered to all other endpoints.
```
lucasdemarchi, line 193: this is only true for broadcast message. A better wording could be: "A message received on one endpoint is delivered to all endpoints in which that target system/component has been seen. If it's a broadcast message, it's delivered to all endpoints. A message is never sent back to the same endpoint it came from".

README.md
jbeyerstedt: Rephrased all of the "sent right away" parts.

README.md
jbeyerstedt: I thought of these rules as a generic rule with multiple exceptions. But since that doesn't seem to be clear everybody, I changed the order and phrasing of the rules.

README.md
```diff
@@ -145,12 +142,93 @@ for a link-local address is auto-detected.
...
+    * Behavior: Data is received and sent without waiting for incoming data first
+  - UDP:
+    * Configuration: Mode (client or server), IP address and port; both unicast
+      and multicast addresses are possible
+    * Behavior in client mode: Endpoint is configured with a target IP and port
+      combination. So MAVLink messages can be sent directly after startup, but
+      will only be recevied after the first message was received by the remote
+      side, it doesn't know our IP and port otherwise.  
+      When using a broadcast or multicast IP address, messages will always be
+      sent to that broadcast/multicast IP.
```
jbeyerstedt, line 173: @lucasdemarchi Is that how it should work with UDP client and a broadcast IP address? Or do we really need to start with sending IP broadcasts, switch to a unicast connection after the first incoming packet and switch back to broadcast after a timeout.

lucasdemarchi: current behavior when using a broadcast address is that it will send to that address until it receives a reply. When we receive a reply from, say, 192.168.0.10:54243  we "lock in this address:port" and all new packets are sent there.
We never switch back to broadcast, but it would be good if we did: otherwise if the connection is broken we never recover.

jbeyerstedt: The current behavior is clear, but it's quite clearly wrong (issue exists) and therefore I would like to describe the way it should be in this PR.
The question is, if it should be fixed to broadcast forever or whether we really need that timeout.

lucasdemarchi: Broadcast forever would be wrong. We need a timeout.

lucasdemarchi: All systems should be sending a hearbeat, so we should receive at least 1 message every second or so. A 5s timeout would be reasonable.

jbeyerstedt: OK. But I would propose a third mode for that in this case. Because it behaves quite differently than all the other modes. What do you think?

lucasdemarchi: No, I think not going back to broadcast was a missing feature. It's only triggered if the address supplied by the user is a broadcast address.

jbeyerstedt: So no new mode for the UDP endpoint but rather some special behavior triggered by a broadcast IPv4 address? But how to reliably detect a broadcast address without knowing the subnet size? We can't just assume everyone uses /24 networks.

lucasdemarchi: If we needed, we could query it from the kernel. But just setting the broadcast flag regardless if it's a broadcast address proved to work: it will be ignored by the kernel if it's not a broadcast address.

jbeyerstedt: But how do we determine then (just with the existing settings), if we need to check for a timeout or not? Or do we just always check for a timeout in client-mode regardless if the address is broadcast or not?
When thinking about the latter option: It should be any different than currently for unicast addresses, because the source address of the latest incoming packet should be the same as the configured one. So resetting to the configured address after a timeout has no effect. Is that what you were thinking of?

lucasdemarchi: We need to keep a copy of the address that was configured by the user and switch back to it after a disconnect timeout.
If it was not a broadcast address, nothing will actually change, because recvfrom() won't have changed that.

Supposing the following scenario:
mavlink-router: inet 192.168.2.42/24 brd 192.168.2.255
GCS: inet 192.168.2.1/24 brd 192.168.2.255

Endpoint configuration: 192.168.2.255:14550

So when we start mavlink-router we will start sending the packets to 192.168.2.255 (broadcast). GCS will receive with src_address=192.168.2.42:<something>, so it responds to that IP:port. When that happens, mavlink-router will receive a packet from, recvfrom() in mavlink-router will receive a packet from 192.168.2.1:14550. Currently we are overriding the configured address and losing it - we should rather save it.  All subsequent packets we send through the Endpoint we are then sending to 192.168.2.1:14550.  Proposal here is that when a timeout happens, we handle the "disconnect" by switching back to 192.168.2.255:14550.

jbeyerstedt:
From my side, this could be merged.

src/mainloop.cpp
```diff
@@ -154,13 +154,16 @@ int Mainloop::write_msg(const std::shared_ptr<Endpoint> &e, const struct buffer
...
-void Mainloop::route_msg(struct buffer *buf)
+void Mainloop::route_msg(const Endpoint *msg_receiver, struct buffer *buf)
 {
     bool unknown = true;
 
     for (const auto &e : this->g_endpoints) {
-        auto acceptState = e->accept_msg(buf);
+        if (e.get() == msg_receiver) {
+            continue; // don't offer message to the endpoint where it was received
+        }
```
lucasdemarchi, line 164: what case this covers what is not covered by `has_sys_comp_id()` inside Endpoint::accept_msg() that is called just below?

src/endpoint.cpp
```diff
@@ -87,13 +87,13 @@ static bool is_ipv6(const char *ip)
...
-    /* link-local addresses start with fe80 */
-    return strncmp(ip, "fe80", 4) == 0;
+    /* link-local addresses start with fe80, ULA addresses are in fc::/7 range */
+    return (strncmp(ip, "fe80", 4) == 0 || strncmp(ip, "fc", 2) == 0 || strncmp(ip, "fd", 2) == 0);
 }
 
 static bool ipv6_is_multicast(const char *ip)
 {
-    /* link-local addresses start with ff0x */
+    /* multicast addresses start with ff0x (most of the time) */
```
lucasdemarchi, line 96: commit message should mention ipv6:  "Endpoint: Allow IPv6 unique local addresses as link local"

src/endpoint.cpp
```diff
@@ -1013,10 +1013,10 @@ bool UdpEndpoint::open(const char *ip, unsigned long port, UdpEndpointConfig::Mo
 {
     const int broadcast_val = 1;
 
-    this->ipv6 = is_ipv6(ip);
+    this->is_ipv6 = ip_str_is_ipv6(ip);
 
-    // setup the special ipv6/ipv4 part
-    if (this->ipv6) {
+    // setup the special is_ipv6/ipv4 part
```
lucasdemarchi, line 1018: wrong replacement

lucasdemarchi: maybe this could be updated to IPv6/IPv4

src/endpoint.cpp
```diff
@@ -1323,12 +1323,12 @@ int TcpEndpoint::open_ipv4(const char *ip, unsigned long port, sockaddr_in &sock
 
 bool TcpEndpoint::open(const std::string &ip, unsigned long port)
 {
-    this->ipv6 = is_ipv6(ip.c_str());
+    this->is_ipv6 = ip_str_is_ipv6(ip.c_str());
 
-    // setup the special ipv6/ipv4 part
+    // setup the special is_ipv6/ipv4 part
```
lucasdemarchi, line 1328: ditto

src/mainloop.cpp
```diff
@@ -228,7 +228,7 @@ void Mainloop::handle_tcp_connection()
     log_debug("TCP Server: New client");
 
     auto *tcp = new TcpEndpoint{"dynamic"};
-    tcp->retry_timeout = 0; // disable retry on dynamic endpoints by the TCP server
+    tcp->set_retry_timeout(0); // disable retry on dynamic endpoints by the TCP server
```
lucasdemarchi, line 231: this can only be set be the configuration, which is now handled in the ::setup() method. So maybe we should just remove this line

src/endpoint.h
```diff
@@ -242,30 +238,29 @@ class UdpEndpoint : public Endpoint {
...
-    struct sockaddr_in sockaddr;
-    struct sockaddr_in6 sockaddr6;
-    bool ipv6;
-    int retry_timeout = 0; // disable retry by default
-
-    inline std::string get_ip() { return _ip; }
-    inline unsigned long get_port() const { return _port; }
-    bool is_valid() override { return _valid; };
-    bool is_critical() override { return false; };
+    void set_retry_timeout(int retry_timeout) { _retry_timeout = retry_timeout; };
```
lucasdemarchi, line 263: This method doesn't seem to be needed as it can only be set by the config

src/endpoint.h
```diff
@@ -58,7 +58,7 @@ struct TcpEndpointConfig {
     std::string name;
     std::string address;
     unsigned long port;
-    int retry_timeout;
+    int retry_timeout{5};
```
lucasdemarchi, line 61: Commenting here, but this is about the commit message:

```
Previously some default values were set in the struct definition and others
explicitly in the parsing functions. Not there's just one place: In each config
struct's definition.
```

Typo: "Now there's ..."

src/endpoint.cpp
```diff
@@ -415,7 +415,7 @@ void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
 bool Endpoint::has_sys_id(unsigned sysid) const
 {
     for (const auto &id : _sys_comp_ids) {
-        if (((id >> 8) | (sysid & 0xff)) == sysid) {
+        if (uint16_t(id >> 8) == (sysid & 0x00ff)) {
```
lucasdemarchi, line 418: I think we need this fix as a separate PR, or at the very least a separate commit

src/endpoint.cpp
```diff
@@ -1179,18 +1179,18 @@ bool UdpEndpoint::validate_config(const UdpEndpointConfig &config)
         return false;
     }
 
-    if (config.mode == UdpEndpointConfig::Mode::Server && config.port == ULONG_MAX) {
-        log_error("UdpEndpoint %s: Port is required in Server mode", config.name.c_str());
-        return false;
-    }
-
```
lucasdemarchi, line 1179: why is this in this commit?

lucasdemarchi: Looks like a wrong squash.... should be in the commit that cleanup the defaults?

lucasdemarchi:
Overall it looks a very great improvement. I left some comments on a few issues I found. Generally it's good to give longer explanations in the commit message why something is being done and I miss them here.

Also, in a PR "Add detailed description of the routing behavior to Readme", I was really expecting only documentation changes, not anything else. The first commit with doc changes is ready to be applied, the others need tweaks as commented above.

Thanks!

lucasdemarchi:
I moved the first and part of the last commits to separate PR so they can be applied without having to wait for the others here.

src/mainloop.cpp
jbeyerstedt: It originates from the parallel data links issue and actually should replace the `has_sys_comp_id()` check. But at a second thought, it's more of a workaround now than a real fix to that issue. (I don't think discussing the details in this PR is the right place.)
I don't know, why that isn't in the commit. But then, it'll be a bigger discussion topic and I'll open another PR for that.

src/mainloop.cpp
jbeyerstedt: Since the method will not change the underlying setting (timeout handling) at any place in time, I removed the method all together and incorporated the setting to zero into `accept()`.

src/endpoint.h
jbeyerstedt: See above.

src/endpoint.cpp
jbeyerstedt: Already done by you, thanks.

src/endpoint.cpp
jbeyerstedt: It was something, that came up during testing that method, so I added the fix to the unit test commit. I can move it, if you like.

jbeyerstedt:
Sorry for trying to sneak in some changes with the documentation. I thought, it was a good idea to bundle the unit tests testing the refined specification with the specification itself.

PR #306, opened 2021-12-01
Modernize code and enable unit test in CI again

jbeyerstedt:
Based on #304, but already posted to show what I'm working on. Just ignore the first three commits.

To improve code quality, I let clang-tidy fix some things I consider a bad or error prone coding style. The checks were hand-picked and all changed reviewed manually before committing. For easier review, I created one commit per clang-tidy rule, but I would suggest squashing them before merging (when the other contributors are happy with the changes).

I also realized, that the unit test aren't run on the CI, so I snug a fix into this PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
       - name: install dependencies (native)
         if: ${{ matrix.arch == 'native' }}
         run: |
-          sudo apt-get install --yes ninja-build python3-pip
+          sudo apt-get install --yes ninja-build python3-pip libgtest-dev
           pip3 install meson pymavlink
       - name: install dependencies (cross)
         if: ${{ matrix.arch != 'native' }}
--- a/examples/arm-authorizer.cpp
+++ b/examples/arm-authorizer.cpp
@@ -70,8 +70,8 @@ static void setup_signal_handlers()
 
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
-    sigaction(SIGTERM, &sa, NULL);
-    sigaction(SIGINT, &sa, NULL);
+    sigaction(SIGTERM, &sa, nullptr);
+    sigaction(SIGINT, &sa, nullptr);
 }
 
 static int msg_send(mavlink_message_t *msg)
--- a/examples/arm-authorizer.cpp
+++ b/examples/arm-authorizer.cpp
@@ -31,7 +31,7 @@
 
 #define SYSTEM_ID 10
 
-#define AUTH_TIMEOUT_SEC 10
+#define AUTH_TIMEOUT_SEC      10
 #define HEARTBEAT_TIMEOUT_SEC 3
 
 #define MAX_VALID_ALTITUDE 15.0f
@@ -49,12 +49,14 @@ static timespec heartbeat_timeout;
 
 static uint8_t target_system;
 
-static enum {
+enum state_t {
     STATE_IDLE = 0,
     STATE_WAITING_MISSING_COUNT,
     STATE_WAITING_MISSION_ITEMS,
     STATE_LAST
-} state = STATE_IDLE;
+};
+
+static state_t state = STATE_IDLE;
 static uint16_t mission_item_count;
 static bool mission_valid;
 static uint16_t first_mission_item_invalid;
@@ -66,7 +68,7 @@ static void exit_signal_handler(int signum)
 
 static void setup_signal_handlers()
 {
-    struct sigaction sa = { };
+    struct sigaction sa = {};
 
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
@@ -119,9 +121,8 @@ static void mission_ack_send()
 
 static void handle_command_long(const mavlink_message_t *msg, mavlink_command_long_t *cmd)
 {
-    if (cmd->command != MAV_CMD_ARM_AUTHORIZATION_REQUEST
-            || state != STATE_IDLE
-            || cmd->target_system != SYSTEM_ID) {
+    if (cmd->command != MAV_CMD_ARM_AUTHORIZATION_REQUEST || state != STATE_IDLE
+        || cmd->target_system != SYSTEM_ID) {
         return;
     }
 
@@ -154,8 +155,7 @@ static void handle_command_long(const mavlink_message_t *msg, mavlink_command_lo
 
 static void handle_mission_count(mavlink_mission_count_t *mission_count)
 {
-    if (state != STATE_WAITING_MISSING_COUNT
-            || mission_count->target_system != SYSTEM_ID) {
+    if (state != STATE_WAITING_MISSING_COUNT || mission_count->target_system != SYSTEM_ID) {
         return;
     }
 
@@ -186,8 +186,7 @@ static void handle_mission_count(mavlink_mission_count_t *mission_count)
 
 static void handle_mission_item_int(mavlink_mission_item_int_t *mission_item)
 {
-    if (state != STATE_WAITING_MISSION_ITEMS
-            || mission_item->target_system != SYSTEM_ID) {
+    if (state != STATE_WAITING_MISSION_ITEMS || mission_item->target_system != SYSTEM_ID) {
         return;
     }
 
@@ -195,13 +194,14 @@ static void handle_mission_item_int(mavlink_mission_item_int_t *mission_item)
     x /= 10000000;
     y /= 10000000;
 
-    printf("Mission item { seq=%u mission_type=%u x=%f y=%f altitude=%f }\n",
-           mission_item->seq, mission_item->mission_type, x, y, mission_item->z);
+    printf("Mission item { seq=%u mission_type=%u x=%f y=%f altitude=%f }\n", mission_item->seq,
+           mission_item->mission_type, x, y, mission_item->z);
 
     if (mission_valid && mission_item->z > MAX_VALID_ALTITUDE) {
         mission_valid = false;
         first_mission_item_invalid = mission_item->seq;
-        printf("Arm request will be denied because mission item seq=%u have a invalid altitude, altitude=%f max accepted altitude=%f\n",
+        printf("Arm request will be denied because mission item seq=%u have a invalid altitude, "
+               "altitude=%f max accepted altitude=%f\n",
                mission_item->seq, mission_item->z, MAX_VALID_ALTITUDE);
     }
 
--- a/examples/heartbeat-print.cpp
+++ b/examples/heartbeat-print.cpp
@@ -41,8 +41,8 @@ static void setup_signal_handlers()
 
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
-    sigaction(SIGTERM, &sa, NULL);
-    sigaction(SIGINT, &sa, NULL);
+    sigaction(SIGTERM, &sa, nullptr);
+    sigaction(SIGINT, &sa, nullptr);
 }
 
 static void handle_new_message(const mavlink_message_t *msg)
@@ -104,14 +104,16 @@ int main(int argc, char *argv[])
         uint8_t buf[1024];
         ssize_t n = read(fd, buf, sizeof(buf));
         if (n == -1) {
-            if (errno == EINTR)
+            if (errno == EINTR) {
                 continue;
+            }
             break;
         }
 
         for (int i = 0; i < n; i++) {
-            if (mavlink_parse_char(MAVLINK_COMM_0, buf[i], &msg, &status))
+            if (mavlink_parse_char(MAVLINK_COMM_0, buf[i], &msg, &status)) {
                 handle_new_message(&msg);
+            }
         }
 
     }
--- a/examples/px4-offboard-mode.cpp
+++ b/examples/px4-offboard-mode.cpp
@@ -147,9 +147,9 @@ static int setup_signal_handlers()
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
 
-    if (sigaction(SIGTERM, &sa, NULL) != 0 ||
-        sigaction(SIGINT, &sa, NULL) != 0)
+    if (sigaction(SIGTERM, &sa, nullptr) != 0 || sigaction(SIGINT, &sa, nullptr) != 0) {
         return -1;
+    }
 
     return 0;
 }
@@ -436,7 +436,7 @@ static void timeout_callback()
         set_mode_send(PX4_MODE_OFFBOARD);
         break;
     case ARM_DISARM: {
-        if (armed == t->param1) {
+        if (static_cast<int>(armed) == t->param1) {
             task_list_index++;
             break;
         }
@@ -642,7 +642,7 @@ static int setup_timeout()
     ts.it_value.tv_sec = ts.it_interval.tv_sec;
     ts.it_value.tv_nsec = ts.it_interval.tv_nsec;
 
-    timerfd_settime(timeout_fd, 0, &ts, NULL);
+    timerfd_settime(timeout_fd, 0, &ts, nullptr);
 
     return 0;
 }
@@ -662,10 +662,9 @@ int main(int argc, char *argv[])
         }
     }
 
-    if (setup_connection(ip, port) < 0 ||
-        setup_timeout() < 0 ||
-        setup_signal_handlers() < 0)
+    if (setup_connection(ip, port) < 0 || setup_timeout() < 0 || setup_signal_handlers() < 0) {
         goto fail;
+    }
 
     free(ip);
 
@@ -679,10 +678,12 @@ int main(int argc, char *argv[])
 fail:
     free(ip);
 
-    if (tcp_fd >= 0)
+    if (tcp_fd >= 0) {
         close(tcp_fd);
-    if (timeout_fd >=0)
+    }
+    if (timeout_fd >= 0) {
         close(timeout_fd);
+    }
 
     return EXIT_FAILURE;
 }
--- a/examples/px4-offboard-mode.cpp
+++ b/examples/px4-offboard-mode.cpp
@@ -41,10 +41,10 @@
 #include <poll.h>
 #include <signal.h>
 #include <stdint.h>
-#include <stdlib.h>
 #include <stdio.h>
-#include <sys/timerfd.h>
+#include <stdlib.h>
 #include <sys/socket.h>
+#include <sys/timerfd.h>
 #include <time.h>
 #include <unistd.h>
 
@@ -52,10 +52,10 @@
 
 #include "common/util.h"
 
-#define SYSTEM_ID 11
+#define SYSTEM_ID        11
 #define TARGET_SYSTEM_ID 1
 
-#define PX4_MAIN_MODE_CUSTOM 1
+#define PX4_MAIN_MODE_CUSTOM     1
 #define PX4_CUSTOM_MODE_OFFBOARD 6
 
 #define POLL_ERROR_EVENTS (POLLERR | POLLHUP | POLLNVAL)
@@ -69,16 +69,9 @@ static struct sockaddr_in sockaddr;
 
 static int timeout_fd = -1;
 
-enum px4_modes {
-    PX4_MODE_UNKNOWN = 0,
-    PX4_MODE_OFFBOARD,
-    PX4_MODE_LAST
-};
+enum px4_modes { PX4_MODE_UNKNOWN = 0, PX4_MODE_OFFBOARD, PX4_MODE_LAST };
 
-static const char * mode_names[] = {
-    "unknown",
-    "offboard"
-};
+static const char *mode_names[] = {"unknown", "offboard"};
 
 static enum px4_modes current_mode = PX4_MODE_UNKNOWN;
 static bool armed = false;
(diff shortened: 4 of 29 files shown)


examples/heartbeat-print.cpp
```diff
@@ -17,11 +17,11 @@
  */
 
 #include <arpa/inet.h>
-#include <assert.h>
-#include <errno.h>
-#include <signal.h>
-#include <stdio.h>
-#include <stdlib.h>
+#include <cassert>
```
lucasdemarchi, line 20: then we mix headers with and without .h ?  never understood the benefit of this "modernization"

src/endpoint.cpp
```diff
@@ -360,17 +360,17 @@ void Endpoint::_add_sys_comp_id(uint16_t sys_comp_id)
 
 bool Endpoint::has_sys_id(unsigned sysid)
 {
-    for (auto it = _sys_comp_ids.begin(); it != _sys_comp_ids.end(); it++) {
-        if (((*it >> 8) | (sysid & 0xff)) == sysid)
+    for (auto &id : _sys_comp_ids) {
```
lucasdemarchi, line 365: this looks ok

src/autolog.cpp
```diff
@@ -35,15 +35,13 @@ int AutoLog::write_msg(const struct buffer *buffer)
     uint8_t source_component_id;
 
     if (mavlink2) {
-        struct mavlink_router_mavlink2_header *hdr
-            = (struct mavlink_router_mavlink2_header *)buffer->data;
+        auto *hdr = (struct mavlink_router_mavlink2_header *)buffer->data;
```
lucasdemarchi, line 38: mixed feelings, but lgtm too

src/common/conf_file.h
```diff
@@ -132,8 +130,8 @@ class ConfFile {
 #undef DECLARE_PARSE_INT
 
 private:
-    struct conffile *_files;
-    struct section *_sections;
+    struct conffile *_files{nullptr};
+    struct section *_sections{nullptr};
```
lucasdemarchi, line 135: lgtm

src/common/conf_file.h
```diff
@@ -23,7 +23,8 @@
  */
 class ConfFile {
 public:
-    ConfFile(){};
+    ConfFile() = default;
+    ;
```
lucasdemarchi, line 27: this looks wrong

lucasdemarchi: but the = default is ok

src/mainloop.h
```diff
@@ -62,6 +62,8 @@ class Mainloop {
         assert(_initialized);
         return _instance;
     }
+    Mainloop(const Mainloop &) = delete;
+    Mainloop &operator=(const Mainloop &) = delete;
```
lucasdemarchi, line 66: I don't really understand the choice of moving the constructor to be public. Should rather continue to be private

src/endpoint.h
```diff
@@ -85,7 +85,7 @@ class Endpoint : public Pollable {
     };
 
     Endpoint(const char *name);
-    virtual ~Endpoint();
+    ~Endpoint() override;
```
lucasdemarchi, line 88: ok

examples/arm-authorizer.cpp
```diff
@@ -70,8 +70,8 @@ static void setup_signal_handlers()
 
     sa.sa_flags = SA_NOCLDSTOP;
     sa.sa_handler = exit_signal_handler;
-    sigaction(SIGTERM, &sa, NULL);
-    sigaction(SIGINT, &sa, NULL);
+    sigaction(SIGTERM, &sa, nullptr);
+    sigaction(SIGINT, &sa, nullptr);
```
lucasdemarchi, line 75: ok

src/common/util.h
```diff
@@ -28,8 +28,8 @@
 extern "C" {
 #endif
 
-typedef uint64_t usec_t;
-typedef uint64_t nsec_t;
+using usec_t = uint64_t;
+using nsec_t = uint64_t;
```
lucasdemarchi, line 32: ok

examples/heartbeat-print.cpp
```diff
@@ -63,7 +63,7 @@ int main(int argc, char *argv[])
     struct sockaddr_in sockaddr;
     const socklen_t addrlen = sizeof(sockaddr);
     int port = 14550;
-    int fd;
+    int fd = 0;
```
lucasdemarchi, line 65: this is nonsense and hides bug that would rather be caught by valgrind

src/endpoint.cpp
```diff
@@ -508,7 +507,7 @@ UartEndpoint::~UartEndpoint()
 
 int UartEndpoint::set_speed(speed_t baudrate)
 {
-    struct termios2 tc;
+    struct termios2 tc = {};
```
lucasdemarchi, line 510: we do a bzero just below. Initializing structs like this actually has a problem: if the struct has padding, the padding is not initialized, so when passed to a syscall it is flagged as "passing garbage to a syscall" by valgrind

examples/px4-offboard-mode.cpp
```diff
@@ -661,10 +662,9 @@ int main(int argc, char *argv[])
         }
     }
 
-    if (setup_connection(ip, port) < 0 ||
-        setup_timeout() < 0 ||
-        setup_signal_handlers() < 0)
+    if (setup_connection(ip, port) < 0 || setup_timeout() < 0 || setup_signal_handlers() < 0) {
```
lucasdemarchi, line 662: isn't the previous one easier to read?

src/common/conf_file.cpp
```diff
@@ -119,20 +119,23 @@ int ConfFile::_parse_file(const char *addr, size_t len, const char *filename)
 
         // Make sure all casts to int are safe. No need to support options or sections
         // larger than INT_MAX
-        if (line_len > INT_MAX)
+        if (line_len > INT_MAX) {
             return -EINVAL;
+        }
```
lucasdemarchi, line 124: I find the addition { } verbose, but I can survive with them added. However here for example you removed the extra newline, that IMO makes the statements easier to read.

src/common/conf_file.cpp
```diff
@@ -143,8 +146,9 @@ int ConfFile::_parse_file(const char *addr, size_t len, const char *filename)
             }
 
             ret = _add_config(s, addr, line_len, filename, line);
-            if (ret < 0)
+            if (ret < 0) {
                 return ret;
+            }
```
lucasdemarchi, line 151: ditto

src/endpoint.cpp
```diff
@@ -409,7 +409,7 @@ Endpoint::AcceptState Endpoint::accept_msg(int target_sysid, int target_compid,
     }
 
     // If filter is defined and message is not in the set: discard it
-    if (msg_id != UINT32_MAX && _allowed_msg_ids.size() > 0
+    if (msg_id != UINT32_MAX && !_allowed_msg_ids.empty()
```
lucasdemarchi, line 403: ok

examples/heartbeat-print.cpp
```diff
@@ -79,7 +79,7 @@ int main(int argc, char *argv[])
 
     char *ip = strdup(argv[1]);
     char *portstr = strchrnul(ip, ':');
-    if (portstr && *portstr) {
+    if ((portstr != nullptr) && (*portstr != 0)) {
```
lucasdemarchi, line 82: sorry, I like the implicit conversion more.

src/autolog.h
```diff
@@ -30,7 +30,7 @@ class AutoLog : public LogEndpoint {
     {
     }
 
-    int write_msg(const struct buffer *pbuf) override;
+    int write_msg(const struct buffer *buffer) override;
```
lucasdemarchi, line 33: ok

src/main.cpp
```diff
@@ -869,10 +869,12 @@ static int parse_conf_files()
...
     DIR *dir = nullptr;
     struct dirent *ent = nullptr;
-    const char *filename = nullptr, *dirname = nullptr;
+    const char *filename = nullptr;
+    const char *dirname = nullptr;
     int ret = 0;
     char *files[128] = {};
-    int i = 0, j = 0;
+    int i = 0;
+    int j = 0;
```
lucasdemarchi, line 879: too verbose

src/endpoint.cpp
```diff
@@ -434,7 +434,7 @@ Endpoint::AcceptState Endpoint::accept_msg(int target_sysid, int target_compid,
     return Endpoint::AcceptState::Rejected;
 }
 
-bool Endpoint::_check_crc(const mavlink_msg_entry_t *msg_entry)
+bool Endpoint::_check_crc(const mavlink_msg_entry_t *msg_entry) const
```
lucasdemarchi, line 428: ok

src/logendpoint.cpp
```diff
@@ -432,7 +432,7 @@ bool LogEndpoint::_start_alive_timeout()
 }
 
 void LogEndpoint::_handle_auto_start_stop(uint32_t msg_id, uint8_t source_system_id,
-                                          uint8_t source_component_id, uint8_t *payload)
+                                          uint8_t source_component_id, const uint8_t *payload)
```
lucasdemarchi, line 434: ok

src/endpoint.cpp
```diff
@@ -89,7 +89,7 @@ static unsigned int ipv6_get_scope_id(const char *ip)
                         nullptr, 0, NI_NUMERICHOST);
 
             /* cut the interface name from the end of a link-local address */
-            auto search = strrchr(ipAddress, '%');
+            auto *search = strrchr(ipAddress, '%');
```
lucasdemarchi, line 83: ok

src/endpoint.cpp
```diff
@@ -51,28 +51,19 @@
 static bool is_ipv6(const char *ip)
 {
     /* Square brackets always exist on IPv6 addresses b/c of input validation */
-    if (strchr(ip, '[') != nullptr) {
-        return true;
-    }
-    return false;
+    return strchr(ip, '[') != nullptr;
```
lucasdemarchi, line 54: lgtm

src/endpoint.cpp
```diff
@@ -678,7 +678,7 @@ int UartEndpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_c
 
     if (_change_baud_timeout != nullptr && ret == ReadOk) {
         log_info("Baudrate %lu responded, keeping it", _baudrates[_current_baud_idx]);
-        Mainloop::get_instance().del_timeout(_change_baud_timeout);
```
lucasdemarchi, line 670: seems wrong, nops

examples/arm-authorizer.cpp
```diff
@@ -54,7 +54,8 @@ static enum {
     STATE_WAITING_MISSING_COUNT,
     STATE_WAITING_MISSION_ITEMS,
     STATE_LAST
-} state = STATE_IDLE;
+} state
+    = STATE_IDLE;
```
lucasdemarchi, line 58: this one seems like clang-format was drunk and decided to do whatever

.github/workflows/main.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: install dependencies (native)
         if: ${{ matrix.arch == 'native' }}
         run: |
-          sudo apt-get install --yes ninja-build python3-pip
+          sudo apt-get install --yes ninja-build python3-pip libgtest-dev
```
lucasdemarchi, line 25: lgtm

lucasdemarchi (commented):
Overall most of the changes lgtm. There are some I really disliked and would prefer to keep without them. Thanks

examples/heartbeat-print.cpp
jbeyerstedt: I have researched the differences again and found no real difference, than "it the C++ style". Will change it back later today or tomorrow.

src/common/conf_file.h
jbeyerstedt: Yes. Will fix.

src/mainloop.h
jbeyerstedt: It should give you one compiler error less when trying to use the c-tor. But I'll revert the change as it gives no real benefit.
When taking a look at the changes by Auterion regarding the testability of the mainloop, they changed the ownership so that the singleton isn't really needed any more. Maybe I'll incorporate something similar when trying to write more tests. (And I don't like singletons anyways, because they make testing quite complicated.)

src/common/conf_file.cpp
jbeyerstedt: I'm a really big fan of adding {} at every if statement. But you're right about the extra newline.

src/endpoint.cpp
jbeyerstedt: Because del_timeout doesn't change any member variable of the class, the method should be static. So calling it via `Classname::static_method()` is actually the right way.

jbeyerstedt:
I'll take some time tomorrow to incorporate the changes/ other notes I didn't comment on.

src/common/conf_file.cpp
jbeyerstedt: There was no blank line after the return before.

examples/heartbeat-print.cpp
jbeyerstedt: Changed back and added a comment about it.

src/endpoint.cpp
jbeyerstedt: Fixed for all `struct termios2 tc` and added a comment about the padding issue.

src/endpoint.cpp
lucasdemarchi: But that is more an implementation detail than anything else. It's a pain if the user of an API has to guess on each method if it's a class or object method. If all of them were bound to the class, or just some obvious ones like factory-like methods, then it'd be fine. If the compiler want's to optimize it away, fine, but don't enforce it on the source.

jbeyerstedt: OK, I reverted and disabled the whole clang-tidy rule for now.

jbeyerstedt:
All comments are resolved, the branch is rebased on the latest master and all the individual auto-fixes were squashed. So from my side, this is ready to be merged.

examples/heartbeat-print.cpp
lucasdemarchi: what about this comment?

src/binlog.cpp
```diff
@@ -77,25 +77,23 @@ void BinLog::_send_stop()
...
-    uint8_t source_system_id;
-    uint8_t source_component_id;
-    mavlink_remote_log_data_block_t *binlog_data;
+    uint32_t msg_id = 0;
+    uint8_t *payload = nullptr;
+    uint16_t payload_len = 0;
+    uint8_t trimmed_zeros = 0;
+    uint8_t source_system_id = 0;
+    uint8_t source_component_id = 0;
+    mavlink_remote_log_data_block_t *binlog_data = nullptr;
```
lucasdemarchi, line 86: as I said previously, leaving it unsigned is a feature - all of those are supposed to be assigned in one of the branches below. If we initialize them, then we basically lose the convenience of running valgrind to detect bugs.

meson.build
```diff
@@ -137,3 +137,21 @@ configure_file(
...
 subdir('examples')
+
+# Use clang-tidy to modernize code and check for common code style issues
+# meson keeps "fixing" the command, so we can't write "\." to escape the dot in the regex
+# or use any character classes. We want to match any path starting with "../src/".
+clangtidy = find_program('run-clang-tidy', required: false)
+if clangtidy.found()
+        run_target('clangtidy-autofix',
+                command : ['run-clang-tidy', '-p', meson.build_root(),
+                        '-header-filter', '^.{2}/src/.*',
```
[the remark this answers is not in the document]
jbeyerstedt: But I need the location of the compile_commands.json in the `-p` parameter, which is in the build root and has all file parts relative to that. So that part is actually right!?

The second issue is to have a regex, which matches these relative paths to the source_root. I just don't know how to do this.

src/binlog.cpp
jbeyerstedt: Sounds a bit strange to me. But I can leave that check out, if you like it better the way it was.

meson.build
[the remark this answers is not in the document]
lucasdemarchi: use meson.build_root() where you need the build dir and meson.source_root() where you need the source directory. You can do things like `meson.source_root() + 'src/*.h'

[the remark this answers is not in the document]
jbeyerstedt: But this doesn't work, because it'll add the full path instead of the relative one. I specifically don't need the full path, but the path to the source file as it is stated in the compile_commands.json file, which is generated by meson/ ninja.

src/binlog.cpp
lucasdemarchi: This recommendation is pretty bad. It came from Java or maybe from when gcc/clang wouldn't be able to help you. Nowadays compilers have -Wuninitialized / -Wmaybe-uninitialized and valgrind also helps you. If you force the initialization of local variables just for the sake of it, then it makes it harder for you to root cause bugs

meson.build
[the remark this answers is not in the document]
lucasdemarchi: Ok, let's leave it as is then and I can try to figure out later an alternative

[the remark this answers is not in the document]
jbeyerstedt: `@SOURCE_ROOT@` seems to give me a relative path. Still need to find a solution, that the `../` needs to be `\.\./` for the regex (but not today).

While we're at it, is `meson.current_build_dir()` or `meson.project_build_root()` the right path for the build directory? Because `meson.build_root()` is deprecated since 0.56.0.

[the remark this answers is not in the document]
jbeyerstedt: The only quick and not-so-hacky solution I came up with, was a small wrapper shell script, which manipulates the arguments to `run-clang-tidy` before calling it. Since the build root output by meson should always be an absolute path, this shouldn't interfere with the `-p`, but only with the `-header-filter` part.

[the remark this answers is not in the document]
lucasdemarchi: small wrapper script is fine

jbeyerstedt:
Removed the variable initialization auto-fixes all together, as well as "readability-isolate-declaration".

lucasdemarchi:
@jbeyerstedt there is still this one: https://github.com/mavlink-router/mavlink-router/pull/306/files#r763167529

I pulled the PR and tried to find what was the specific setting doing that change but I couldn't. Then I noticed we are using glob matches for clang-tidy. We probably shouldn't as this means different versions of clang-tidy not only behaves a little different regarding specific checks, but it also does completely different changes. That makes it harder to work with multiple toolchains/distros.

jbeyerstedt:
Sorry, overlooked that line. I'll also try to use the config file instead of having a very long line in the meson.build file to only enable specific checks instead of using the glob and removing checks again.

This should be what's currently enabled:
```
modernize-avoid-bind
modernize-deprecated-ios-base-aliases
modernize-loop-convert
modernize-make-shared
modernize-make-unique
modernize-pass-by-value
modernize-redundant-void-arg
modernize-replace-auto-ptr
modernize-use-auto
modernize-use-bool-literals
modernize-use-default-member-init
modernize-use-emplace
modernize-use-equals-default
modernize-use-equals-delete
modernize-use-nodiscard
modernize-use-nullptr
modernize-use-override
modernize-use-using
readability-braces-around-statements
readability-const-return-type
readability-container-size-empty
readability-delete-null-pointer
readability-identifier-naming
readability-implicit-bool-conversion
readability-inconsistent-declaration-parameter-name
readability-isolate-declaration
readability-make-member-function-const
readability-non-const-parameter
readability-qualified-auto
readability-redundant-declaration
readability-redundant-function-ptr-dereference
readability-redundant-member-init
readability-redundant-smartptr-get
readability-redundant-string-cstr
readability-redundant-string-init
readability-simplify-boolean-expr
readability-simplify-subscript-expr
readability-string-compare
readability-uniqueptr-delete-release
```

lucasdemarchi:
Yeah, if you move these to a .clang-tidy it's much easier to check what's enabled (and even call clang-tidy directly without using the build infra). We can use the wrapper script to pass the `-c <config-file>` to clang tidy if needed. Also maybe the issue you've seen with -header-filter is related to https://github.com/mesonbuild/meson/issues/957?

jbeyerstedt:
Regarding the header-filter: Getting a relative path is actually fine (because it's the same as in the compile_command.json). My issue is, that the argument I need to pass to run-clang-tidy (or clang-tidy directly) has to be a regex which matches the file paths as they are in the compile_commands (or as they are printed by clang-tidy). And since `.` has a special meaning in regex (any character) instead of the literal dot symbol, it needs to be escaped.

While it would be nice to have everything in the meson.build file, the small script works just fine, too.

lucasdemarchi:
Looks like you removed readability-implicit-bool-conversion from the configuration, but it's still being applied in the second commit? I see several s/if (bla)/if (bla !=0 )/

jbeyerstedt:
Sorry, I'm pretty sure, that I also reverted the corresponding commit (before squashing them into one commit during an interactive rebase), but something must have gone wrong. I'll take a look at it on Monday.

lucasdemarchi:
> Hopefully now everything is fine. I kept some of the explicit `!= nullptr` checks in main.cpp and mainloop.cpp because of two reasons: I think this makes the intention more clear and I rework big parts in the next PR anyways and didn't want to solve all the merge (actually rebase) conflicts. But all the explicit integer conversions (`!= 0` or `== 0`) are gone/ reverted now.

did I understand correctly that you did not just re-run clang-tidy on the previous state of the codebase?

lucasdemarchi (approved):

jbeyerstedt:
I didn't just re-run the clang-tidy to not re-introduce the formatting issues (there were one or two things you mentioned). But I ran the clang-tidy meson/ninja task on the changed state and it had no additional changes.

PR #308, opened 2021-12-13
Rework the internal flow of the config options

jbeyerstedt:
The configuration of mavlink-router has to be merged from config file(s) and CLI options. Previously the config file was parsed to one set of structs, passed to a function as a set of parameters (instead of the full struct), which converted the options to another config structure collecting a list of endpoints, which was finally used to actually create the endpoints. I think having two config structs is quite confusing and overly complicated.

This PR reworks the internal flow of the configuration parameters to be like this:
- Parse config file(s) and CLI options to one struct per endpoint type (UDP, TCP, UART and Log)
- Have a single config struct with the full, merged configuration (global options + one list of endpoints per type), which is properly owned by the main function instead of a global variable
- Mainloop still creates the actual endpoint instances from the single configuration
- Pass a full struct instead of multiple parameters to methods, where possible. 
This leads to less parameters and therefore makes future extensions easier.

Additional changes:
- Fixed (and extended) some log messages
- Move config file option tables to be owned by the endpoints instead of the main.cpp
- Rework the mainloop to only use one list of endpoints (instead of one for TCP + and another for everything else)
- Make use of std::vector instead of home-grown concatenated lists (easier to read/ understand and less error-prone when being extended)
- Configuration validation is a static method of each endpoint now instead of being scattered around multiple places in the main.cpp
- Endpoint setup is mainly done with one method and the config struct instead of multiple setters now

Sorry to have so many things in one PR, but they are closely related.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -18,6 +18,7 @@
 #include "endpoint.h"
 
 #include <algorithm>
+#include <utility>
 
 #include <arpa/inet.h>
 #include <assert.h>
@@ -97,8 +98,9 @@ static unsigned int ipv6_get_scope_id(const char *ip)
     return scope;
 }
 
-Endpoint::Endpoint(const char *name)
-    : _name{name}
+Endpoint::Endpoint(std::string type, std::string name)
+    : _type{std::move(type)}
+    , _name{std::move(name)}
 {
     rx_buf.data = (uint8_t *)malloc(RX_BUF_MAX_SIZE);
     rx_buf.len = 0;
@@ -146,7 +148,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
     uint8_t *payload, seq, payload_len;
 
     if (fd < 0) {
-        log_error("Trying to read invalid fd");
+        log_error("%s %s: Trying to read invalid fd", _type.c_str(), _name.c_str());
         return -EINVAL;
     }
 
@@ -177,7 +179,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
             return r;
         }
 
-        log_debug("%s [%d] got %zd bytes", _name, fd, r);
+        log_debug("%s [%d]%s: got %zd bytes", _type.c_str(), fd, _name.c_str(), r);
         rx_buf.len += r;
     }
 
@@ -297,7 +299,7 @@ int Endpoint::read_msg(struct buffer *pbuf, int *target_sysid, int *target_compi
     *target_compid = -1;
 
     if (msg_entry == nullptr) {
-        log_debug("No message entry for %u", *msg_id);
+        log_debug("%s [%d]%s: No message entry for %u", _type.c_str(), fd, _name.c_str(), *msg_id);
     } else {
         if (msg_entry->flags & MAV_MSG_ENTRY_FLAG_HAVE_TARGET_SYSTEM) {
             // if target_system is 0, it may have been trimmed out on mavlink2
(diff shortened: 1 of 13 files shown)


src/main.cpp
```diff
@@ -714,92 +510,131 @@ static int parse_mode(const char *val, size_t val_len, void *storage, size_t sto
...
+    }
+
+    ret = conffile.extract_options("General", option_table_log, ARRAY_SIZE(option_table_log),
+                                   &config.log_config);
     if (ret < 0) {
         return ret;
     }
 
     iter = {};
     pattern = "uartendpoint *";
```
lucasdemarchi, line 436: this should probably be a constant in UartEndpoint

src/endpoint.cpp
```diff
@@ -97,8 +98,9 @@ static unsigned int ipv6_get_scope_id(const char *ip)
     return scope;
 }
 
-Endpoint::Endpoint(const char *name)
-    : _name{name}
+Endpoint::Endpoint(std::string type, std::string name)
+    : _type{std::move(std::move(type))}
+    , _name{std::move(std::move(name))}
```
lucasdemarchi, line 156: should be just one std::move(), no?

src/endpoint.cpp
```diff
@@ -541,13 +541,52 @@ void Endpoint::log_aggregate(unsigned int interval_sec)
     }
 }
 
+UartEndpoint::UartEndpoint(std::string name)
+    : Endpoint{ENDPOINT_TYPE_UART, std::move(name)}
+{
+    ; // nothing else to do here
```
lucasdemarchi, line 547: and no need for an empty statement neither

src/mainloop.cpp
```diff
@@ -557,7 +523,7 @@ void Mainloop::_del_timeouts()
     }
 }
 
-void Mainloop::_add_tcp_retry(TcpEndpoint *tcp)
+void Mainloop::add_tcp_retry(TcpEndpoint *tcp)
```
lucasdemarchi, line 523: Now that our code I much cleaner, we probably want to remove this from Mainloop. No need to keep the Timeout in the Mainloop rather than TcpEndpoint

lucasdemarchi:
Overall looks good to me and makes the code much cleaner. Thanks.

src/main.cpp
jbeyerstedt: Makes sense, right 😉

src/endpoint.cpp
jbeyerstedt: Sorry, overlooked that one.

src/endpoint.cpp
jbeyerstedt: OK, I'll leave the comment tough, so it's clear, that the c-tor is intentionally empty.

src/mainloop.cpp
jbeyerstedt: Just this method (which will call `Mainloop::get_instance().add_timeout()` then) or or the whole timeout stuff?

jbeyerstedt: Just implemented this as well. Because `add_timeout()` is also used by the mainloop itself, I just moved the all of the reconnect handling to the TCP endpoint. The `valid` flag now indicates, that the endpoint can be removed.

jbeyerstedt:
The last commit (moving the TCP client reconnect handling to the TcpEndpoint class) also fixes some issue with the TCP endpoints. When a client connected to the TCP server while a TCP client endpoint was trying to reconnect, the TCPEndpoint's IP and port got lost somehow. I think it had something to do with the TcpEndpoint not being added to the endpoints list when the remote side wasn't available during the first connection attempt. But it works now, so the actual reason it not so important.

Issue #309, opened 2021-12-17
Compiled from master this morning and no logs


highfreq:
I just compiled and installed from master and log is not written anymore. Configuration file is same as before with log destination directory configured and dialect specified.
If i start mavlink-router from command line i see no log being created or any feedback from the software telling me log has been started or similar.
If i start it with the -l option from command line than it creates the file.

This is what i have for logging in my main.conf

MavlinkDialect=auto
Log=/home/pi/Logs
LogMode=always

It looks like it ignores Log configuration from conf file.

lucasdemarchi:
Fix in #310. Thanks for the report

PR #310, opened 2021-12-20
logendpoint: fix not not parsing config file correctly

lucasdemarchi:
We changed the storage type from char* to std::string bug failed to
update the configfile and arg parse.

Fix: #309

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -228,7 +228,7 @@ static int parse_argv(int argc, char *argv[], Configuration &config)
             break;
         }
         case 'l': {
-            config.log_config.logs_dir = strdup(optarg);
+            config.log_config.logs_dir.assign((const char *)optarg);
             break;
         }
         case 'g': {
--- a/src/logendpoint.cpp
+++ b/src/logendpoint.cpp
@@ -45,7 +45,7 @@
 #define MAX_RETRIES   10
 
 const ConfFile::OptionsTable LogEndpoint::option_table[] = {
-    {"Log", false, ConfFile::parse_str_dup, OPTIONS_TABLE_STRUCT_FIELD(LogOptions, logs_dir)},
+    {"Log", false, ConfFile::parse_stdstring, OPTIONS_TABLE_STRUCT_FIELD(LogOptions, logs_dir)},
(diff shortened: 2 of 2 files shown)


highfreq:
Recompiled now and same problem, no logs

lucasdemarchi:
Did you include the patch in that PR ?

highfreq:
No, compiled from master, i tought it was already merged.

highfreq:
Should i clone tip-fix-logging branch?

highfreq:
Ok compiled from that branch and now it logs, still makes hundreds of useless 56k files over and over, i hoped it was fixed

lucasdemarchi:
No, we didn't touch that part yet.

highfreq:
Wonder why not yet, it is the really only broken part in mavlink-router but it doesn't get fixed....................

khancyr:
@highfreq , PR and funding welcome

highfreq:
> @highfreq , PR and funding welcome

If you fund my days debugging and try to have logging working and report bugs, i can fund your dev time, no prob.

src/main.cpp
```diff
@@ -228,7 +228,7 @@ static int parse_argv(int argc, char *argv[], Configuration &config)
             break;
         }
         case 'l': {
-            config.log_config.logs_dir = strdup(optarg);
+            config.log_config.logs_dir.assign((const char *)optarg);
```
jbeyerstedt, line 231: Why is this better than `config.log_config.logs_dir = std::string(optarg)`? (Just out of curiosity.)

highfreq:
> @highfreq , PR and funding welcome

Hey Genius, fixed the issue, recompiled and now it works as it should, took me under 3 hours :)

jbeyerstedt (approved):
LGTM and printing the config where it is used works. (I have no FMU or simulation at hand right now.)

src/main.cpp
jbeyerstedt: OK. I only found the `=` assignment while skipping through the code, but I seems to have used both variants. So I might need to change it in the future to one consistent style.
As far as I have read, both variants should be the same, because `assign()` and the std::string c-tor both copy the C string contents. I just was curious, if I have missed something.

PR #313, opened 2021-12-22
Update endpoint.cpp

mark-nick-o:
cleared warnings

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -506,14 +506,14 @@ void Endpoint::print_statistics()
     printf("%s Endpoint [%d]%s {", _type.c_str(), fd, _name.c_str());
     printf("\n\tReceived messages {");
     printf("\n\t\tCRC error: %u %u%% %luKBytes", _stat.read.crc_error,
-           (_stat.read.crc_error * 100) / read_total, _stat.read.crc_error_bytes / 1000);
+           (_stat.read.crc_error * 100) / read_total, (long unsigned int)_stat.read.crc_error_bytes / 1000);
     printf("\n\t\tSequence lost: %u %u%%", _stat.read.drop_seq_total,
            (_stat.read.drop_seq_total * 100) / read_total);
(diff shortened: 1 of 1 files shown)


jbeyerstedt:
Thanks, but could you please give some more information about the issue you're trying to fix? Especially on which kind of system you're trying to compile mavlink-router (CPU architecture, operating system, 64 vs 32 bit OS).

I assume you got some compiler warnings about the printf format specifiers not matching the size of the variable. Printf expects an unsigned long `%lu`, but the variable is `uint64_t`. I'm pretty sure (but haven't checked back), that this is fine on a 64-bit linux system, but not on a 32-bit linux system (like Raspbian), because the length of an `unsigned long` is different depending on the operating systems architecture size.

If I'm correct with my assumption, it's impossible to have a single line of `printf` code, which has no warning on both 32- and 64-bit systems. And you would be down-casting from a 64-bit unsigned integer to a 32-bit unsigned integer, which might lead to unexpected numbers. (And I personally don't like C-style casts in C++ code. But using a static_cast might give you a warning (or even an error) about the down-casting.)

mark-nick-o:
Thanks for your message

Yes don’t worry it compiles and works fine

I’m using raspberry pi and it’s the difference between 32bit and 64bit so I have to change it in my version to compile without warnings
you can define different architectures to do it but it doesn’t matter, it works for me and you  the code is fine, it’s much more work to do that for every possible use.

I don’t think it uses the upper part of that integer I best check it

You can also type cast it as static_cast it’s the same, sorry I didn’t notice you were in c++

Mostly I just keep these versions and don’t report them back but someone said to me to start feeding back the changes

All the best

Mark

lucasdemarchi:
We could use PRIu64 to print uint64_t so it's expanded to the correct type. Although I'm fine with the cast too

lucasdemarchi:
Humn... maybe we should add to CI passing -m32, which should take care of 32 vs 64 bit differences

lucasdemarchi:
See #321

PR #314, opened 2021-12-30
Tweak clang-format

lucasdemarchi:
A few formatting and struggled with while doing some refactors in #143. First commits are actually from that PR and can be ignored here, I'm only including them because otherwise we would have some needless conflicts.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -20,6 +20,11 @@ AllowShortFunctionsOnASingleLine: Inline
 AlignConsecutiveDeclarations: false
 AlignConsecutiveAssignments: false
 AlignConsecutiveMacros: Consecutive
+
+BinPackArguments: false
+AllowAllParametersOfDeclarationOnNextLine: false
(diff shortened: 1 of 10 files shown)


.clang-format
```diff
@@ -20,6 +20,11 @@ AllowShortFunctionsOnASingleLine: Inline
 AlignConsecutiveDeclarations: false
 AlignConsecutiveAssignments: false
 AlignConsecutiveMacros: Consecutive
+
+BinPackArguments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowAllArgumentsOnNextLine: false
```
jbeyerstedt, line 26: I think, I also fiddled around with these parameters in my previous commits, because the arguments list inconsistencies bothered me. But didn't want to change too much about the format rules back then.
tl;dr: This is fine.

jbeyerstedt (approved):

PR #315, opened 2021-12-30
Fix assert crash when stopping log

lucasdemarchi:
When we exit mavlink-router we call logger->stop() to stop
and properly finalize the flight stack log. However this
call may want to send something to the flight stack to stop
sending the log.

We don't care about that call as if we are really stopping the flight
stack should detect the other side died/disconnected after a timeout.
Instead of marking the mainloop as not initiliazed upon signal,
serialize it with the other shutdown task in the main loop.

This fixes the crash:
    Logging target system_id=1 on 00001-2021-12-29_23-57-32.bin
    mavlink-routerd: ../src/mainloop.h:77: static Mainloop&Mainloop::get_instance(): Assertion `_initialized' failed.
    Aborted

### DIFF
--- a/src/mainloop.cpp
+++ b/src/mainloop.cpp
@@ -76,7 +76,6 @@ Mainloop &Mainloop::instance()
 void Mainloop::request_exit(int retcode)
 {
     _retcode = retcode;
-    _initialized = false;
     should_exit.store(true, std::memory_order_relaxed);
 }
 
--- a/src/mainloop.cpp
+++ b/src/mainloop.cpp
@@ -63,6 +63,11 @@ Mainloop &Mainloop::init()
     return _instance;
 }
 
+void Mainloop::teardown()
+{
+    _initialized = false;
+}
+
 Mainloop &Mainloop::instance()
 {
     return _instance;
--- a/src/mainloop.h
+++ b/src/mainloop.h
@@ -84,6 +84,12 @@ public:
      */
     static Mainloop &init();
 
+    /*
+     * De-initialize singleton so we can start a fresh on the same
+     * thread
+     */
+    static void teardown();
+
     static Mainloop &instance();
 
     /*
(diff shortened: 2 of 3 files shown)


khancyr:
Doesn't work for me : 
```
mavlink-router version v2-146-gebf5b4a
Opened TCP Client [4]to_SITL: 127.0.0.1:5762
File 00001-2021-12-30_21-51-14.bin not read-only yet, marking as RO
Logging target system_id=1 on 00002-2021-12-31_02-12-56.bin
^Cmavlink-routerd: ../src/mainloop.h:78: static Mainloop& Mainloop::get_instance(): Assertion `_initialized' failed.
[1]    182433 IOT instruction (core dumped)  mavlink-routerd -c ../mavlink-router.conf -t 0
```
I used CTRL-C to quit.
Mavlink-router was connected to ArduPilot SITL with this config file : 
```
[General]
MavlinkDialect=ardupilotmega
Log=/tmp/flightstack
ReportStats=false
FcuId=1

[TcpEndpoint to_SITL]
Mode = Normal
Address = 127.0.0.1
Port = 5762
RetryTimeout=2
```
and launched with `mavlink-routerd -c ../mavlink-router.conf -t 0 `

lucasdemarchi:
@khancyr are you sure you executed the mavlink-router binary after applying this PR?

khancyr:
@lucasdemarchi yep.
I just did another test by putting printf where you did the changes
```
mavlink-router version v2-146-gebf5b4a+
Opened TCP Client [4]to_SITL: 127.0.0.1:5762
Logging target system_id=1 on 00005-2021-12-31_13-40-58.bin
^Cno _initialized = false;_initialized = false;
mavlink-routerd: ../src/mainloop.h:78: static Mainloop& Mainloop::get_instance(): Assertion `_initialized' failed.
[1]    184379 IOT instruction (core dumped)  mavlink-routerd -c ../mavlink-router.conf -t 0

```
so it looks good but we still have a get_instance() triggered before stop

khancyr:
The remaining error comes from TcpEndpoint::close() that call Mainloop::get_instance().remove_fd(fd); Then it crash on the get_instance() there !

lucasdemarchi:
Oh. Ok, thanks for testing. So we should move this further down, or even just remove the setting back to -1

Issue #163, opened 2019-04-10
Logging not working for MavId other than 1


anandbaskaran:
I am trying to log the mavlink messages of the drone with the following settings in the config file.

```
[General]
MavlinkDialect=auto
Log=/var/log/flight-stack
LogMode=always
DebugLogLevel=info
```

But this fails with the message `MAV_CMD_LOGGING_START result(2) is different than accepted`. Change back the `SYS_ID` to 1 resolves this problem.

lucasdemarchi:
ArduPilot or PX4?

ashwinsushil:
Hey @anandbaskaran @lucasdemarchi I have the exact same issue.
I am using PX4 and my configurations are:
```
[General]
TcpServerPort=5760
ReportStats=true
MavlinkDialect=common
Log=/var/lib/mavlink-router/
```
Let me know if you need additional information to close this issue.

PR #316, opened 2021-12-31
Allow to set fcuid to log explicitly

khancyr:
this is mandatory for multi fcu configuration to select the right one to log

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -37,6 +37,11 @@
 #       If set to while-armed, a new log file is created whenever the vehicle is
 #       armed, and closed when disarmed.
 #
+#   LogSystemId
+#       Set the FCU Mavlink Id for logging. The Id should be between 0 and 255.
+#       No default value.
+#       If absent, the FCU Id is settled to the first autopilot that we get a heartbeat from.
+#
 #   MinFreeSpace
 #       The Log Endpoint will delete old log files until there are MinFreeSpace bytes
 #       available on the storage device of the logs. Set to 0 to ignore this limit.
--- a/src/logendpoint.h
+++ b/src/logendpoint.h
@@ -44,6 +44,7 @@ struct LogOptions {
     MavDialect mavlink_dialect{MavDialect::Auto}; // conf "MavlinkDialect"
     unsigned long min_free_space;                 // conf "MinFreeSpace"
     unsigned long max_log_files;                  // conf "MaxLogFiles"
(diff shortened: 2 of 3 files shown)


examples/config.sample
```diff
@@ -53,6 +53,12 @@
 #       most verbose.
 #       Default:<info>
 #
+#   FcuId
```
lucasdemarchi, line 56: suggestion: LogSystemId, and move it near to "LogMode".

src/logendpoint.h
```diff
@@ -44,6 +44,7 @@ struct LogOptions {
     MavDialect mavlink_dialect{MavDialect::Auto}; // conf "MavlinkDialect"
     unsigned long min_free_space;                 // conf "MinFreeSpace"
     unsigned long max_log_files;                  // conf "MaxLogFiles"
+    int fcu_id{-1};                                   // conf "LogMasterId"
```
lucasdemarchi, line 47: LogMasterId?

src/binlog.cpp
```diff
@@ -101,10 +101,13 @@ int BinLog::write_msg(const struct buffer *buffer)
         source_component_id = msg->compid;
     }
 
-    /* set the expected system id to the first autopilot that we get a heartbeat from */
-    if (_target_system_id == -1 && msg_id == MAVLINK_MSG_ID_HEARTBEAT
-        && source_component_id == MAV_COMP_ID_AUTOPILOT1) {
-        _target_system_id = source_system_id;
+    /* set the expected system id to the first autopilot that we get a heartbeat from if fcu_id isn't settled */
+    if (_target_system_id == -1) {
```
lucasdemarchi, line 105: maybe easier to just set _target_system_id in the constructor looking at the config value we receive as parameter?

lucasdemarchi:
Overall looks good, thanks. This should fix #163

src/logendpoint.h
khancyr: leftover from some thinking on the name...

examples/config.sample
khancyr: ok

src/binlog.cpp
khancyr: done

khancyr:
tested with ArduPilot SITL: 
test target id out of range
test SITL ID 1, mavlink-router target 2 : no log
test SITL ID 2, mavlink-router target 2 : log
test SITL ID 42, mavlink-router no target settle : lLogging target system_id=42

PR #318, opened 2021-12-31
Mitigate binlog storm issue

khancyr:
This provide a way to mitigate the issue with ArduPilot binlog.
The issue is that if we arm an ArduPilot vehicle and disarm, every 15s the binlog will end and a new one will start. This lead to generate a lot of mostly empty logs. 
This comes from ArduPilot, https://github.com/ArduPilot/ardupilot/blame/7b5ab8bb8cfd6fded545beb49585c54b38b04999/libraries/AP_Logger/AP_Logger_Backend.cpp#L454
ArduPilot logging log critical change after disarm and push the log on all Log Backend including Mavlink
The critical change are small: 1 state message or 1 mode change and the logging FMT.
We don't have way to filters those. 
The proposed change detect a small log (< 400msg, current FMT is at more than 300msg), and instead of just closing it to create a new one, delete it as it doesn't got anything useful.

The issue could be simply test by using sitl and setting LOG_BACKEND_TYPE 3, reboot, arm, and wait auto disarm.

### DIFF
--- a/src/binlog.cpp
+++ b/src/binlog.cpp
@@ -224,3 +224,25 @@ void BinLog::_restart()
     stop();
     start();
 }
+
+bool BinLog::_alive_timeout()
+{
+    if (_timeout_write_total == _stat.write.total) {
+        // we probably just got FMT plus a MODE change
+        // https://github.com/ArduPilot/ardupilot/blame/7b5ab8bb8cfd6fded545beb49585c54b38b04999/libraries/AP_Logger/AP_Logger_Backend.cpp#L454
+        if (_last_acked_seqno > 400) {
+            log_warning("No Log messages received in %u seconds restarting Log...", ALIVE_TIMEOUT);
+            stop();
+            start();
+        } else {
+            // silent restart
+            LogEndpoint::_delete_last_log();
+            stop();
+            start();
+        }
+
+    }
+
+    _timeout_write_total = _stat.write.total;
+    return true;
+}
--- a/src/binlog.h
+++ b/src/binlog.h
@@ -42,6 +42,7 @@ protected:
     bool _logging_start_timeout() override;
 
     const char *_get_logfile_extension() override { return "bin"; };
+    bool _alive_timeout() override;
 
 private:
     uint32_t _last_acked_seqno = 0;
--- a/src/logendpoint.cpp
+++ b/src/logendpoint.cpp
@@ -41,7 +41,6 @@
 
 #include "mainloop.h"
 
-#define ALIVE_TIMEOUT 5
 #define MAX_RETRIES   10
 
 const ConfFile::OptionsTable LogEndpoint::option_table[] = {
--- a/src/logendpoint.cpp
+++ b/src/logendpoint.cpp
@@ -125,6 +125,23 @@ void LogEndpoint::mark_unfinished_logs()
     closedir(dir);
 }
 
+void LogEndpoint::_delete_last_log()
+{
+    fsync(_file);
+    close(_file);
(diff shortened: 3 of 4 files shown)


lucasdemarchi:
I was trying to pinpoint the place in ardupilot that was pushing the additional log messages. I'm not confident this is because of messages after disarm like explained here, because it's being pushed every time we send a LOG_START message, even if LOG_DISARMED is not set. 

This behavior only happens when LOG_DISARMED is not set, because ardupilot simply stops responding and then after each START it logs a few messages.

I was comparing the end result with a File backend, and there this doesn't happen. So I'm actually leaning towards a bug in ardupilot that we should fix rather than implementing a workaround in ardupilot. I'd be fine with deleting the a 0-sized file, but I'm not very happy with "this is small, we can delete".

khancyr:
I don't have zero size log with ArduPilot master, they are like 64k with FMT msg as explained

The line I mentioned is the right culprit as that the one that allow AP_Logger to use the _writeblocks() function in response to LOG_START msg even if disarmed. 
If you disable the if block mentioned on ArduPilot, the log storm don't happen anymore as AP_Logger won't have permission to write until we are armed.

The fix on ArduPilot should be to prevent writing critical messages on Mavlink backend.
Why I chose to do a patch on Mavlink router too is because the current v4.1 series is touched by the issue and we won't have a backport

lucasdemarchi:
> I don't have zero size log with ArduPilot master, they are like 64k with FMT msg as explained

yep, the same here. What I meant was: if the log was 0-sized, then I'd be ok with the approach of deleting/re-using the file. But having to check for an arbitrary small size seems like a workaround to an issue that should be fixed somewhere else. 

> The line I mentioned is the right culprit as that the one that allow AP_Logger to use the _writeblocks() function in response to LOG_START msg even if disarmed.
> If you disable the if block mentioned on ArduPilot, the log storm don't happen anymore as AP_Logger won't have permission to write until we are armed.

Indeed. And as far as I can see this was introduced to fix a problem with **file** log rotation, now causing issue with mavlink log, because our client side is doing rotation on START/STOP and if the flight stack logs something on START and suddenly stops, we can't differentiate it from a link disconnection.

> The fix on ArduPilot should be to prevent writing critical messages on Mavlink backend.

or to all backends? Couldn't we just log that block on the rotated file?

khancyr:
> a workaround to an issue that should be fixed somewhere else.

Yep, this is a workaround as the protocol is weak. I intend to implement a status message so we can have a feedback if the logger is start or stop on ArduPilot side. That would help, but unfortunately, it won't work with previous release. 
That is my main issue, how to patch this for the v4.1 series ? The main user off mavlink-router are mostly industrial, therefore providing a solution for the v4.1 would be great.

> or to all backends?
Just mavlink once, doing it on logfile backend would mean to bring back the mentionned issue on ArduPilot.

> Couldn't we just log that block on the rotated file?
Which file ? the one on mavlink-router side ?

lucasdemarchi:
@khancyr  I think stopping the log while disarmed even when instructed not to do that would be the best fix. I can take a look on ardupilot side how to do that.

On Ardupilot it's also possible to backport patches to previous releases. I did that some time ago when I fixed an issue with I2C. https://github.com/ArduPilot/ardupilot/projects/19 - should list what needs to be backported.

khancyr:
@lucasdemarchi I have submit a PR on ArduPilot with a fix, if you want to look at it
The main issue on backport is that they are quite rare and mostly for critical fix. Hopefully, this one should be simple enough to be able to be backported

PR #321, opened 2022-01-06
Fix print warnings due to 32 vs 64 bits

lucasdemarchi:


jbeyerstedt:
While you're at it, please change https://github.com/mavlink-router/mavlink-router/blob/master/src/mainloop.cpp#L197 to signed integer.
```diff
- log_debug("Message %u to unknown sysid/compid: %u/%u",
+ log_debug("Message %u to unknown sysid/compid: %d/%d",
            buf->curr.msg_id,
            buf->curr.target_sysid,
            buf->curr.target_compid);
```

jbeyerstedt:
Thanks, LGTM.

PR #322, opened 2022-01-07
ci: enable Werror

lucasdemarchi:
In general I think -Werror can't be the default due to different
compilers being used. However, for CI it's good to enable it because we
control the exact version we are using.

Enable -Werror to avoid introducing warnings.


khancyr (approved):
LGTM

lucasdemarchi:
But before committing we should now fix the warnings for it to be warnings-clean

Issue #240, opened 2020-11-23
Add message de-duplication/ redundant data links


jbeyerstedt:
Some days ago there was a small discussion about the support of redundant telemetry links to the FMU, so let's move the requirements here.
From my understanding, there are two similar, but different use cases, which a de-duplication feature should solve.

## Redundant Data Links
In some use cases it might be useful to have a redundant telemetry link between UAS and ground station. Both connections might be handled by a companion computer running mavlink-router on the UAS and another mavlink-router instance on the ground.
The current implementation will either lead to message loops or systems being falsely detected on an endpoint, because all incoming messages are routed out on all endpoints. And even when both links are properly used simultaneously, the FMU and ground station will still receive messages multiple times.

A solution for de-duplication might be to monitor the link quality (HEARTBEAT timeouts, estimate drop rate, etc) any only pass on messages of one of the redundant links. This would mean, that the config has to specify the "type" of the endpoint more precisely (e.g. grouping them to one redundant link and/or specifying communication directions).

## Mesh Network
Another use case is a mesh network for UTM and telemetry purposes. This will lead to even more advanced de-duplication requirements, because multiple "remote" mavlink-router instances will be received on one wireless link. The goal would be to have a "wave front" of each message propagating through the network, but no duplicated messages behind the wave front.

The amateur radio community already has some solutions for this issue. One of them is to cache the message counter (which is included in each MAVLink message) of each message for some time. This cache would be needed for each endpoint and a message is only transmitted through that endpoint, if it wasn't already transmitted once (the message counter isn't higher, than the cached value).

christophkTEA:
The mavlink message signing mechanism could be used for avoiding duplicate messages (https://mavlink.io/en/guide/message_signing.html). You can already tell, by checking the message signature, if you get the same message over the same link twice. If that is the case the router just drops the new message. If the router compares the mavlink messages to those previously received messages and they are the same except for the link id and signature field, then the router has received the same message before over a different link and can discard the new message.
This would also add to the security of such a network because each message is signed with the private key of the sender.

jbeyerstedt:
That's a good idea, I just have to check, if message signing is already activated in PX4 or how to activate it. (I've already seen, that there is a MAVLink message to setup the secret key).
But the issue with using the signature is, that not everybody populates the value (as far as I know). So we might need a fallback.

[...]

PR #323, opened 2022-01-12
Add optional message de-duplication

jbeyerstedt:
As discussed in #240, parallel data links have some special needs. One of which is de-duplication of incoming messages.

This implementation has been used for some time and works just fine. I just needed to make some changes to make it compatible with the latest master, squashed some of Julian's commits and added some comments on the behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,14 @@ Message filters:
   - AllowSrcCompOut: If set, only allow messages from the listed MAVLink source
     compoent IDs to be sent via this endpoint
 
+Message de-duplication:
+
+  - If enabled, each incoming message is checked, whether another copy was
+    already received the last `DeduplicationPeriod` milliseconds ago. If it's
+    already known, the message will be dropped as it was never received and the
+    timeout counter for that message will be reset. Messages are identified via
+    their `std::hash` value.
+
 ### Flight Stack Logging
 
 Mavlink router can also collect flight stack logs. It supports collecting both
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -33,6 +33,12 @@
 # Default: <info>
 #DebugLogLevel = info
 
+# Enable de-duplication of incoming messages. If a message is received another
+# time in the configured time period (in milliseconds), it will be dropped. The
+# second message will reset the timer.
+# Default: 0 (de-duplication disabled)
+#DeDuplicationPeriod = 0
+
 ## TCP Server Endpoints
 
 # Listen for TCP connections on this port. Set to 0 to disable.
--- a/src/dedup.cpp
+++ b/src/dedup.cpp
@@ -1,0 +1,99 @@
+/*
+ * This file is part of the MAVLink Router project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -234,6 +234,14 @@ Message filters:
   - AllowSrcCompOut: If set, only allow messages from the listed MAVLink source
     compoent IDs to be sent via this endpoint
 
+Message de-duplication:
+
+  - If enabled, each incoming message is checked, whether another copy was
+    already received the last `DeduplicationPeriod` milliseconds ago. If it's
+    already known, the message will be dropped as it was never received and the
+    timeout counter for that message will be reset. Messages are identified via
+    their `std::hash` value.
```
lucasdemarchi, line 243: Either here or in the config.sample we need a note about potentially harmful side effect: if it's a link with packet loss, the re-transmission from either the transport or by the flightstack/mavlink-component could cause the re-transmissions to be dropped.

One case I think this is particularly bad: log streaming over mavlink. Example with the Ardupilot log format:

```
    <message id="184" name="REMOTE_LOG_DATA_BLOCK">
      <description>Send a block of log data to remote location.</description>
      <field type="uint8_t" name="target_system">System ID.</field>
      <field type="uint8_t" name="target_component">Component ID.</field>
      <field type="uint32_t" name="seqno" enum="MAV_REMOTE_LOG_DATA_BLOCK_COMMANDS">Log data block sequence number.</field>
      <field type="uint8_t[200]" name="data">Log data block.</field>
    </message>
```

Suppose the log is being saved by another endpoint (example: mavproxy) and not mavlink-router itself.

1. Flight stack sends seqno 1, we route it to the endpoint of whatever is saving the logs. 
2. Message gets lost
3. Logger checks seqno and notice it never received seqno, but after some time it receives seqno 2. Logger requests seqno 1 to be resent
4. Flight stack sends seqno 1
5. That message with seqno 1 will be dropped since it's exactly the same as the one previously sent

Logger here was an example, but there are probably several commands from the mavlink spec that would suffer from a similar issue. I'm not blocking the feature, but we need to document that this is known, and the user probably needs to tune the DeDuplicationPeriod for their application

examples/config.sample
```diff
@@ -33,6 +33,12 @@
 # Default: <info>
 #DebugLogLevel = info
 
+# Enable de-duplication of incoming messages. If a message is received another
+# time in the configured time period (in milliseconds), it will be dropped. The
+# second message will reset the timer.
+# Default: 0 (de-duplication disabled)
```
lucasdemarchi, line 39: please also mention what be a good value for "normal use"

lucasdemarchi:
The comments I mentioned above can be handled on top since they are just documentation.

README.md
jbeyerstedt: I'm not 100% sure, but wouldn't the re-transmitted packet have a different [packet sequence number](http://mavlink.io/en/guide/serialization.html#mavlink2_packet_format) in the MAVLink header? Since the hash is calculated over the full mavlink packet, this should give a different hash on the re-transmitted packet.

lucasdemarchi: Humn... right. I think I was misled by the `seqno` in this specific message - it needs to be the same in the re-transmitted message. However that is part of the payload, not the header, and the packet sequence number should take care of making the hash different as you noticed.

However, there is still the fact that packet sequence number is a uint8 only, so only covers 256 messages before rotating, which may be a very short time: that could cause occasional issues. However this is very less likely than the original issue I was worried about. Thanks.

jbeyerstedt: We actually came across the issue with the 8 bit header sequence number during the initial tests of de-duplication. Julian's first draft didn't remove the hashes from the cache, so heartbeats got dropped after some time. A grounded vehicle has a quite static heartbeat message and after some time all sequence numbers in the header were seen. IIRC this took a few minutes tough. But I'll try to test out a recommended max value.

lucasdemarchi: during those few minutes you experienced a steadily increasing "packet loss". I guess the safe configuration is to avoid the situation the sequence number could wrap and make the hashes collide... Take the stream with the highest rate that could be problematic due to its content, and define the timeout as less than 1/rate.

PR #325, opened 2022-01-14
UdpClient: switch back to broadcast on inactivity

lucasdemarchi:
More details on last commit message and on !10.

Fix: #110

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -904,6 +904,13 @@ UdpEndpoint::UdpEndpoint(std::string name)
     bzero(&sockaddr6, sizeof(sockaddr6));
 }
 
+UdpEndpoint::~UdpEndpoint()
+{
+    if (nomessage_timeout) {
+        Mainloop::get_instance().del_timeout(nomessage_timeout);
+    }
+}
+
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
We should probably have a TimeoutImpl to be used by the main loop and a Timeout to wrap it for other users. This would allow us to remove those *_timeout  methods from the Mainloop class

jbeyerstedt (approved):
LGTM and even works with the IPv6 broadcast equivalent (ff02::1). I'll adapt the specification PR accordingly (all non-unicast addresses will have this "broadcast until first response" behavior).

jbeyerstedt:
You're right about the additional cleanup work, but I don't known how much effort this would be. So I'll be fine merging this PR for now and doing the cleanup in a later commit.

lucasdemarchi:
> You're right about the additional cleanup work, but I don't known how much effort this would be. So I'll be fine merging this PR for now and doing the cleanup in a later commit.

yeah... I was tempted to start it, but that will mean some pending PRs we want to be in v3 will need to be rebased. I will wait them to get merged before changing this.

PR #330, opened 2022-01-17
Support redundant data links: Endpoint groups

jbeyerstedt:
As discussed in #240, parallel/ redundant data links will lead to one link not being used any more with the current routing rules. Since mavlink-router can not determine the network topology by itself, it has to be provided in the configuration.

The list of connected systems should be the same on all of the endpoints terminating the parallel data links. If that's the case, routing rule 1 (don't send data on an endpoint, if it's source system is connected on that endpoint) should prevent data being received on one of the grouped endpoint to be sent via all other group members.

A hierarchical structure of endpoints (GroupEndpoint containing the group members) would probably be a cleaner solution. But this should work as a proof of concept and doesn't change too much of the general routing architecture.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,14 @@ Message filters:
   - AllowSrcCompOut: If set, only allow messages from the listed MAVLink source
     compoent IDs to be sent via this endpoint
 
+Endpoint groups:
+
+  - Multiple endpoints can be configured to be in the same endpoin group.
+    Endpoints in the same group will share the same list of connected systems.  
+    When using two (or more) **parallel data links**, e.g. LTE and telemetry
+    radio, the endpoint **must** be grouped on both sides. Otherwise one link
+    will not be used any more because of routing rule 1.
+
 ### Flight Stack Logging
 
 Mavlink router can also collect flight stack logs. It supports collecting both
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -105,6 +105,13 @@
 # Default: Empty list (disabled)
 #AllowSrcCompOut = 
 
+# Group parallel/ redundant data links to use the same list of connected
+# systems. This is needed to prevent messages from one of the parallel links
+# being send back on the other one right away.
+# Set the same name (arbitrary string) on multiple endpoints to group them.
+# Default: Empty (no group)
+#Group = 
+
 ## Example
 [UartEndpoint alpha]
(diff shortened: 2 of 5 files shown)


jbeyerstedt:
I still need to do extensive testing, but wanted to post this for comments on the approach.

jbeyerstedt:
Works just fine (at least with one UAV since I only have one FMU here right now).

jbeyerstedt:
@lucasdemarchi Any comments on this and #323? It would be really cool, if mavlink-router v3 has support for redundant data links. Both PRs are actively used in our systems and work fine together.

README.md
```diff
@@ -234,6 +234,14 @@ Message filters:
   - AllowSrcCompOut: If set, only allow messages from the listed MAVLink source
     compoent IDs to be sent via this endpoint
 
+Endpoint groups:
+
+  - Multiple endpoints can be configured to be in the same endpoin group.
```
lucasdemarchi, line 239: endpoint

examples/config.sample
```diff
@@ -105,6 +105,13 @@
 # Default: Empty list (disabled)
 #AllowSrcCompOut = 
 
+# Group parallel/ redundant data links to use the same list of connected
+# systems. This is needed to prevent messages from one of the parallel links
+# being send back on the other one right away.
```
lucasdemarchi, line 110: sent

src/mainloop.cpp
```diff
@@ -390,6 +390,19 @@ bool Mainloop::add_endpoints(const Configuration &config)
...
+        if (e->get_group_name().empty()) {
+            continue;
+        }
+
+        for (auto other : g_endpoints) { // find other endpoints in group
+            if (other != e && e->get_group_name() == e->get_group_name()) {
+                e->link_group_member(other);
+            }
+        }
+    }
```
lucasdemarchi, line 404: I am seeing the setup of a group, but I'm missing where you change the route rules to avoid sending the message to endpoints if they belong to the same group. Am I missing something?

lucasdemarchi:
Oh... the solution is actually neat. At the first time you've seen the sysid/compid in one link you just add it to the linked groups. So during route the normal routing rules apply. LGTM

lucasdemarchi:
Closing in favor of #338

PR #331, opened 2022-01-20
CI: Upload cross-compiled artifacts as well

jbeyerstedt:
Currently only the result of the alpine linux 64 bit build is uploaded as a build artifact, in action runs as well as releases. Especially for the releases, I think we should provide the cross-compiled builds as well.  
(Unfortunately it doesn't solve cross-compiling for my board since my yocto linux glibc is too old. But I'm not sure, if we want to support legacy systems as well.)

This PR also fixes an issue with the version number that's automatically included in the build and displayed in the log messages during startup. Since the git tags are used for this version number, a shallow clone as it's done by default just gets us the commit hash and no "nice" version number.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
             arch: arm-linux-gnueabihf
           - name: aarch64
             arch: aarch64-linux-gnu
-          - name: native
+          - name: x86_64
             arch: native
           - name: i686
             arch: native
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+          fetch-depth: 0  # otherwise the version header just contains the commit hash
 
       - name: install dependencies (native)
         if: ${{ matrix.arch == 'native' }}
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Are these uploads useful at all? We are building dynamic binaries and uploading them which doesn't feel good. I think we should either get it packaged for/by some distros or use a better approach like a appimage, minimized docker container, flatpak etc

jbeyerstedt:
You're probably right, that proper packaging is the better way, if we want to provide binaries. I just saw, that one binary is uploaded to the workflow result (which will only be kept 90 days or so) and also tagged releases (indefinitely).

lucasdemarchi:
It was @patrickelectric  who added that I think. @patrickelectric  are you using these binaries? Are they useful?

patrickelectric:
Yep, they are pretty hand for anyone that depends of specific versions and wants to download is easily. 
+1 also for the static musl binaries.

lucasdemarchi:
well... I guess we can merge this and when we have something else that is better, we can replace it.

PR #332, opened 2022-01-21
Add support for unix domain sockets to TCP endpoint

jbeyerstedt:
Quick stab at implementing unix domain socket support. I deliberately wanted to use a `SOCK_STREAM` instead of `SOCK_DGRAM` to have bi-directional communication from the start on. (Somehow it didn't work with datagram first.)

I extended the TCP endpoint because I wanted to have the re-connect logic. Somehow it sending didn't work using `sendto` because of "Transport endpoint is already connected".

The code is not great, but may be a starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,9 +184,11 @@ config file format):
       incoming message was received.
   - TCP Client:
     * Configuration: Target IP address and port, reconnection interval in case
-      of disconnection
-    * Behavior: Data is received and sent right after the TCP session is
-      established
+      of disconnection  
+      Alternately a unix domain socket path can be used as address, the port
+      will be ignored in that case.
+    * Behavior: Data is received and sent right after the TCP (or unix domain
+      socket) session is established 
 
 Defining endpoints:
 
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -156,6 +156,7 @@ Port = 11000
 
 # Server IP address to connect to.
 # IPv6 addresses must be encosed in square brackets like `[::1]`.
+# Unix domain sockets paths must begin at root (start wtih a `/`).
 # Mandatory, no default value
 #Address = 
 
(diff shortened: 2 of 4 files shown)


lucasdemarchi:
Why adding it as client? Wouldn't it make more sense as server?

jbeyerstedt:
> Why adding it as client? Wouldn't it make more sense as server?

You mean having it the same as TcpServerPort? In general, yes. **But**, I need to use endpoint groups (PR #330) with these endpoints in my use case. And that doesn't work with the server on mavlink-router's side.

lucasdemarchi:
Sorry, I think this is too much if/else to consider extending TcpEndpoint. We could move the re-connection logic to a helper class and make both TcpEndpoint and UnixEndpoint to use it

lucasdemarchi:
I think we should leave this for v4. We don't have to take a long time to release it after v3

jbeyerstedt:
Definitely don't wait for this PR with the release of v3. I'm not sure though, what the best versioning scheme and release cycle would be, but that's better discussed in #344.

jbeyerstedt:
I'm sorry to say this, but I don't have any time in the foreseeable future to rework this. We changed the software, which needed the UNIX domain socket support in the first place to standard TCP server because of several other reasons. So there is no urge to get this done from our side any more, unfortunately.

thealienthing:
Several years ago before I knew about mavlino router, I wrote my own router which utilized Unix sockets for coms. If it's still a welcome feature I can take a look at what I had and see if I can can bring it in as a feature. I believe I used a single unix socket as a server node which would send a string notifying the client of the file descriptor it should use for its own private connection. I'll take a look and see if I can dig it up.

lucasdemarchi:
Using FD-passing? Yeah, I think that would ok.

PR #333, opened 2022-01-28
README: Write down some guidelines

lucasdemarchi:
These guidelines were adapted from https://github.com/rostedt/libccli.
For people like me who generally work in the Linux kernel, it's hard to
apply some practices common to other projects. This is not as strict as
rules to contribute to Linux kernel, but at least give a hint on how
to make it easier for the PR to be reviewed and applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,16 +252,30 @@ is an increasing number.
 
 ## Contributing
 
-Pull-requests are accepted on GitHub.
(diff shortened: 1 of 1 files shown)


jbeyerstedt:
Clear and concise, I like it. It's actually how I want contributions to my project to be. (Even though we sometimes have different thoughts on how small the steps should be. Or I was just too lazy to split things up.)

PR #335, opened 2022-01-28
conf: Change OptionsTable to use a sentinel

lucasdemarchi:
Instead of always having to update the size of array, use a sentinel
item at the end.


jbeyerstedt:
Cool, nice solution.

lucasdemarchi:
yeah... using a sentinel is what I usually do in C.  I was checking if we could replace it by an std::vector, but we don't have a constexpr constructor in std::vector until c++20.

PR #336, opened 2022-01-29
TCP endpoint: Reject message when not connected

jbeyerstedt:
Just a small fix of the error messages to be less confusing. Otherwise you'll get "Trying to write invalid fd" on a TcpEndpoint which was connected once, but then lost it's socket connection.

Steps to reproduce:
1. Add a TcpEndpoint to the config (e.g. to another mavlink-router) and a connection to a FMU (to generate telemetry data)
2. Start mavlink-router while the server to connect to is also started
3. Stop the server
4. See many "Trying to write invalid fd" errors until the server is started again

I suppressed the error message in `TcpEndpoint::write_msg()` first, but then thought, that rejecting the message in `Endpoint::accept_msg()` is actually nicer even though it introduces a very special condition to the previously generic `accept_msg` method.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -1525,6 +1525,17 @@ int TcpEndpoint::write_msg(const struct buffer *pbuf)
     return r;
 }
 
+Endpoint::AcceptState TcpEndpoint::accept_msg(const struct buffer *pbuf) const
+{
+    // reject when TCP endpoint is not connected (but trying to re-connect)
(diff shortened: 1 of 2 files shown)


lucasdemarchi:
Any reason for not making TcpEndpoint override the method and call the parent after the additional check?

jbeyerstedt:
Other than being too lazy to change it after the committed quick fix, no. I'll re-do that and force-push the proper version.

PR #337, opened 2022-01-29
Sniffer function added

grodnay:
Hello Maintainer of mavlink-router!

I added a sniffer function to mavllink-router.

The user can choose a Mavlink system-id that all messages are forwarded to. This is useful for real time debugging. One can choose his favorite mavlink inspector (Missionplanner and QGroundcontrol have nice ones...) and observe all Mavlink traffic.

You can leave this option on permanently with sys-id=254 for example, and whenever you wand to see what's going on just change you GCS id to 254 or connect with another GCS with this id.

I found this useful, so I added command line and configuration file functionality, so everyone can use it. 
Remark: Even on sniffer endpoint filters apply and messages are not echoed back.

Cheers,
-Guy

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -22,7 +22,8 @@
 ## General Configuration
 ##
 
-[General] # this section only has one instance and therefore no name
+[General]
+# this section only has one instance and therefore no name
 
 # Print traffic statistics to stdout
 # Default: <false>
@@ -78,6 +79,12 @@
 # disable this functionality.
 # Default: 0 (disabled)
 #MaxLogFiles = 0
+
+# SnifferSysid
+# Forward all traffic to endpoints on which this MAVLink system ID is connected.
+# This can be used to log or view all messages flowing though mavlink-router.
+# Default: 0 (disabled)
+#SnifferSysid=254
 
 
 ##
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -52,6 +52,8 @@
 
 #define UART_BAUD_RETRY_SEC 5
 
+uint16_t Endpoint::sniffer_sysid = 0;
+
 // clang-format off
 const char *UartEndpoint::section_pattern = "uartendpoint *";
 const ConfFile::OptionsTable UartEndpoint::option_table[] = {
@@ -421,6 +423,11 @@ void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
         return;
     }
 
+    if ((sniffer_sysid != 0) && ((sys_comp_id >> 8) == sniffer_sysid)) {
+        log_info("Sniffer sysid %u identified. [%d] is now sniffing all messages",
+                 sniffer_sysid,
+                 fd);
+    }
     _sys_comp_ids.push_back(sys_comp_id);
 }
 
@@ -495,6 +502,10 @@ Endpoint::AcceptState Endpoint::accept_msg(const struct buffer *pbuf) const
     // accept
     if ((pbuf->curr.target_compid == 0 || pbuf->curr.target_compid == -1)
         && has_sys_id(pbuf->curr.target_sysid)) {
+        return Endpoint::AcceptState::Accepted;
+    }
+    // This endpoint has the sniffer_sysid: accept
+    if ((sniffer_sysid != 0) && has_sys_id(sniffer_sysid)) {
         return Endpoint::AcceptState::Accepted;
     }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -51,9 +51,10 @@ static const struct option long_options[] = {{"endpoints", required_argument, nu
                                              {"debug-log-level", required_argument, nullptr, 'g'},
                                              {"verbose", no_argument, nullptr, 'v'},
                                              {"version", no_argument, nullptr, 'V'},
+                                             {"sniffer-sysid", required_argument, nullptr, 's'},
                                              {}};
 
-static const char *short_options = "he:rt:c:d:l:p:g:vV";
+static const char *short_options = "he:rt:c:d:l:p:g:vV:s:";
 
 static void help(FILE *fp)
 {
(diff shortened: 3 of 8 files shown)


lucasdemarchi:
That's an interesting idea. It's more or less what the "eavesdropping" functionality was supposed to be but never was. The difference here is that you reserve a system_id for it instead of attaching the functionality to a certain endpoint, set on the configuration file.

It seems this makes it easier to use as you can just connect e.g. a "mavlink inspector" or "mavlink logger" via TCP instead of explicitly dedicating an endpoint for it. @jbeyerstedt any thoughts on this?

@grodnay please take a look in the coding style issues pointed out by CI

jbeyerstedt:
On the general concept: I can see a use case for both, ignoring the normal routing rules for a specific ground control station (as it's proposed here) as well as a dedicated endpoint getting a copy of all data flowing though mavlink-router (eavesdropping). The opportunity to use a connection to the TCP server is definitely a plus for this variant. Your analysis application just has to send a heartbeat at least once and you'll effectively have the same functionality as an eavesdropping mode on any endpoint.
tl;dr: This is more flexible than an eavesdropping mode and can even do the same with a minor tweak to your data sink.

On the code itself: As Lukas already said, please check the coding style guide and your spelling in the comments. And please add a check for the value range since MAVLink system IDs are 8 bit unsigned, but the input parsing accepts unsigned long (for the config file) or signed int (for the CLI).

src/main.cpp
```diff
@@ -81,6 +82,7 @@ static void help(FILE *fp)
         "                               <error|warning|info|debug>\n"
         "  -v --verbose                 Verbose. Same as --debug-log-level=debug\n"
         "  -V --version                 Show version\n"
+        "  -s --sniffer-sysid           Sysid that all messages are sent to. 0 to disable\n"
```
jbeyerstedt, line 85: "0 to disable" doesn't match your check in line 246 ff.

examples/config.sample
```diff
@@ -73,6 +74,12 @@
 # Default: 0 (disabled)
 #MaxLogFiles = 0
 
+# SnifferSysid
+# Sysid of sniffer. All traffic is forwarded to endpoints with this sys id.
+# Useful for debug porposes with a mavlink inspector on this endpont.
+#  Default: 0 (disabled)
```
jbeyerstedt, line 80: Please use the same style as the other descriptions. Like: "Forward all traffic to endpoints on which this MAVLink system ID is connected. This can be used to log or view all messages flowing though mavlink-router.
Default: 0 (disabled)"

src/main.cpp
```diff
@@ -593,6 +606,11 @@ int main(int argc, char *argv[])
 
     Log::set_max_level(config.debug_log_level);
 
+    Endpoint::sniffer_sysid = config.sniffer_sysid;
```
jbeyerstedt, line 609: I think, thing should better be at `Mainloop::add_endpoints()` since the main.cpp is only doing config parsing and not dealing with the endpoints themselves.
Also only change the endpoint variable when `config.sniffer_sysid != 0` since it was initialized with 0 (similar to [here](https://github.com/mavlink-router/mavlink-router/blob/fa0f260a3ad6372c5f8c0f212a0295e11b58d3ff/src/mainloop.cpp#L421))

jbeyerstedt:
@lucasdemarchi I added comments instead of opening a review because I'll be on vacation the next two weeks and don't know whether you can resolve my review or if I need to do this before merging.

src/main.cpp
grodnay: Added range check and fixed message.

examples/config.sample
grodnay: done

src/endpoint.cpp
```diff
@@ -491,6 +495,9 @@ Endpoint::AcceptState Endpoint::accept_msg(const struct buffer *pbuf) const
         && has_sys_id(pbuf->curr.target_sysid)) {
         return Endpoint::AcceptState::Accepted;
     }
+    // This endpoint has the sniffer_sysid: accept
+    if ((sniffer_sysid!=0)&&has_sys_id(sniffer_sysid)) 
+        return Endpoint::AcceptState::Accepted;
```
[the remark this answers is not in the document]
grodnay: done

src/main.cpp
grodnay: done

grodnay:
fixed according to comments and run clang-format.

src/endpoint.cpp
```diff
@@ -414,6 +416,10 @@ void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
     if (has_sys_comp_id(sys_comp_id)) {
         return;
     }
+    if ((sniffer_sysid != 0) && ((sys_comp_id >> 8) == sniffer_sysid))
```
lucasdemarchi, line 419:
```suggestion
    if (sniffer_sysid != 0 && sysid == sniffer_sysid)) {
```

Humn... I noticed you run clang-format, but not sure why it left he if without braces here pass. @jbeyerstedt ?

src/endpoint.cpp
```diff
@@ -414,6 +416,10 @@ void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
...
+        log_info("Sniffer sysid %u identified. [%d] is now sniffing all messages",
+                 sniffer_sysid,
+                 fd);
```
lucasdemarchi, line 422: not sure the _add_sys_comp_id() would be the most appropriate place for this message, but we can move it later.

lucasdemarchi (commented):
LGTM. I left some comments. It's also practice in this repo not to have fixup commits on top. Could you can squash everything and check my other comments here? Thanks.

src/endpoint.cpp
grodnay: 1. Did the squash. 
2. Fixed the braces. BTW, first time I am using clang. Installation that works with the build on Ubuntu 20 not trivial (you need version 13+ and two sym links...). Would you like me to add some words in the Readme in a different pull request?
3. The reason I chose this place for the message: In order that endpoint become a sniffer you just have to add the sniffer id to it. No other action is need. So, once we do that I notified the user so he's aware.

lucasdemarchi: Sounds good to me. Thanks for working on this feature

lucasdemarchi:
@grodnay there were some conflicts due to the message duplication feature being merged. I fixed the conflicts and pushed to your sniffer  branch so this PR is updated. After CI is complete we can merge it.

src/endpoint.cpp
grodnay: My pleasure

grodnay:
Excellent! :-)

PR #339, opened 2022-02-16
UART Endpoint: Speeds never returned success

jbeyerstedt:
add_speeds and the change_baud_cb never set the return value correctly if the tried baud rate was successful.

We had some UART issues and wondered, how a `add_speeds()` has ever worked and not got stuck in an endless loop of callbacks. This seems way more logical (correct me, if I'm wrong).

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -98,9 +98,11 @@
 # Mandatory, no default value
 #Device = 
 
-# Baudrate of the device
+# List of baudrates to try to connect with the device. A default baud rate of
+# 115200 baud will always be added to the end of the list.
+# Format: Comma separated list of integers
 # Default: 115200
-#Baud = 115200
+#Baud = 57600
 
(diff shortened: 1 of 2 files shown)


jbeyerstedt:
OK, didn't look far enough. I wondered how set_speed could detect if it's the right baud rate. So I'll degrade this PR to just fix the documentation and leave the rest as it was.

jbeyerstedt:
Closing because the underlying issue doesn't exist (in this way).

lucasdemarchi:
> OK, didn't look far enough. I wondered how set_speed could detect if it's the right baud rate. So I'll close this and open another PR for the small documentation fix. I also had a question on your de-duplication comment, but let's rather discuss this in the other PR.

there isn't a proper way... the only thing we can do here is: once we receive something and can understand it, mark it as the right baud.

PR #340, opened 2022-02-16
TCP endpoint: Fix invalid write errors during reconnect

jbeyerstedt:
When a TCP endpoint detects an error, it schedules a reconnect (if not disabled). During the reconnection, the fd is invalid, but the endpoint still exists, so the write_msg() method is still called. But write_msg gives an error when the fd is invalid, even though this is expected as long as the reconnect attempts are running.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -1522,8 +1522,8 @@ int TcpEndpoint::write_msg(const struct buffer *pbuf)
     socklen_t addrlen;
 
     if (fd < 0) {
-        log_error("TCP %s: Trying to write invalid fd", _name.c_str());
-        return -EINVAL;
+        // skip this endpoint if not connected (e.g. during reconnect)
+        return 0;
     }
 
     /* TODO: send any pending data */


src/endpoint.cpp
```diff
@@ -1521,6 +1524,10 @@ int TcpEndpoint::write_msg(const struct buffer *pbuf)
     struct sockaddr *sock;
     socklen_t addrlen;
 
+    if (!_connected) {
+        return 0; // skip this endpoint, if not connected (e.g. during reconnect)
```
lucasdemarchi, line 1528: please add comments on its own line rather than inline like this.

lucasdemarchi: Also, instead of adding a new "connected" flag, can't we just do handle the check below as a valid case? I.e.:

```
    if (fd < 0) {
        // skip this endpoint if not connected (e.g. during reconnect)
        return 0;
    }
```

jbeyerstedt: I wasn't sure, if the `fd < 0` check and the return value is needed in some other condition. But only `EPIPE` is handled specially, so you're probably right with the simplification.

Issue #88, opened 2017-05-10
Move configuration parse to each class


lucasdemarchi:
The point of heaving ConfFile::OptionsTable was exactly to allow each class to have its own section with options.

We should move things out of `parse_confs()` into an specific class. Example:

```c++
    static const ConfFile::OptionsTable option_table_uart[] = {                                                        
        {"baud",    false,  ConfFile::parse_ul,         OPTIONS_TABLE_STRUCT_FIELD(option_uart, baud)},                
        {"device",  true,   ConfFile::parse_str_dup,    OPTIONS_TABLE_STRUCT_FIELD(option_uart, device)},              
    };      
```

This should be in UartEndpoint

... and so on.

jbeyerstedt:
This sounds like a good idea at first, but when thinking about an implementation, it didn't look so good to me any more. Since we have two sources of endpoint configuration (config file and CLI parameters), the endpoint would need to take care of all those variants.

I would rather like to have _one_ interface to give the endpoint it's configuration, which is independent from the actual source of configuration. So assembling the "merged" configuration is the responsibility of one part of the software and "executing" the config is the sole purpose of the rest. This split will also make testing easier.

The options tables themselves could be moved to the respective endpoints, though. Because in general, you're right, that the actual configuration parameters are the responsibility of the endpoint classes.

Because I started to dig though the setup code with the intention to also fix this issue. But with [the current state]https://github.com/jbeyerstedt/mavlink-router/tree/feature/configparsing-rework, I thinks it's quite usable again (PR will follow as soon as my cleanup PR is merged).

lucasdemarchi:
> The options tables themselves could be moved to the respective endpoints, though. Because in general, you're right, that the actual configuration parameters are the responsibility of the endpoint classes.

this is about moving the options tables ^. Idea is that the generic code matches a specific section with a specific class, gets the table from that specific class and does the assembly with command line.

jbeyerstedt:
Thanks, this makes more sense. I already have the log endpoint in a separate table as well, so this should be a quick addition to my upcoming PR.

PR #341, opened 2022-02-16
Small documentation fixes

jbeyerstedt:
Update the documentation to match the implemented behavior.

1. UART baud rates can be a list
2. Recommended values for de-duplication and a note about potential drawbacks

TODO: Fill in dedup max value.


examples/config.sample
```diff
@@ -98,9 +98,11 @@
 # Mandatory, no default value
 #Device = 
 
-# Baudrate of the device
+# List of baudrates to try to connect with the device. A default baud rate of
+# 115200 baud will always be added to the end of the list.
```
lucasdemarchi, line 104: humn... where is this coming from? If user configured the baud as 57600 we should never try to switch to 115200

lucasdemarchi: It looks like a regression caused by 46ce50943667 ("Config: Move option tables to endpoint classes (fixes #88)"). We should only append the default baud rate if none is set.

jbeyerstedt: It was in the original `parse_baudrates` function, which I only moved to the class and haven't changed (see https://github.com/mavlink-router/mavlink-router/commit/46ce50943667b762b4b09a6acbe5f8cbc7e326bb#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2L535 (in main.cpp)). But you're right, that it doesn't make sense to append it to the list, if it was explicitly configured by the user.

lucasdemarchi: yeah... it was some other commit then. I'm pretty sure this worked as intended before as we used that on the Aero drone. Looking at my change in 2f62ec7bfe3d220240b42b2a3d920d3fd684a53b, that one looks correct. So it was probably something in between.

jbeyerstedt: I'll change that in this PR shortly. So we'll have default value or list as configured.

jbeyerstedt:
The UART baud rate config does not add the default value now unless the config is empty.

jbeyerstedt:
With standard PX4 telemetry settings in grounded state (about 30 messages per second overall), I got the heartbeats to drop out after 5-10 minutes with a deduplication period of 1000 seconds. 100 seconds still worked fine after 20 minutes. I think these two tests should be fine to give us a ballpark figure.

I recommended anything below 1000 ms in the readme and sample config though. This should be safe for 1 Hz heartbeats even in the case that exactly 254 other messages were sent between two heartbeats. I reality, even a few seconds should be fine.

lucasdemarchi:
Thanks, that looks good.

PR #355, opened 2022-03-30
ci: Add misssing update

lucasdemarchi:
Update package database before trying to install.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -1,6 +1,10 @@
 name: clang-format Check
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 
 jobs:
   formatting-check:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,12 +29,15 @@ jobs:
               sudo dpkg --add-architecture i386
               sudo apt-get update
               sudo apt-get install --yes libgtest-dev:i386
+          else
+              sudo apt-get update
           fi
           sudo apt-get install --yes pkg-config gcc-multilib g++-multilib ninja-build python3-pip libgtest-dev
           pip3 install meson pymavlink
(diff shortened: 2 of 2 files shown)


jbeyerstedt:
While you're at the CI pipeline: Do we also want to add the tag name to the build artifacts uploaded to the release like this:
```patch
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 8708731..f6c9751 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,7 +84,7 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: build-${{ matrix.name }}/src/mavlink-routerd
-          asset_name: mavlink-routerd-glibc-${{ matrix.name }}
+          asset_name: mavlink-routerd-glibc-${{ matrix.name }}-$tag
           tag: ${{ github.ref }}
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
           overwrite: true
@@ -121,6 +121,7 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: build/src/mavlink-routerd-musl-x86-64
+          asset_name: mavlink-routerd-musl-x86-64-$tag
           tag: ${{ github.ref }}
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
           overwrite: true
```

Otherwise: LGTM

lucasdemarchi:
I don't really like to have different names. I'd rather provide a proper per-distro .deb/.rpm package and hopefully integrate in a few package managers.

At the very least this should be better planned when we are doing the next release.

lucasdemarchi:
Thanks for the review. I'm going ahead and merging this to unblock CI.

jbeyerstedt:
It was just an idea and I understand the issue with changing file names (as I have some ansible roles, which try to download binaries from GitHub releases automatically). So let's keep it as it is.

PR #370, opened 2022-08-02
Improve CI builds

patrickelectric:
image


.github/workflows/main.yml
```diff
@@ -6,8 +6,20 @@ on:
...
+      - uses: actions/checkout@v2
+      - name: Run clang-format style check on src directory
+        uses: jidicula/clang-format-action@v4.3.0
+        with:
+          clang-format-version: "13"
+          check-path: "src"
+
   build:
     runs-on: ubuntu-20.04
+    needs: [formatting-check]
```
lucasdemarchi, line 22: why? They are 2 independent things. Also if we want to move formatting-check to main.yml, then you need to remove that file in the same commit

.github/workflows/main.yml
```diff
@@ -2,7 +2,6 @@ name: Build and Test
 
 on:
   push:
-    branches: [master]
```
lucasdemarchi, line 5: ok, but AFAIR we had the following problem:  when submitting a PR I usually push a tip-<feature>  branch to this repo. I don't want to run CI on both the branch and the PR.  Any other way to "Allow creation of binaries for tags" without removing this ?

.github/workflows/main.yml
khancyr: I think the idea is :
Formating is mandatory so if formating check failed, the pr will need update so that isn't worse it to do full check for something that will need correction anyway.

.github/workflows/main.yml
khancyr: We can keep both push and pull_request
And maybe add workflow_dispatch to allow manual run

lucasdemarchi: problem here is that it's removing the filter "branch == master", so it runs on any branch we push, even those that are pushed just to create a PR

.github/workflows/main.yml
lucasdemarchi: CI will still fail and PR will need to be corrected.
I believe it's important to know if code builds and runs regardless of any coding format issue (that can be easily fixed up later)

.github/workflows/main.yml
patrickelectric: > even those that are pushed just to create a PR

As a general development flow, PR's should be done from the developer fork, and not from the main repository, this will turn out in a polluted environment for others developers. If it's a branch that others developers will contribute, makes sense for the CI to run.

.github/workflows/main.yml
patrickelectric: Besides @khancyr comment, software style is usually mandatory to be fixed, the check runs in 20 seconds and limit the build resource (the github actions machines) to be free and build others PRs.

.github/workflows/main.yml
patrickelectric: this also empowers people that are doing their development on their own fork, with that, they are able to have CI running own their custom branches to checkout the binaries and CI. Otherwise, the only way is to creating a PR here.

lucasdemarchi: I'm not against running it on branches, only against running it on anything other than the master branch.

well... I don't really care much, so if it's useful for people, ok.

PR #385, opened 2023-02-06
Ignore comp id fix

kaangoksal:
This PR adds option to route ignore component ids for a system.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -67,7 +67,7 @@ const ConfFile::OptionsTable UartEndpoint::option_table[] = {
     {"AllowSrcCompIn",  false, ConfFile::parse_uint8_vector,    OPTIONS_TABLE_STRUCT_FIELD(UartEndpointConfig, allow_src_comp_in)},
     {"AllowSrcSysIn",   false, ConfFile::parse_uint8_vector,    OPTIONS_TABLE_STRUCT_FIELD(UartEndpointConfig, allow_src_sys_in)},
     {"group",           false, ConfFile::parse_stdstring,       OPTIONS_TABLE_STRUCT_FIELD(UartEndpointConfig, group)},
-    {"IgnoreCompId",    false, ConfFile::parse_bool,            OPTIONS_TABLE_STRUCT_FIELD(UartEndpointConfig, ignore_comp_id)},
+    {"IgnoreCompId",    false, ConfFile::parse_bool,            OPTIONS_TABLE_STRUCT_FIELD(UartEndpointConfig, ignorecompid)},
     {}
 };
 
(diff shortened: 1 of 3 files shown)


src/endpoint.cpp
```diff
@@ -706,6 +714,10 @@ bool UartEndpoint::setup(UartEndpointConfig conf)
         }
     }
 
+    if (conf.ignorecompid) {
```
jbeyerstedt, line 717: This if-statement isn't needed as you have a default value in the config struct, so the value always exists. (Applies to all three occurrences.)

src/endpoint.h
```diff
@@ -48,6 +48,7 @@ struct UartEndpointConfig {
     std::vector<uint8_t> allow_src_comp_in;
     std::vector<uint8_t> allow_src_sys_in;
     std::string group;
+    bool ignorecompid{false};
```
jbeyerstedt, line 51: Please stick to the snake_case naming style (applies to all three occurrences).

src/mainloop.cpp
```diff
@@ -186,7 +186,13 @@ void Mainloop::route_msg(struct buffer *buf)
...
             break;
         case Endpoint::AcceptState::Rejected:
-            // fall through
+            log_debug("Endpoint [%d] rejected message %u to %d/%d from %u/%u",
+                      e->fd,
+                      buf->curr.msg_id,
+                      buf->curr.target_sysid,
+                      buf->curr.target_compid,
+                      buf->curr.src_sysid,
+                      buf->curr.src_compid);
```
jbeyerstedt, line 195: Please add a `break;` statement after this log line as this is a full case and not just an intentional fall-through now.

src/endpoint.cpp
```diff
@@ -453,6 +456,7 @@ void Endpoint::_add_sys_comp_id(uint8_t sysid, uint8_t compid)
                  sniffer_sysid,
                  fd);
     }
+    log_info("%s Adding sysid: %d compid: %d", _name.c_str() ,sysid, compid);
```
jbeyerstedt, line 459: Is this log line really needed? You'll get the list of connected components on the endpoint for the next routed message anyways.
If you think the line is useful, please include the endpoint ID (fd) in the log line like: "Endpoint [%d]%s: Adding %d/%d to known components"

src/endpoint.cpp
```diff
@@ -540,6 +544,10 @@ Endpoint::AcceptState Endpoint::accept_msg(const struct buffer *pbuf) const
         && has_sys_id(pbuf->curr.target_sysid)) {
         return Endpoint::AcceptState::Accepted;
     }
+
+    if(ignore_comp_id && has_sys_id(pbuf->curr.target_sysid)) {
```
jbeyerstedt, line 548: Please add a comment what this does and add a blank line before the next one (like the other if statements). E.g. "If comp ID should be ignored and this endpoint has the target of message (sysid only): accept"

jbeyerstedt (changes requested):
I'm not 100% convinced by the name of the parameter, but have no better idea right now either.

Please add the new parameter to the example config with a short explanation as well as the [message routing rules](https://github.com/mavlink-router/mavlink-router#message-routing) in the readme.

kaangoksal:
okay, will fix it. Thanks for your time.

lucasdemarchi:
@kaangoksal any updates here?

kaangoksal:
Sorry for the delay, will fix my pr this week!

src/endpoint.cpp
kaangoksal: Added the requested description

src/mainloop.cpp
kaangoksal: added the break statement

src/endpoint.cpp
kaangoksal: removed it.

src/endpoint.h
kaangoksal: its now snake case!

src/endpoint.cpp
kaangoksal: fixed

kaangoksal:
I'm testing the latest updates. not ready yet.

PR #387, opened 2023-02-25
Replace deprecated get_pkgconfig_variable

holden-zenith:
[get_pkgconfig_variable](https://mesonbuild.com/Reference-manual_returned_dep.html#depget_pkgconfig_variable) => [get_variable](https://mesonbuild.com/Reference-manual_returned_dep.html#depget_variable)


lucasdemarchi:
LGTM, thanks

jbeyerstedt (approved):
We depend of meson >= 0.57 (according to the Readme) anyways and the old option was deprecated in 0.56. So this actually seems like the right way to do it.

PR #391, opened 2023-04-04
Fix environment variable documentation

lopsided98:
The readme gives the wrong names for the configuration file environment variables. They are prefixed with `MAVLINK_ROUTERD_` not `MAVLINK_ROUTER_`.


jbeyerstedt (approved):
You're right. Thanks for the PR.

PR #395, opened 2023-05-30
add cross build for mips(for openwrt)

Vkd4U2NtVkZOVg:
tested on MT7688 works fine.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -30,7 +30,7 @@
 #ReportStats = false
 
 # Logging verbosity (stderr)
-# Valid values: <error>, <warning>, <info> or <debug>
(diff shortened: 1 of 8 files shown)


jbeyerstedt:
Please rebase to let the CI run again.

meson-cross-mipsel.ini
```diff
@@ -0,0 +1,15 @@
+[constants]
+arch = 'mipsel'
+
+[binaries]
+c = arch + '-openwrt-linux-gcc'
```
lucasdemarchi: where do we get that cross-compiler from? Can you add it to our CI so we don't regress?

PR #402, opened 2023-08-05
Filters: Add blocker filters

OhadNir9:
Like the filters that allow messages with specific parameters (msg_id, sys_id and comp_id) to pass as outgoing/incoming messages in/to endpoints, the new filters enabling blockage of messages with such parameters.
This feature co-exists with the current "allow" filters, with the only exception that a message that appears in both filter types (allow and block) will eventually be **blocked**.

Unit-tested, but not yet tested in real setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,11 @@ Message filters:
   - BlockSrcSysIn: If set, block messages from the listed MAVLink source
     systems to be received on this endpoint
 
+    Note that while using "Allow" and "Block" filters on the same identifier 
+    within an endpoint doesn't make sense, using them on different identifiers 
+    can be useful (for example, allowing only specific outgoing SysID, and
+    blocking this system from sending some unwanted message IDs).
+
 Message de-duplication:
 
   - If enabled, each incoming message is checked, whether another copy was
--- a/README.md
+++ b/README.md
@@ -231,31 +231,26 @@ Message filters:
 
   - AllowMsgIdOut: If set, only allow messages with the listed message IDs to
     be sent via this endpoint
-  - BlockMsgIdOut: If set, block messages with the listed message IDs to
-    be sent via this endpoint
-
+  - BlockMsgIdOut: If set, block messages with the listed message IDs to be sent
+    via this endpoint
   - AllowSrcCompOut: If set, only allow messages from the listed MAVLink source
     component IDs to be sent via this endpoint
   - BlockSrcCompOut: If set, block messages from the listed MAVLink source
     component IDs to be sent via this endpoint
-  
   - AllowSrcSysOut: If set, only allow messages from the listed MAVLink source
     systems to be sent via this endpoint
   - BlockSrcSysOut: If set, block messages from the listed MAVLink source
     systems to be sent via this endpoint
-  
   - AllowMsgIdIn: If set, only allow messages with the listed message IDs to
     be received on this endpoint. Since message ID 0 is not used, only allowing
     this message ID can be used to block all incoming traffic on this endpoint,
     e.g. to block interaction with the drone while still receiving telemetry.
   - BlockMsgIdIn: If set, block messages with the listed message IDs to
-    be received on this endpoint. 
-  
+    be received on this endpoint
   - AllowSrcCompIn: If set, only allow messages from the listed MAVLink source
     component IDs to be received on this endpoint
   - BlockSrcCompIn: If set, block messages from the listed MAVLink source
     component IDs to be received on this endpoint
-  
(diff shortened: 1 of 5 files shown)


src/endpoint.cpp
```diff
@@ -709,22 +763,41 @@ bool UartEndpoint::setup(UartEndpointConfig conf)
...
+    for (auto src_comp : conf.block_src_comp_out) {
+        this->filter_add_blocked_out_src_comp(src_comp);
+    }
     for (auto src_sys : conf.allow_src_sys_out) {
         this->filter_add_allowed_out_src_sys(src_sys);
     }
+    for (auto src_sys : conf.block_src_sys_out) {
+        this->filter_add_blocked_out_src_sys(src_sys);
+    }
+
```
jbeyerstedt, line 781: Again: Why the extra newline? There already was one to separate the two "blocks" of for each loops.

src/endpoint.cpp
```diff
@@ -1379,23 +1470,40 @@ bool TcpEndpoint::setup(TcpEndpointConfig conf)
...
+    for (auto src_comp : conf.block_src_comp_in) {
+        this->filter_add_blocked_in_src_comp(src_comp);
+    }
     for (auto src_sys : conf.allow_src_sys_in) {
         this->filter_add_allowed_in_src_sys(src_sys);
     }
-
+    for (auto src_sys : conf.block_src_sys_in) {
+        this->filter_add_blocked_in_src_sys(src_sys);
+    }
```
jbeyerstedt, line 1506: And why was the newline before `this->_group_name ...` removed?

src/endpoint.h
```diff
@@ -165,26 +183,51 @@ class Endpoint : public Pollable {
...
+    }
     void filter_add_allowed_in_src_sys(uint8_t src_sys)
     {
         _allowed_incoming_src_systems.push_back(src_sys);
     }
+    void filter_add_blocked_in_src_sys(uint8_t src_sys)
+    {
+        _blocked_incoming_src_systems.push_back(src_sys);
+    }
+
```
jbeyerstedt, line 230: Again: Why extra newline?

README.md
```diff
@@ -231,18 +231,35 @@ Message filters:
 
   - AllowMsgIdOut: If set, only allow messages with the listed message IDs to
     be sent via this endpoint
+  - BlockMsgIdOut: If set, block messages with the listed message IDs to
+    be sent via this endpoint
```
jbeyerstedt, line 235: Lines should be max. 80 characters wide, so the `be sent` shall still be on the first line.

jbeyerstedt (changes requested):
Sorry to be quite nit-picky about the formatting, but it helps keeping the code in a readable and maintainable state. In general the code seems fine though.

But could you please add some sentence to the readme, that it doesn't make any sense to specify an "allow" and a "block" filter for the same identifier type. But mixing e.g. a message ID block list with a system ID allow list is fine.

jbeyerstedt:
Out of curiosity (and reasoning why to add more features): What's your use case for this? (I only had use for explicit allow lists when implementing the full set of filters.)

src/endpoint.h
jbeyerstedt: clang-format has the same question by the way

src/endpoint.cpp
jbeyerstedt: clang-format has the same question by the way

OhadNir9:
Sorry for the formatting issues..I'll fix them soon.
My use case was MissionPlanner - appearently every time it starts up, it sends right away a REQUEST_DATA_STREAM message with some specific rates, that conflicts with the IMU data message rates I needed for another algorithm. I looked for an easy way to block this REQUEST_DATA_STREAM messages from coming out the MP endpoint

OhadNir9:
Seems like the CI is broken? @jbeyerstedt

jbeyerstedt:
> Seems like the CI is broken? @jbeyerstedt

Please rebase, CI should be fixed now.

jbeyerstedt:
The Readme section about message filters got quite long now, so I would propose to rework it like in the attached patch file:[0001-Readme-Rework-message-filter-section.patch](https://github.com/mavlink-router/mavlink-router/files/12535080/0001-Readme-Rework-message-filter-section.patch)

(I can't push to your branch, so this was the quickest solution which came to my mind. Just try to "git apply" it to your branch.)

OhadNir9:
Hi, I've applied the requested changes @jbeyerstedt

jbeyerstedt:
Squashed and merged manually

PR #409, opened 2023-10-05
Don't perform UART reset on shutdown

### LINKED ISSUE #213: Error setting serial port latency, unable to reset UART afterwards, Jetson TX2
When running the router on a NVIDIA Jetson TX2 dev kit running stock jetpack, the router throws an error stating:

"Error while trying to write serial port latency: Unknown error -1"

the command is:
mavlink-routerd /dev/ttyTHS2 -e 192.168.0.50:49411

The port does appear to be set correctly, as the router works as intended after this. I can confirm with stty; the port is at 9600 at boot, and after running mavlinkrouter the port is now at 115200

That said, if I stop and restart the router, it fails to start, saying "could not reset uart"

in this state, stty says "input/output error" when I query the port. 

It does it same behavior on all the ttyTHS# ports, even if no serial device is attached to them.

I am able to configure and view the serial data using screen as many times as desired. 

I am attaching straces from the initial run, and the second failed run. 
[strace_second.5557.txt](https://github.com/intel/mavlink-router/files/4670090/strace_second.5557.txt)
[strace_first.5543.txt](https://github.com/intel/mavlink-router/files/4670091/strace_first.5543.txt)

Any help would be greatly appreciated, and am happy to run any troubleshooting tests as might be required.

niwhsa9:
This fixes the issue exhibited in #213. I have tested this on a Jetson TX2 and confirmed that by not performing the reset, the UART port is still accessible with minicom after shutting down mavlink router. For what its worth, setting stty's sane on the UART does not cause the same problem, so there is evidently some difference between the `reset_uart`  and stty sane despite the comment in the code. 

I hate to be the guy who offers a solution that's just "delete this code and it works", but its not clear to me what the underlying issue is and this does fix the bug.  I would totally understand if you all don't feel comfortable merging something like this, but I figured it might be helpful to throw this out there just in case. 

Would like to hear thoughts from the maintainers: Is there a compelling reason to keep this logic around?

closes #213

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -727,13 +727,6 @@ UartEndpoint::UartEndpoint(std::string name)
     // nothing else to do here
 }
 
-UartEndpoint::~UartEndpoint()
-{
-    if (fd > 0) {
-        reset_uart(fd);
-    }
-}
-
 bool UartEndpoint::setup(UartEndpointConfig conf)
 {
     if (!this->validate_config(conf)) {
--- a/src/endpoint.h
+++ b/src/endpoint.h
@@ -294,7 +294,7 @@ private:
 class UartEndpoint : public Endpoint {
 public:
     UartEndpoint(std::string name);
-    ~UartEndpoint() override;
+    ~UartEndpoint() override = default;
 
     int write_msg(const struct buffer *pbuf) override;
     int flush_pending_msgs() override { return -ENOSYS; }


lucasdemarchi:
This was added in 4b719070b1784b9793bcfe5ff68281c6ec48389c because we use vmin/vtim in the UART to be notified of new data and some programs (I think I tested with screen 6 years ago) wasn't working after shutting down mavlink-router.

We could either 
a) remove it like you did if we confirm things still work now
b) just reset vmin/vtime to 1
c) see what stty does again and change reset_uart() to follow it

niwhsa9:
gotcha, I double checked sane flags in the stty man page against what `reset_uart()` sets and it seems to be a 1:1 match. I will try screen. I'd be shocked if it didn't work seeing as how minicom did. Will report back soon.

niwhsa9:
There seems to be no problems launching `screen` version 4.9.0 on the UART after killing this version of mavlink-router. Looks like there is data coming through also. I'm guessing whatever the issue was in screen has since been resolved.

lucasdemarchi:
Sounds good then. @jbeyerstedt do you see any problem with this? Maybe if the uart is shared with the script to upload a new firmware?

jbeyerstedt:
Screen definitely sets some options on the serial port when opened, I'm not sure about vmin/vtime though.

Regarding firmware update scripts: Isn't this usually done via USB? If so, just re-plugging the cable would reset the TTY port, so it shouldn't be a big issue even if this change breaks things. (But if your firmware update needs some specific flags, why doesn't it set them itself.)

I think this change is fine.

niwhsa9:
@lucasdemarchi just checking back in, do you want to merge?

lucasdemarchi:
I amended the commits with more details in the commit message and pushed here. Thanks.

PR #422, opened 2024-05-15
Add TRACE log level and syslog output option

bkueng:
This adds a TRACE log level and changes the frequently output `log_debug`s to `log_trace`, so `debug` output can be enabled without flooding the output.

Additionally, log output can be redirected to syslog via `--syslog` CLI flag. By default, stderr is used.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -30,7 +30,7 @@
 #ReportStats = false
 
 # Logging verbosity (stderr)
-# Valid values: <error>, <warning>, <info> or <debug>
+# Valid values: <error>, <warning>, <info>, <debug> or <trace>
 # Default: <info>
 #DebugLogLevel = info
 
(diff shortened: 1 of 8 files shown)


lucasdemarchi:
First 2 commits look good. Third one: I don't really see a reason - you're adding more code to handle syslog and then changing the systemd unit to use syslog, which it was already doing via stderr.

bkueng:
The reason is that you get the proper log level in the journal

lucasdemarchi:
ok... I wonder if it wouldn't be better to just follow `man 3 sd-daemon` which recommends to print the level:

```
       Example 1. Hello World

       A daemon may log with the log level NOTICE by issuing this call:

           fprintf(stderr, SD_NOTICE "Hello World!\n");
```

bkueng:
Up to you, either way works for me. Depending on the output it might be undesired to have it prefixed with `<d>`.

PR #430, opened 2024-08-19
Update mavlink_c_library_v2

dakejahl:
I was experiencing an issue where the OPEN_DRONE_ID_SYSTEM message was not being received by any of my applications. Once I updated mavlink to master, the issue was solved. I think the mavlink message definition must have changed and therefore mavlink-router knew about the message ID, but the CRC did not match, so it drops the message.


lucasdemarchi:
LGTM. @patrickelectric , ok for you?

lucasdemarchi (approved):

patrickelectric (approved):

patrickelectric:
It appears that CO os failing though

lucasdemarchi:
Fixed in #431 . Normally I push to the same PR, but this time github didn't allow me. Probably because @dakejahl 's repo is forked from PX4 org rather than mavlink-router.

lucasdemarchi:
Replaced by #431

Issue #426, opened 2024-07-05
Add tlog rotation


patrickelectric:
Tlogs can go over GBs, it would be nice to limit the maximum size of the file.

PR #432, opened 2024-09-04
Add support for tlog file splitting

goasChris:
Add configurable file size limit for tlog files. By default its about 200MB. Updated the example config file to reflect this, as well as the switches / CLI entries.

## Testing
Tried to running the binary with the new defaults, both blank and set config file, with and without CLI. Confirmed the files did in fact manage to split, and that the file is parsable in UAVLogViewer (https://plot.ardupilot.org/). Tested via SITL connecting to the mavlink router instance.

## Additional info

Opening this as a draft for now, as I am looking for input into the default, as well as a small annoyances (functionally works) with the file size check spamming complaints at the start before the first file is initialized.

Had to shuffle around some of the private/protected headers to get what I needed.

I am very happy to make any changes to this, as I am not very experienced with CPP, so this is more of a first proposal, and if someone wants to take further, I am happy to hand it off.

I've not performance tested this, and I am slightly worried about the frequency of which we check the size of the current log file, and maybe there is a way to do this more efficiently (haven't managed to think of any non-stupid ideas so far). Very open to suggestions of how it could be implemented, and I'll be happy to give it a go.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -94,6 +94,10 @@
 # Default: <false>
 # LogTelemetry = false
 
+# Max size for tlog files before they are split/rotated. Specified in bytes.
+# Valid values: <0 - 999999999999>
+# Default: 200209715200 (200MB)
+# MaxTLogFileSize = 200209715200
 
 ##
 ## UART Endpoint Configurations
--- a/src/logendpoint.h
+++ b/src/logendpoint.h
@@ -46,6 +46,7 @@ struct LogOptions {
     unsigned long max_log_files;                  // conf "MaxLogFiles"
     int fcu_id{-1};                               // conf "LogSystemId"
     bool log_telemetry{false};                    // conf "LogTelemetry"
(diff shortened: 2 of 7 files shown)


goasChris:
This was targeted to fix #426

goasChris:
Should also check one of these boxes: https://github.com/bluerobotics/BlueOS/issues/2328

goasChris:
I was unable to run the clang formatting as per specified in the README. I believe it should be fixed now tho. My apologies.

goasChris:
I've got another proposal for reducing the amount of file read operations (for the split on size check), in which I essentially just increment a counter for the number of messages we've written, and if modulo 1000, then we can check the file size (every 1000 messages). If that is sane and acceptable, I can update the PR with this code. If you would prefer to leave it, or alter that in any way, I'm more than happy to do so.

goasChris:
when the tlog is split, the full parameter list is not included in the split files. Look into if we can somehow request all params on split.

PR #441, opened 2024-11-08
Add systemd-dev package to build instructions for Ubuntu 23+

### LINKED ISSUE #440: Dependency systemd not found on Ubuntu 23+
It appears that as of Ubuntu 23 (mantic), they've split the systemd package into "systemd" and "systemd-dev". Without systemd-dev installed on newer versions of Ubuntu, meson will fail to build. Thought this might be worth updating in the build instructions.

```
pi@raspberrypi:~/mavlink-router$ lsb_release -d
Description:    Ubuntu 24.10

pi@raspberrypi:~/mavlink-router$ sudo apt install systemd
systemd is already the newest version (256.5-2ubuntu3).
Summary:
  Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0

pi@raspberrypi:~/mavlink-router$ meson setup build .
The Meson build system
Version: 1.5.1
Source dir: /home/pi/mavlink-router
Build dir: /home/pi/mavlink-router/build
Build type: native build
Project name: mavlink-router
Project version: 3
C compiler for the host machine: cc (gcc 14.2.0 "cc (Ubuntu 14.2.0-4ubuntu2) 14.2.0")
C linker for the host machine: cc ld.bfd 2.43.1
C++ compiler for the host machine: c++ (gcc 14.2.0 "c++ (Ubuntu 14.2.0-4ubuntu2) 14.2.0")
C++ linker for the host machine: c++ ld.bfd 2.43.1
Host machine cpu family: aarch64
Host machine cpu: aarch64
Library m found: YES
Library rt found: YES
Run-time dependency threads found: YES
Found pkg-config: YES (/usr/bin/pkg-config) 1.8.1
Found CMake: /usr/bin/cmake (3.30.3)
Run-time dependency systemd found: NO (tried pkgconfig and cmake)

meson.build:33:16: ERROR: Dependency "systemd" not found, tried pkgconfig and cmake

A full log can be found at /home/pi/mavlink-router/build/meson-logs/meson-log.txt
```

After an `sudo apt install systemd-dev` I was able to complete the build.

danlaks:
Please double check me, but I believe the systemd-dev package needs to be installed on Ubuntu 23 or newer. See #440.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ Debian/Ubuntu:
 
     $ sudo apt install git meson ninja-build pkg-config gcc g++ systemd
 
+If you are using Ubuntu 23 or newer, you may also need to install systemd-dev:
+
+    $ sudo apt install systemd-dev
+
 Fedora:
 
     $ sudo dnf install git meson gcc g++ systemd


jbeyerstedt:
I have two things which are not clear for me:

1. According to https://packages.ubuntu.com/noble/systemd, `systemd-dev` is a hard dependency of `systemd` and should therefore be installed automatically. So I wonder what happens when you install that explicitly again.
2. The package contents of `systemd-dev` also doesn't look like something which I expect from a meson dependency. I would rather expect headers and shared/ static libraries in the package

So I already asked in #440, if you're sure about needing `systemd-dev` instead of `libsystemd-dev`.

jbeyerstedt:
Ok, I spawned an Ubuntu VM. Ubuntu 24.04 LTS already has the `systemd-dev` package installed, but according to packages.ubuntu.com 24.10 doesn't seem to have the dependency I mentioned before.

And I guess, the meson config needs to read `/usr/share/pkgconfig/systemd.pc`, which only comes with the `systemd-dev` package.

So I guess this is actually required as mentioned in this PR. Sorry for the confusion.

README.md
```diff
@@ -29,6 +29,10 @@ Debian/Ubuntu:
 
     $ sudo apt install git meson ninja-build pkg-config gcc g++ systemd
 
+If you are using Ubuntu 23 or newer, you may also need to install systemd-dev:
```
jbeyerstedt, line 32: This will probably also affect the next Debian release (newer than 12)

Issue #373, opened 2022-08-09
crash program in uart endpoint


mohammada315:
dear mavlink-router developers
I need to send a uart stream to some other telemetry
for example
                               
```
#[UartEndpoint echo]
#Device = /dev/serial/by-id/usb-Holybro_Durandal_2C0024001151303138363434-if00
#Baud = 115200

#[UartEndpoint alpha]
#Device = /dev/serial/by-id/usb-FTDI_FT230X_Basic_UART_D306NG9H-if00-port0 
#Baud = 115200
```
But after a few moments, the program exits

mark-nick-o:
run the mavlink-routerd from the command line after stopping it with a ps -ael | grep mavlink-routerd and look at the output
if you have no usb device attached you would get something like this

[mav-sony@mav-sony ~]$ mavlink-routerd
mavlink-router version v2-235-gcf93476
Could not open /dev/ttyACM0 (No such file or directory)

lucasdemarchi:
The uart link is considered a "critical" link. If it disappears, mavlink-router considers it can't function normally anymore and exits with an error. The systemd service is supposed to restart it when it exits with an error, so if it can reset and open the uart again, it would recover. However if the device is really not available anymore, there isn't much that can be done. 

The device disappearing is usually one of:

1) being disconnected, in the case of USB
2) kernel bug - update your kernel to the latest stable release
3) hardware bugs - there may be workarounds in newer kernel versions

jbeyerstedt:
Closing because of no activity and the observed behavior was the intended behavior. Feel free to comment and re-open, if there are things left to discuss.

PR #451, opened 2025-02-08
Add an option to disable failed endpoints, without exiting

zdanek:
When USB endpoints are not available during startup of MR, it fails with fatal error and exits.
Also, when USB is disconnected during operation, it fails and exit.

IMO starting without configured endpoint might be argued, but failing during operation and exiting may lead to severe consequences.

Also, this PR is a starting point for next feature - recovery from USB failures. Mainloop could poll disabled endpoints and enable them once are back again.

PR applies to USB only, because I did not test TCP endpoints that can also fail due to TCP's nature.
Also, UDP could fail if configured to bind on non existing local IP (net changed OR cable fail thus not bringing eth itf up). This also can be supported in next PRs.

Consider my PR. I got what I wanted in my fork, but if you find this valuable, I'd be happy to contribute.

### DIFF
--- a/src/mainloop.h
+++ b/src/mainloop.h
@@ -44,6 +44,7 @@ struct Configuration {
     std::vector<UdpEndpointConfig> udp_configs;
     std::vector<TcpEndpointConfig> tcp_configs;
     unsigned long sniffer_sysid;
+    bool skip_failed_endpoints{false};
 };
 
 struct endpoint_entry {
@@ -92,7 +93,7 @@ public:
     /*
      * Initialize and return singleton.
      */
-    static Mainloop &init();
+    static Mainloop &init(const Configuration &configuration);
 
     /*
      * De-initialize singleton so we can start a fresh on the same
@@ -110,6 +111,7 @@ public:
 private:
     static const unsigned int LOG_AGGREGATE_INTERVAL_SEC = 5;
 
+    Configuration _configuration;
     std::vector<std::shared_ptr<Endpoint>> g_endpoints{};
     int g_tcp_fd = -1; ///< for TCP server
     std::shared_ptr<LogEndpoint> _log_endpoint{nullptr};
--- a/src/mainloop.h
+++ b/src/mainloop.h
@@ -93,7 +93,7 @@ public:
     /*
      * Initialize and return singleton.
      */
-    static Mainloop &init(const Configuration &configuration);
+    static Mainloop &init();
 
     /*
      * De-initialize singleton so we can start a fresh on the same
--- a/src/endpoint.h
+++ b/src/endpoint.h
@@ -234,7 +234,10 @@ public:
     void link_group_member(std::shared_ptr<Endpoint> other);
 
     std::string get_type() const { return this->_type; }
-    std::string get_group_name() const { return this->_group_name; };
+    std::string get_group_name() const { return this->_group_name; }
+    std::string get_name() const { return this->_name; }
+    bool is_disabled() const { return this->_disabled; }
+    void disabled(bool disabled) { this->_disabled = disabled; }
 
     struct buffer rx_buf;
     struct buffer tx_buf;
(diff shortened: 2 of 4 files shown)


jbeyerstedt:
Sorry to be a bit pedantic on the wording here, but there are not USB endpoints just UART -- which might be a USB-UART device, but usually is a hardware UART.
So failing on UART error is actually an expected feature as this is usually caused by a serious and unrecoverable issue (see https://github.com/mavlink-router/mavlink-router/issues/373#issuecomment-1233731765). So I would like to have some more information on the actual use case because handling UART re-connects might have additional consequences like clearing the list of connected endpoints. And dynamic re-configuration is not a feature of mavlink-router (right now).

Finally a comment on potentially adding this TCP and UDP endpoints: TCP endpoint already handle disconnects gracefully. UDP endpoint won't fail in "normal" mode and can only fail in "server" mode if the address and port binding is already in use. The latter will always happen during startup and is an unrecoverable configuration issue, so I would expect mavlink-router to fail instead of hiding this error.

lucasdemarchi:
> Sorry to be a bit pedantic on the wording here, but there are not USB endpoints just UART -- which might be a USB-UART device, but usually is a hardware UART.
> So failing on UART error is actually an expected feature as this is usually caused by a serious and unrecoverable issue (see [#373 (comment)](https://github.com/mavlink-router/mavlink-router/issues/373#issuecomment-1233731765)). So I would like to have some more information on the actual use case because handling UART re-connects might have additional consequences like clearing the list of connected endpoints. And dynamic re-configuration is not a feature of mavlink-router (right now).

Since I made that comment, let me try to clarify:  1) it was about the current state of mavlin-router; 2) we shouldn't just use that to workaround kernel issues.

That said, I think an improvement to mavlink-router to handle hotpluggable devices/endpoints would be good. That would probably be a per-endpoint config - if it's a "non-recoverable" endpoint, it exits the program. If it's not, it tries to reconnect with a timer - the tcp handling could be generalized and implemented by the individual types.

lucasdemarchi:
> When USB endpoints are not available during startup of MR, it fails with fatal error and exits.

this is usually solved by adding the device dependency to the systemd unit.

zdanek:
Ok, but you focused on my imagination about next steps. Let's discuss the value of this PR, in the current state.

I'm new to MR. Currently I just introduced it in my company as a replacement for Mavproxy which is heavy, bulky, complicated and does not serve the purpose. 

I still persist that bailing out after disconnection of USB cable with UART is a bad idea. It's better to cut off failed arm, than die completely ;-)  So it's up to you if you wish to reconnect or not. I don't have a strong opinion here, since there are a lot of pros and cons. I like to give users an option, so they can decide. Forcing solutions based only on own experience might cover only a fraction of use cases. 

So, for now - please focus on my PR. Shall I fix something? Are you interested in the feature? Or we close it.

Next PR is on the way and I need to know what are the rules here :)

jbeyerstedt:
Looked at the changes of this PR again. This is only handling unavailable UART endpoints on startup, right? Because I still don't get the use case of having "optional" UART endpoints which might be there or not. So please explain how you want to use this feature.

On the other hand I understand the need to set endpoints a hot-pluggable. But this would be a per-endpoint setting (as proposed by Lucas):

> That would probably be a per-endpoint config - if it's a "non-recoverable" endpoint, it exits the program. If it's not, it tries to reconnect with a timer

In that case, we need a config parameter for the "is_critical" property of the endpoint. This should already solve the issue of mavlink-router exiting when a poll error occurs. Automatic reconnect will probably need some refactoring (as mentioned by Lucas), but will also be a nice feature (as mentioned in the PR description).

zdanek:
@jbeyerstedt this PR adds an option to skip failed UART endpoints. When enabled with an -F option:
1) disables unavailable UART endpoints during startup - allowing MR to start 
2) disabled suddenly unavailable UART endpoints during MR operation - allowing MR to work continuously, without bailing out and suddenly stopping mavlink routing, which is dangerous during UAV operations

src/mainloop.h
```diff
@@ -66,7 +67,7 @@ class Mainloop {
     void del_timeout(Timeout *t);
     void mod_timeout(Timeout *t, uint32_t timeout_msec);
 
-    bool add_endpoints(const Configuration &config);
+    bool add_endpoints(const Configuration &configuration);
```
zdanek: this is not necessary change, to be reverted.

PR #469, opened 2025-11-12
Add automatic reconnection support for UART endpoints

K4HVH:
Basically re-using the logic for the TCP endpoints to allow for reconnection and handling of UART endpoint disconnections

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,8 +162,10 @@ Endpoint types (see [examples/config.sample](examples/config.sample) for the
 config file format):
 
   - UART: For telemetry radios or other serial links
-    * Configuration: UART device path/name and baudrate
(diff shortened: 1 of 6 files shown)


dancingSombrero:
WOMM. Seems legit. 
What are we waiting for to merge this in? Approval from?
@lucasdemarchi ? @jbeyerstedt ?

dancingSombrero:
Probably a more flexible alternative to #451

K4HVH:
> Probably a more flexible alternative to #451 

Yes just waiting for approval. I ended up building my own router instead which provides significantly better performance and stability. 

Was running with 20 drones connected via UART with no issues.

It also supports auto-discovery so check it out if you want. 

https://github.com/K4HVH/mav-lite

PR #473, opened 2026-02-10
feat: add interface name binding for UDP server endpoints

SylvainPastor:
### Problem

When configuring a UDP server endpoint, an explicit IP address must be
provided. If the IP changes (DHCP renewal, network reconfiguration),
the configuration file must be manually updated and the service restarted.

### Solution

Introduce an `Interface` configuration option that accepts a network
interface name (e.g. `eth0`, `eth0.10`, `wlan0`). At startup, the IPv4
address assigned to the interface is resolved via `getifaddrs()` and
used as the bind address.

Either `Address` or `Interface` must be specified. When both are present,
`Interface` takes priority.

### Example configuration
```ini
[UdpEndpoint my-endpoint]
Mode = Server
Interface = eth0
Port = 10001
```

### Changes

- `UdpEndpointConfig`: new `std::string interface` field
- `UdpEndpoint::resolve_interface_address()`: static method, resolves
  interface name to IPv4 address using `getifaddrs()`
- `UdpEndpoint::setup()`: resolves interface before calling `open()`
- `UdpEndpoint::validate_config()`: accepts `interface` as an
  alternative to `address`
- `option_table`: new `Interface` entry, `Address` changed from
  mandatory to optional
- `endpoints_test.cpp`: two new tests (`ConfigValidateInterface`,
  `ResolveInterfaceAddress`)
- `config.sample`: updated documentation and added examples

### Testing

- All 30 unit tests pass (including 2 new tests)
- Manual validation: `Interface = enp2s0` correctly resolves and binds
  to `192.168.1.230:10001`

### Limitations

- IPv4 only (consistent with existing `open_ipv4()` behavior)
- Address is resolved once at startup; a service restart is needed if
  the IP changes while running

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -78,7 +78,8 @@ const ConfFile::OptionsTable UartEndpoint::option_table[] = {
 
 const char *UdpEndpoint::section_pattern = "udpendpoint *";
 const ConfFile::OptionsTable UdpEndpoint::option_table[] = {
-    {"address",         true,   ConfFile::parse_stdstring,      OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, address)},
+    {"address",         false,   ConfFile::parse_stdstring,     OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, address)},
+    {"interface",       false,   ConfFile::parse_stdstring,     OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, interface)},
     {"mode",            true,   UdpEndpoint::parse_udp_mode,    OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, mode)},
     {"port",            false,  ConfFile::parse_ul,             OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, port)},
     {"filter",          false,  ConfFile::parse_uint32_vector,  OPTIONS_TABLE_STRUCT_FIELD(UdpEndpointConfig, allow_msg_id_out)}, // legacy AllowMsgIdOut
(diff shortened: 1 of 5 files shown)


jbeyerstedt:
> When configuring a UDP server endpoint, an explicit IP address must be provided. 

I need to object this. You can run the UDP server perfectly fine listening on `[::]` or `0.0.0.0`. If you need to limit incoming connections to a certain interface, you can use firewall rules for that.
So I don't really understand the use case for this.

(And having an IPv4 only implementation is not what I expect from in 2026.)

SylvainPastor:
> (And having an IPv4 only implementation is not what I expect from in 2026.)

Okay, I understand your point of view. I can add IPv6 support (If you want the functionality).

> So I don't really understand the use case for this.

This is a MavLink communication between two devices, each with its own bridge. Each device has a dedicated endpoint with a dedicated network interface. One of the two devices provides a global endpoint to the system.

>  you can use firewall rules for that.

However, using the name of a target interface can also be very "user-friendly" for an operator. 
It is used in conjunction with the firewall.

jbeyerstedt:
> However, using the name of a target interface can also be very "user-friendly" for an operator.

I can understand why it may look user friendly in the first place. Since a UDP socket can't bind to an interface, this option makes the more or less bold assumption that the network interface's IP won't change. But this is already not true with DHCP in IPv4 and with IPv6 it's even in the protocol design that addresses will change.
So the connection can suddenly break when the interface's IP address is not stable. And if the interface is configured to a static IP, it shouldn't be an issue to add this IP to the mavlink-router config as well.

Another issue: How should multiple IP addresses on an interface be handled? Currently you're selecting some address, but especially with IPv6 it depends on the use case which address is "the right one" (global vs. link-local vs. static ULA if present).

So my main issue is, that this features looks like an easy choice for (new) users, but can be quite deceptive instead and leads to issues in miscellaneous cases. But maybe this is just me having run into too many issues with "ease of use" features in the past where I needed to debug the abstraction and to switch to the in-depth configuration in the end. So maybe @lucasdemarchi has another view on this.

> This is a MavLink communication between two devices, each with its own bridge. Each device has a dedicated endpoint with a dedicated network interface. One of the two devices provides a global endpoint to the system.

But can't this already be achieved with the current feature set?

src/endpoint.cpp
```diff
@@ -1076,12 +1077,63 @@ UdpEndpoint::~UdpEndpoint()
...
+    struct ifaddrs *ifaddr = nullptr;
+
+    if (getifaddrs(&ifaddr) == -1) {
+        log_error("Failed to enumerate network interfaces: %m");
+        return false;
+    }
+
+    bool found = false;
+    for (auto *ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+        if (!ifa->ifa_addr)
```
jbeyerstedt: Please use curly braces for conditions. Omitting them is just quite error-prone (see the "goto fail" bug)

SylvainPastor:
I share your opinion regarding DHCP. I understand your concerns.

Possible update: By using the name of an interface, we can check its configuration (static or DHCP).

> Since a UDP socket can't bind to an interface, this option makes the more or less bold assumption that the network interface's IP won't change.

In my case, the configuration between the devices is static (these endpoints are static).

Furthermore, you can ensure that your interface is properly set up and operational with a service configuration like this example:

```
After=network-online.target NetworkManager.service
Wants=network-online.target
After=sys-subsystem-net-devices-ethX.device
BindsTo=sys-subsystem-net-devices-ethX.device
```

I would also like to explain a simple use case:

- If your endpoint binds an interface by its IP address, if the address changes (and if the configuration isn't updated or is forgotten), the bridge will fail to start because the bind will fail.
- With the interface name option, and if the IP address changes, restarting the device or service will allow the bind to succeed.

This configuration seems to me no more risky than a configuration with an IP address.

And I agree with you, when possible, the best configuration will remain the bind with "0.0.0.0".

PR #478, opened 2026-03-11
Added custom sys id to logging

Thomas-Mccabe:
* Added dynamic loading of the log endpoint sys ID to allow logging on any device
* Removed hardcoded LOG_ENDPOINT_SYSTEM_ID
* Added -L config option to load in any sys ID from 1-255

### DIFF
--- a/src/binlog.cpp
+++ b/src/binlog.cpp
@@ -32,7 +32,7 @@ bool BinLog::_logging_start_timeout()
 {
     mavlink_message_t msg;
 
-    mavlink_msg_remote_log_block_status_pack(LOG_ENDPOINT_SYSTEM_ID,
+    mavlink_msg_remote_log_block_status_pack(log_endpoint_sys_id,
                                              MAV_COMP_ID_ALL,
                                              &msg,
                                              _target_system_id,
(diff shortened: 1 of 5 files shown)


jbeyerstedt:
I'm a bit confused how this is different from the existing `LogSystemId` option in the config file. Because your description states:
> System ID to log from (1-255). If not specified, auto-detects from first autopilot heartbeat

And the config file already has:
> FCU MAVLink ID from which logs should be received. If absent, it will be set to the system ID of the first flight stack heartbeat received.

These descriptions sound like the same thing. But your code actually changes the system ID which mavlink-router uses for itself to send out MAVLink packets and does not change from which target system we request logs.

KhanMechAI:
Yeah the description could be a bit cleaner. LogSystemId sets the target (which FCU we request logs from). This PR is about the source sysid mavlink-router stamps on its own log-control messages (start/stop/ack), which is hardcoded to LOG_ENDPOINT_SYSTEM_ID 2.
Our case is multi-vehicle: multiple router instances on one network, each needing a distinct source identity on those self-generated messages so acks route back to the right instance. There's no runtime knob for that today, unless we've missed something?

Issue #416, opened 2024-02-05
Allow using hostname instead of IP address for endpoints


bringert:
In v3, trying to use a hostname instead of an IP address in the value for the --endpoint argument fails:

```
$ ./mavlink-routerd-glibc-aarch64 -e mycomputer:14550 /dev/serial0:921600
mavlink-router version v3
Could not open conf file '/etc/mavlink-router/main.conf' (No such file or directory)
UdpEndpoint CLI: Invalid IP address mycomputer
```

It would be nice to be able to use a hostname instead of having to hardcode an IP address or run an extra command to do the lookup.

bringert:
Workaround:

```
$ ./mavlink-routerd-glibc-aarch64 -e $(getent ahostsv4 mycomputer | awk '{ print $1 }' | head -1):14550 /dev/serial0:921600
```

lucasdemarchi:
yep, resolving the name when we are starting mavlink-router should be doable. Could you send a patch for that?

bringert:
I've got a first working version here: https://github.com/mavlink-router/mavlink-router/commit/76ca7e348a1a4666b796aac44cc5c3c6e46b63a5

It currently makes the tests fail, since we no longer validate IP addresses. I should probably change the validation function to do hostname lookup too.

JasonBrauer:
@lucasdemarchi Are there plans to get the hostname lookup integrated into master? We use the config file, so it would be especially useful in that case.

PR #481, opened 2026-03-30
Resolve hostnames when used as addresses

Ryanf55:
# Purpose

Support resolving hostnames instead of just addresses, such as support.ardupilot.org.

# Tests Performed

* Used a local mdns address like `jetson.local`
* Used a tailscale VPN address like `jetson.tail8371dd.ts.net`
* Added some unit tests

# Disclosure

Coded with help from claude, validated with manual testing

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -156,6 +156,117 @@ static bool validate_ip(const std::string &ip)
     return validate_ipv4(ip) || validate_ipv6(ip);
 }
 
+static bool validate_hostname(const std::string &hostname)
+{
+    if (hostname.empty() || hostname.size() > 253) {
+        return false;
+    }
+
+    size_t label_len = 0;
+
+    for (size_t i = 0; i < hostname.size(); i++) {
+        char c = hostname[i];
+
+        if (c == '.') {
+            /* dot terminates a label: must not be empty or start/end with hyphen */
+            if (label_len == 0) {
+                return false; /* empty label (leading dot or double dot) */
+            }
+            if (hostname[i - 1] == '-') {
+                return false; /* label ends with hyphen */
+            }
+            label_len = 0;
+        } else {
+            if (label_len == 0 && c == '-') {
+                return false; /* label starts with hyphen */
+            }
+            if (!isalnum((unsigned char)c) && c != '-') {
+                return false; /* invalid character */
+            }
+            label_len++;
+            if (label_len > 63) {
+                return false; /* label too long */
+            }
+        }
+    }
+
+    /* trailing dot is allowed (fully-qualified); otherwise last label must be non-empty */
+    if (label_len == 0 && hostname.back() != '.') {
+        return false;
+    }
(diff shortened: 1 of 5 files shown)


src/endpoint.cpp
```diff
@@ -156,6 +156,117 @@ static bool validate_ip(const std::string &ip)
...
+
+    char buf[INET6_ADDRSTRLEN + 2]; /* +2 for brackets around IPv6 */
+    ResolvedAddr resolved{"", false, false};
+
+    for (struct addrinfo *rp = result; rp != nullptr; rp = rp->ai_next) {
+        if (rp->ai_family == AF_INET) {
+            struct sockaddr_in *sin = (struct sockaddr_in *)rp->ai_addr;
+            inet_ntop(AF_INET, &sin->sin_addr, buf, sizeof(buf));
+            resolved = {buf, false, true};
+            break; /* prefer IPv4 over IPv6 */
```
jbeyerstedt, line 250: That's not so easy. What about systems which have global IPv6, but no legacy IP and try to reach a global resource?
In the end, there is no one-size-fits-all preference between IPv6 and IPv4. It should depend on the host configuration (similar to [RFC6724](https://www.rfc-editor.org/rfc/rfc6724), chapter 1, paragraph 3)

src/endpoint.cpp
```diff
@@ -156,6 +156,117 @@ static bool validate_ip(const std::string &ip)
...
+    if (validate_ipv4(address)) {
+        return {address, false, true};
+    }
+    if (validate_ipv6(address)) {
+        return {address, true, true};
+    }
+
+    /* Hostname — resolve via getaddrinfo.
+     * This handles plain DNS, mDNS (.local via nss-mdns/avahi),
+     * and Tailscale (*.ts.net) transparently. */
```
jbeyerstedt, line 228: Why is Tailscale named explicitly? It's just a normal domain name with regards to us.
Whether you can resolve DNS entries (from public and private name servers as well as mDNS) is up to the host configuration.

jbeyerstedt:
also take the discussion at #416 into consideration.

src/endpoint.cpp
Ryanf55: Good point. Reviewing the RFC and the docs on `getaddrinfo`, it appears that `getaddrinfo` already has the OS sort the results in OS-recommended priority which obeys the RFC (including host configuration). Additionally, the man page for `getaddrinfo` recommends trying the addresses in order.

https://linux.die.net/man/3/getaddrinfo

Based in the info in the RFC,the application should bind/connect each element in the list from `getaddrinfo` until one works. Would you like me to implement that instead?

src/endpoint.cpp
Ryanf55: It was just an example. I'll remove it.

src/endpoint.cpp
lucasdemarchi: yes, I think that would be a good option.

PR #483, opened 2026-04-07
CI: Update CI

lucasdemarchi:


jbeyerstedt:
We also need to update the "checkout" action because Github has deprecated the old NodeJS version.

jbeyerstedt (approved):

jbeyerstedt:
Unfortunately the CI still fails because of the "float-equal" check and the "gtest" source and I wasn't able to find a way to solve that with meson. Tried to move the compiler flags to the "executable" call, but that didn't help.

PR #484, opened 2026-04-26
FreeBSD 15 compat (libepoll-shim + UART/termios + IPv4 jails + example fixes)

Inkeer:
# FreeBSD 15 compat (libepoll-shim + UART/termios + IPv4 jails + example fixes)

This series adds FreeBSD 15.0-RELEASE arm64 support to
mavlink-router by gating Linux-only headers + APIs on
`#ifdef __linux__` and providing POSIX-equivalent fallbacks
elsewhere. Verified end-to-end on the platform in question;
Linux builds are unchanged.

## Summary of changes

| # | Subject | What changes | Why |
|---|---|---|---|
| 1 | Gate Linux UAPI headers + termios on `__linux__` | `<asm/ioctls.h>`, `<asm/termbits.h>`, `<linux/serial.h>`, the `TIOCGSERIAL`/`TIOCSSERIAL` block, and the `set_latency_failed:` label inside `endpoint.cpp` are wrapped in `#ifdef __linux__`. POSIX `tcflush()` replaces the Linux-only `ioctl(fd, TCFLSH, ...)`. `xtermios.h` gains a non-Linux shim that maps `termios2 → struct termios` (via `#define`, so `struct termios2 tc` declarations stay valid) and stubs out `TCGETS2 / TCSETS2 / BOTHER / CBAUD`. | These headers don't exist on non-Linux. The label-and-block guard pattern keeps the success-path `tcflush()` reachable in both branches. |
| 2 | Route epoll through libepoll-shim, neutralise `close()` macro, `getprogname()` | Top-level `meson.build` declares `dep_epoll_shim = dependency('epoll-shim')` on non-Linux and threads it into the executables. `xtermios.h` zero-stubs the Linux-only termios flags `endpoint.cpp` masks (OLCUC/OFILL/OFDEL/...) and aliases `IPV6_ADD_MEMBERSHIP` → `IPV6_JOIN_GROUP`. `#undef close` after `#include "mainloop.h"` in `endpoint.cpp` so `libepoll-shim`'s `#define close(...) epoll_shim_close(__VA_ARGS__)` macro doesn't eat C++ class member names. `main.cpp` swaps glibc's `program_invocation_short_name` for POSIX `getprogname()` under `#ifndef __linux__`. `mainloop.cpp` gets explicit `<sys/socket.h>`/`<netinet/in.h>`/`<unistd.h>` includes (Linux pulls them transitively via `asm/termbits.h`; FreeBSD does not). | FreeBSD has no `<sys/epoll.h>`; `libepoll-shim` provides a kqueue-backed equivalent. The `#undef close` is a workaround for the macro design rather than a rename of the member function — see "Open question" below. |
| 3 | AF_INET fallback for `Mainloop::tcp_open` | On `__FreeBSD__`, the struct type, `socket()` family, and field writes drop to IPv4. | Non-vnet FreeBSD jails without an explicit `ip6.addr` deny IPv6 socket creation entirely (`EPROTONOSUPPORT`). The TCP debug port targets local MAVProxy / Wireshark, IPv4 is sufficient. |
| 4 | BSD-compat `reset_uart` in `xtermios.cpp` | Linux body wrapped in `#ifdef __linux__`; non-Linux branch uses `cfmakeraw` + a small flag tweak set (CLOCAL\|CREAD, no CRTSCTS, VMIN=0/VTIME=0) producing a known-good raw mode for USB CDC ACM. | The Linux body's hand-built 32-byte `c_cc` defaults trip FreeBSD's `NCCS=20` static_assert. CDC ACM's host-side termios state is largely advisory anyway; for real RS-232 / UART at non-standard baud rates a fuller cfsetspeed-based port is the natural follow-up. |
| 5 | examples: fix FreeBSD compile (sockaddr shadow + missing `<netinet/in.h>`) | Renames the local `static struct sockaddr_in sockaddr;` variable to `srv_addr` in `arm-authorizer.cpp`, `heartbeat-print.cpp`, `px4-offboard-mode.cpp` (the local name was shadowing the global `struct sockaddr` typedef; FreeBSD libc++ + recent Linux clang reject it). Adds explicit `<netinet/in.h>` next to `<sys/socket.h>` (Linux pulls it transitively, POSIX does not). | Behaviour unchanged on Linux; lets the examples build out-of-the-box on FreeBSD. |

## Verification

* Tested on FreeBSD 15.0-RELEASE arm64 (Parallels VM hosting a
  jail) with `meson 1.10.2`, `ninja 1.13.2`, clang in the
  base system.
* `meson setup build --buildtype=release` + `ninja -C build`
  completes with **0 warnings, 0 errors** for both the library
  and the three examples.
* `build/src/mavlink-routerd --version` reports
  `mavlink-router version 3`.
* The same series, applied to a fresh upstream `master` clone
  in the FreeBSD jail used by the BUREVII project, has been
  serving live MAVLink traffic between QGC and an Ardal MCU
  via UART since the patch set was put together.

## Tradeoffs / open questions

These are conscious choices on this submission; happy to
iterate on any in review:

* **`#undef close` workaround vs. member rename.**
  `Tcp/UdpEndpoint::close()` member names collide with
  `libepoll-shim`'s `close(...)` macro. Renaming
  `Tcp/UdpEndpoint::close()` → `close_fd()` would be the
  language-level fix; this series keeps the change footprint
  minimal and uses the file-local `#undef` instead. Happy to
  switch if you'd prefer the rename.
* **`#ifdef __linux__` gates vs. `#ifndef __FreeBSD__` /
  feature detection.** Most guards target Linux UAPI headers
  that, by definition, only exist on Linux, so `__linux__` is
  the most semantically correct gate. NetBSD / OpenBSD /
  DragonFly support would need additional gates or feature
  detection; this series targets FreeBSD specifically. Happy
  to broaden if maintainers prefer.
* **`xtermios.cpp` BSD branch.** Uses `cfmakeraw` rather than
  the Linux body's verbatim "stty-sane" defaults. Equivalent
  for USB CDC ACM (the platform's primary physical link); for
  real RS-232 at non-standard baud rates a fuller port using
  `cfsetspeed` + per-flag handling is the natural follow-up.

## Acknowledgements

The series formalises a patch set that has been carried
out-of-tree by the [BUREVII CPE OS](https://github.com/Inkeer/FreeBSD)
project since April 2026; the underlying re-applied patch
scripts in
`scripts/freebsd-mavlink-router/` over there have served the
project's lab VM and field deployments. Submitting in-tree so
future FreeBSD users don't have to reinvent the wheel.

### DIFF
--- a/src/common/xtermios.h
+++ b/src/common/xtermios.h
@@ -22,3 +22,34 @@
  * same time
  */
 int reset_uart(int fd);
+
+/*
+ * FreeBSD compat shim. Linux's <asm/termbits.h> defines
+ * struct termios2, TCGETS2, TCSETS2, BOTHER, CBAUD. FreeBSD has
+ * plain POSIX struct termios with c_ispeed / c_ospeed. Mapping
+ * termios2 to termios + TCGETS2/TCSETS2 to TIOCGETA/TIOCSETA lets
+ * endpoint.cpp's set_speed, set_flow_control and open paths
+ * compile unchanged. BOTHER and CBAUD become no-ops because
+ * FreeBSD does not pack baud rate into c_cflag.
+ *
+ * termios2 is a #define rather than a typedef so that
+ * "struct termios2 tc" declarations expand to "struct termios tc"
+ * (C++ does not allow the "struct" keyword with a typedef alias).
+ */
+#ifndef __linux__
+#include <termios.h>
+#include <sys/ioctl.h>
+#define termios2 termios
+#ifndef TCGETS2
+#define TCGETS2 TIOCGETA
+#endif
+#ifndef TCSETS2
+#define TCSETS2 TIOCSETA
+#endif
+#ifndef BOTHER
+#define BOTHER 0
+#endif
+#ifndef CBAUD
+#define CBAUD 0
+#endif
+#endif /* __linux__ */
--- a/src/common/xtermios.h
+++ b/src/common/xtermios.h
@@ -53,3 +53,43 @@ int reset_uart(int fd);
 #define CBAUD 0
 #endif
 #endif /* __linux__ */
+
+/* Linux-only termios output flags used by endpoint.cpp; map to
+ * 0 so bitwise masks become no-ops on POSIX. */
+#ifndef __linux__
+#ifndef OLCUC
+#define OLCUC 0
+#endif
+#ifndef OFILL
+#define OFILL 0
+#endif
+#ifndef OFDEL
+#define OFDEL 0
+#endif
+#ifndef ONLRET
+#define ONLRET 0
+#endif
+#ifndef IUCLC
+#define IUCLC 0
+#endif
+#ifndef IUTF8
+#define IUTF8 0
+#endif
+#ifndef XCASE
+#define XCASE 0
+#endif
+#ifndef ECHOPRT
+#define ECHOPRT 0
+#endif
+#ifndef ECHOCTL
+#define ECHOCTL 0
+#endif
+#ifndef ECHOKE
+#define ECHOKE 0
+#endif
+/* Linux defines IPV6_ADD_MEMBERSHIP in <netinet/in.h>; POSIX /
+ * BSD only define IPV6_JOIN_GROUP. They identify the same option. */
+#ifndef IPV6_ADD_MEMBERSHIP
+#define IPV6_ADD_MEMBERSHIP IPV6_JOIN_GROUP
+#endif
+#endif /* __linux__ */
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -49,6 +49,18 @@
 
 #include "mainloop.h"
 
+/* Drop libepoll-shim's `#define close(...) epoll_shim_close(...)`
+ * macro after all system includes are pulled in. Without this the
+ * macro would substitute every `close()` call site, including
+ * member functions of TcpEndpoint, breaking compilation. epoll
+ * fds in this code base are only opened/closed in mainloop.cpp,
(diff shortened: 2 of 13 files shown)


src/endpoint.cpp
```diff
@@ -916,6 +918,7 @@ bool UartEndpoint::open(const char *path)
     // chip sets if their driver does not support this
     // configuration request
 
+#ifdef __linux__
```
lucasdemarchi, line 921: I don't like multi-OS support added like this. There should be no ifdefs for OS in the common code.

If you want it ever merged, then it will need to be abstracted by implementing the paths taken on Linux. This can be done by adding a `-I` to build options to bring in appropriate headers and overrides.

src/common/xtermios.h
```diff
@@ -22,3 +22,34 @@
...
+ * termios2 to termios + TCGETS2/TCSETS2 to TIOCGETA/TIOCSETA lets
+ * endpoint.cpp's set_speed, set_flow_control and open paths
+ * compile unchanged. BOTHER and CBAUD become no-ops because
+ * FreeBSD does not pack baud rate into c_cflag.
+ *
+ * termios2 is a #define rather than a typedef so that
+ * "struct termios2 tc" declarations expand to "struct termios tc"
+ * (C++ does not allow the "struct" keyword with a typedef alias).
+ */
+#ifndef __linux__
```
lucasdemarchi, line 39: See how missing/  is handled in kmod: https://github.com/kmod-project/kmod/blob/master/shared/missing.h

We should have a similar header for platforms to override or declare additional macros/symbols.

PR #485, opened 2026-04-28
Delay log stop after disarm to log post-disarm data in while-armed mode

JonasPerolini:
When `LogMode=while-armed` is used, post-disarm data is lost. For example, in PX4 the post-flight performance counters are not logged because once the drone disarms, the `VEHICLE_CMD_LOGGING_STOP` is received (for this specific example, a separate PR in PX4 is required to log the perf counters on disarm instead of on log stop). 

Solution: add a `LogStopDelay` to keep collecting post-disarm logger data before stopping the log. It defaults to `0`, which preserves the existing behavior of stopping immediately on disarm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Or in order to install to another root directory:
 There are two ways to configure mavlink-router: Configuration file(s) and
 command line parameters. You can use just config files or just CLI options or
 even both at the same time. The CLI options will be merged with the settings
-from the config file in the latter case.  
+from the config file in the latter case.
 The configuration file gives more fine-grained control over the endpoints while
 the CLI options enable a quick configuration. When using the systemd unit
 added by the install step, it's recommended to use the configuration file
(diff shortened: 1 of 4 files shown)


fallenmi (approved):
I reviewed exact base `2362c620f483cef1edd574fb962a373a288e4b9e` to exact head `5e10b6b7c9590e3d1e264202112fda8989108e32`.

The delayed-stop state machine looks correct: disarm creates at most one timer, repeated disarmed heartbeats keep the original deadline, re-arm cancels the pending stop without rotating the active file, and the callback clears its pointer before entering the common stop path. The default zero-delay behavior remains immediate, and oversized values are rejected before conversion to the mainloop's millisecond timeout.

On Linux arm64, the exact-head ASan daemon passed the upstream routing test and an independent real-daemon MAVLink heartbeat oracle covering repeated disarm, retained post-disarm telemetry, re-arm cancellation and continued writes, a later clean delayed stop, zero delay, and overflow fallback. The exact base stopped immediately under the same two-second config, while the head retained logging through the configured deadline. The repository's 30 unit cases also passed ten repetitions (300 case runs), and the clang-format 13 and `git diff --check` checks are clean.

Disclosure: I used OpenAI Codex to assist this review; I verified the exact revisions, code paths, build, tests, and daemon-oracle results.

JonasPerolini:
Hi @fallenmi, thanks for the review! Any more changes required? I cannot merge on this repo.

PR #486, opened 2026-05-16
logendpoint: capture point-to-point messages

### LINKED ISSUE #482: .tlog seems to only capture broadcast messages, not point-to-point messages
Hi,
I tried running a mavrouter connecting a PX4 SITL (sysid 1) with a mocked pymavlink GCS (sysid 255) and a custom monitoring application (sysid 254). I actived LogTelemetry, but the .tlog file only contains broadcast telemetry, e.g. HEARTBEAT, and not targeted commands, e.g. COMMAND_INT or COMMAND_ACK. To me it seems that either: I have not registered the logging sysid correctly or: The logging endpoint is registered as just another endpoint, so it only receives broadcast messages
```
# MAVLink Router Configuration

[General]
MavlinkDialect=ardupilotmega
ReportStats=false
SnifferSysid=254
DebugLogLevel=info
Log=/logs
LogMode=always
LogTelemetry=true

# GCS endpoint - receives from Mock GCS on port 14550
[UdpEndpoint gcs]
Mode=server
Address=0.0.0.0
Port=14550

# SITL endpoint - receives from PX4 SITL on port 14540
[UdpEndpoint sitl]
Mode=server
Address=0.0.0.0
Port=14540

# Monitor endpoint - MAVRouter connects to the monitor
[UdpEndpoint monitor]
Mode=normal
Address=172.20.0.30
Port=14560
```

evanofficial:
Fixes #482

## Problem

`LogEndpoint` inherits `Endpoint::accept_msg`, which only accepts a message when the target sysid/compid matches one the endpoint owns (or the message is broadcast, or the endpoint is configured as the `SnifferSysid`). A log endpoint only registers `LOG_ENDPOINT_SYSTEM_ID = 2` in its constructor and never reads inbound traffic, so it never accumulates the FCU/GCS sysids the way a UART/UDP/TCP endpoint does, and it has no way to satisfy the sniffer rule either.

The result is what the reporter described: with the config in #482 (PX4 SITL sysid 1 + mock GCS sysid 255 + monitor sysid 254 + `LogTelemetry=true`), the `.tlog` only captures broadcast frames such as `HEARTBEAT`. Anything addressed with an explicit `target_system`/`target_component` (e.g. `COMMAND_INT`, `COMMAND_ACK`, parameter exchanges) is silently rejected by `accept_msg` and never reaches `TLog::write_msg`.

## Fix

Make `Endpoint::accept_msg` virtual and override it in `LogEndpoint` to accept every routed message, preserving the self-loop check so `BinLog`/`ULog`'s outbound `REMOTE_LOG_DATA_BLOCK` control frames (sourced from `LOG_ENDPOINT_SYSTEM_ID`) are still rejected and never written back into the log.

`TLog` becomes a true passive sniffer of the bus, which is the expected behaviour for a `.tlog`. `BinLog` and `ULog` already filter to the specific `msg_id` they care about inside their own `write_msg`, so their on-disk output is unchanged — they now just see more candidate messages from the router and discard the ones they don't want, exactly as before.

The `TcpEndpoint::accept_msg` declaration also gets an `override` keyword now that the base is virtual; the behaviour there is unchanged.

## Test

Added `LogEndpointTest.AcceptMsg_CapturesTargetedMessages` in `src/endpoints_test.cpp` covering:
- broadcast still accepted
- point-to-point GCS→FCU accepted (the case from the bug report)
- point-to-point FCU→GCS accepted
- self-sourced (`LOG_ENDPOINT_SYSTEM_ID`) still rejected to avoid log feedback

I did not have a working meson/ninja toolchain on this machine to run the full suite — reviewers please run `meson test -C build` to confirm.

### DIFF
--- a/src/endpoint.h
+++ b/src/endpoint.h
@@ -177,7 +177,7 @@ public:
         return has_sys_comp_id(sys_comp_id);
     }
 
-    AcceptState accept_msg(const struct buffer *pbuf) const;
+    virtual AcceptState accept_msg(const struct buffer *pbuf) const;
 
     void filter_add_allowed_out_msg_id(uint32_t msg_id)
     {
@@ -369,7 +369,7 @@ public:
     bool is_valid() override { return _valid; };
     bool is_critical() override { return false; };
 
-    Endpoint::AcceptState accept_msg(const struct buffer *pbuf) const;
+    Endpoint::AcceptState accept_msg(const struct buffer *pbuf) const override;
 
     int accept(int listener_fd);        ///< accept incoming connection
     bool setup(TcpEndpointConfig conf); ///< open connection and apply config
--- a/src/endpoints_test.cpp
+++ b/src/endpoints_test.cpp
@@ -734,3 +734,43 @@ TEST(LogEndpointTest, Init)
     TLog tlog{conf};
     EXPECT_EQ(tlog.get_type(), ENDPOINT_TYPE_LOG);
 }
+
+TEST(LogEndpointTest, AcceptMsg_CapturesTargetedMessages)
+{
+    LogOptions conf;
+    conf.logs_dir = "./";
+
+    TLog tlog{conf};
+    buffer test_msg;
+    test_msg.curr.msg_id = 1;
+
+    // broadcast message: accepted (same as before)
+    test_msg.curr.src_sysid = 1;
+    test_msg.curr.src_compid = 1;
+    test_msg.curr.target_sysid = 0;
+    test_msg.curr.target_compid = 0;
+    EXPECT_EQ(tlog.accept_msg(&test_msg), Endpoint::AcceptState::Accepted);
+
+    // point-to-point message addressed to the FCU: must be captured even
+    // though the log endpoint has neither its sysid nor a matching sniffer
+    test_msg.curr.src_sysid = 255;
+    test_msg.curr.src_compid = 190;
+    test_msg.curr.target_sysid = 1;
+    test_msg.curr.target_compid = 1;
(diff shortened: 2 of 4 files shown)


fallenmi (changes requested):
The new log acceptance path makes a passive observer satisfy delivery to an otherwise nonexistent target. `Mainloop::route_msg()` clears `unknown` for every endpoint returning `Accepted`, while this override now returns `Accepted` for every non-self message. Consequently, merely enabling a log suppresses the router's `messages to unknown endpoints` warning for misaddressed commands even when no UART, UDP, or TCP endpoint knows the target.

I reproduced this with identical exact-ref builds. An oracle configured only the passive log endpoint and routed one `COMMAND_LONG` from `1/1` to absent `77/1`. Exact base `2362c620f483cef1edd574fb962a373a288e4b9e` reports `1 messages to unknown endpoints in the last 5 seconds`; exact head `28b3807171952879f24def6f00ea81ac344d4b5b` emits no unknown-target warning. The asserted base/head pair passed three repetitions under ASan/UBSan. Both exact builds completed and the repository's `mainloop` and `endpoints` test targets passed twice on each revision.

Please separate passive capture from routable-destination acceptance so the logger still sees point-to-point traffic without clearing the message's unknown-target state (for example, with a distinct observer/sniffed result or equivalent endpoint property). Please also add a Mainloop-level regression test that enables logging, routes to an absent sysid/compid, and preserves unknown-destination accounting; the current tests call `LogEndpoint::accept_msg()` in isolation and cannot catch this interaction.

AI disclosure: I used OpenAI Codex to inspect and build the exact revisions, construct and run the routing oracle, and draft this review. I verified the source lines, commit IDs, outputs, and conclusion.

PR #492, opened 2026-06-08
ulog: bound attacker-controlled LOGGING_DATA length/offset (OOB read/write)

### LINKED ISSUE #487: OOB read/write in ULog::_logging_data_process via attacker-controlled first_message_offset
**Affected:** `src/ulog.cpp` @ 2362c62 (HEAD as of 2026-03-30)

**Summary**

`ULog::_logging_data_process` does not validate that `first_message_offset` (a uint8_t field from the MAVLink LOGGING_DATA payload) is `<= msg->length` before subtraction. With `begin > length`, the subtraction wraps modulo 256, producing a much larger length than was bounds-checked. The subsequent `memcpy` reads past the 249-byte `msg->data[]` field and writes past the buffer-fullness check that was performed against the original (pre-underflow) length.

**Location**

`src/ulog.cpp` lines 275–308:

```cpp
if ((_buffer_len + msg->length) > BUFFER_LEN) {   // checked against original length
    log_warning("Buffer full, dropping everything on buffer");
    _buffer_len = 0;
    _waiting_first_msg_offset = true;
}
...
uint8_t begin = 0;
if (_waiting_first_msg_offset) {
    if (msg->first_message_offset == NO_FIRST_MSG_OFFSET) {
        return;
    }
    _waiting_first_msg_offset = false;
    begin = msg->first_message_offset;             // attacker-controlled 0..254
}
...
msg->length = msg->length - begin;                 // uint8 underflow if begin > length
memcpy(&_buffer[_buffer_index + _buffer_len],
       &msg->data[begin],                          // OOB read of msg->data
       msg->length);                               // wrapped length unchecked vs _buffer
```

**Concrete trigger**

`msg->length = 50`, `msg->first_message_offset = 200`. After subtraction `msg->length = 256 - 150 = 106` (well, `(50 - 200) & 0xFF = 106`). The buffer-fullness check was performed against 50; the actual write is 106. With `_buffer_len` near `BUFFER_LEN - 50`, the write overflows `_buffer[2048]` by up to ~56 bytes into adjacent fields of the `ULog` instance (`_buffer_len`, `_buffer_index`, `_buffer_partial[]`).

Corrupting `_buffer_index` / `_buffer_len` with attacker bytes allows a follow-up packet to drive further out-of-bounds access in `_logging_flush`.

**Preconditions**

1. `Log` directive configured and `LogMode = always` or `while_armed` (and target system reports armed).
2. `_waiting_first_msg_offset == true`, set when a sequence drop occurs or the buffer fills (lines 268, 279). An attacker who can inject MAVLink traffic can force either state with a single skipped-sequence packet, then send the crafted packet.
3. CRC must be valid — trivial since CRC-extra bytes are public.

**Suggested fix**

```cpp
if (begin > msg->length) {
    return;  // or log + reset state
}
msg->length -= begin;
```

**Reachability note**

mavlink-router is deployed in ground stations and onboard companion computers in PX4/ArduPilot-based platforms including Blue UAS-eligible builds. Attacker needs to be on the MAVLink network (UDP/TCP endpoint) or have access to a peer device.

Found during a focused static review of the parser/log code paths.

### LINKED ISSUE #488: Off-by-16 in ULog::_logging_data_process header-handling memmove
**Affected:** `src/ulog.cpp` @ 2362c62

**Summary**

In the `_waiting_header` branch of `ULog::_logging_data_process`, the `memmove` that shifts past the ULog magic uses the pre-decremented `msg->length`. For attacker-supplied `msg->length > 233`, this reads beyond the 249-byte `msg->data[]` and writes beyond its end into adjacent rx_buf memory.

**Location**

`src/ulog.cpp:255-259`:

```cpp
_buffer_partial_len = ULOG_HEADER_SIZE;
memcpy(_buffer_partial, msg->data, ULOG_HEADER_SIZE);

memmove(msg->data, &msg->data[ULOG_HEADER_SIZE], msg->length);  // length pre-decrement
msg->length -= ULOG_HEADER_SIZE;
```

For `msg->length = 255`, this reads `msg->data[16..270]` (21 bytes past the 249-byte field) and writes 255 bytes to a 249-byte destination (5 bytes past).

**Preconditions**

`_waiting_header == true` (initial state after `start()` or restart) + a single LOGGING_DATA message with valid ULOG magic and `length > 233`.

**Suggested fix**

Decrement first:

```cpp
msg->length -= ULOG_HEADER_SIZE;
memmove(msg->data, &msg->data[ULOG_HEADER_SIZE], msg->length);
```

**Impact**

OOB read into adjacent rx_buf bytes (checksum/signature/possible next-packet region). OOB write of up to ~6 bytes into the same area. Lower severity than #487 since the corrupted bytes are typically about to be discarded as part of normal rx_buf advancement, but still an out-of-bounds memory access on attacker-controlled length.

Found during the same review pass as #487.

crook3dfingers:
## Summary

`ULog::_logging_data_process()` reassembles incoming `LOGGING_DATA` frames using
three attacker-controlled fields of `mavlink_logging_data_t` (`length`,
`first_message_offset`, `data[249]`) without validating their relationship,
producing two out-of-bounds accesses on the log-reassembly path. Both are
reachable whenever a ULog logging endpoint is configured (`Log`/`logs_dir`),
from any peer that can speak MAVLink to the router (`accept_msg` does no
authentication). Same root cause as #487 and #488.

### 1. Header strip — OOB read + bounded OOB write

While `_waiting_header`, `msg->length` (≤255) is used directly as the `memmove`
count *before* the 16-byte ULog header is subtracted, and is never bounded to
`sizeof(msg->data)` (249):

```c
memmove(msg->data, &msg->data[ULOG_HEADER_SIZE], msg->length);
msg->length -= ULOG_HEADER_SIZE;
```

With `length = 255` the `memmove` reads `data[16..270]` and writes `data[0..254]`
— past the 249-byte field. In the MAVLink-2 trimmed-zeros dispatch the message is
a stack local, so the write is a small stack overflow.

### 2. Reassembly — `first_message_offset` underflow → OOB read

`begin = first_message_offset` is excluded only when `== 255`, and is never
checked against `length`:

```c
msg->length = msg->length - begin;   // uint8 underflow when begin > length
memcpy(&_buffer[...], &msg->data[begin], msg->length);
```

`length - begin` underflows in `uint8_t` (e.g. `1 - 2 → 255`); the `memcpy` then
reads up to ~255 bytes from `&data[begin]`, well past `data[]`. The over-read
bytes are persisted into the `.ulg` file.

## Fix

- Clamp `msg->length` to `sizeof(msg->data)` and drop the header *before* the
  `memmove`, so it shifts only the payload and stays in bounds.
- Reject frames where `begin > length` or `length > sizeof(data)` rather than
  reading out of bounds.

Malformed frames are dropped (with a warning) instead of crashing or leaking.

## Testing

Builds clean (`meson setup build && ninja -C build`, all targets). Well-formed
ULog streaming (header strip + offset-based reassembly) is unaffected, since
valid frames always satisfy `begin ≤ length ≤ sizeof(data)`.

Fixes #487
Fixes #488

### DIFF
--- a/src/ulog.cpp
+++ b/src/ulog.cpp
@@ -255,8 +255,17 @@ void ULog::_logging_data_process(mavlink_logging_data_t *msg)
         _buffer_partial_len = ULOG_HEADER_SIZE;
         memcpy(_buffer_partial, msg->data, ULOG_HEADER_SIZE);
 
-        memmove(msg->data, &msg->data[ULOG_HEADER_SIZE], msg->length);
+        /*
+         * msg->length is an attacker-controlled wire field (up to 255) but
+         * msg->data[] is only sizeof(msg->data) bytes. Clamp before using it,
+         * and drop the header (shifting only the payload) so the memmove reads
+         * and writes stay inside msg->data[].
+         */
+        if (msg->length > sizeof(msg->data)) {
+            msg->length = sizeof(msg->data);
+        }
         msg->length -= ULOG_HEADER_SIZE;
+        memmove(msg->data, &msg->data[ULOG_HEADER_SIZE], msg->length);
         _waiting_header = false;
     }
 
@@ -301,6 +310,17 @@ void ULog::_logging_data_process(mavlink_logging_data_t *msg)
     }
 
     if (!msg->length) {
+        return;
+    }
+
+    /*
+     * begin (first_message_offset) and length are untrusted wire fields. Without
+     * begin <= length the subtraction underflows in uint8 and the memcpy reads far
+     * past msg->data[]; length itself must also fit msg->data[]. Drop if either
+     * relationship is violated rather than reading out of bounds.
+     */
+    if (begin > msg->length || msg->length > sizeof(msg->data)) {
+        log_warning("Invalid ULog LOGGING_DATA offset/length, dropping message");
         return;
     }
 


fallenmi (approved):
Reviewed at exact head `4c76135454e0de04ce5abad632412da59ca62084` against base `2362c620f483cef1edd574fb962a373a288e4b9e` with an external two-case ASan/UBSan regression oracle covering:

- `LOGGING_DATA.length == 255` while waiting for the ULog header;
- recovery with `first_message_offset > length`.

The base aborts on the first case with an ASan stack-buffer-overflow (`exit 134`); this head passes both cases (`exit 0`). The upper-bound clamp now precedes the header `memmove`, and the offset/length relationship is rejected before the subtraction and final `memcpy`.

Approved.

Reviewed by OpenAI Codex for @fallenmi. No repository files were changed.

PR #493, opened 2026-06-10
Add UART RetryTimeout and reconnect support for PX4 USB serial recovery

teyssieuman:
## Summary of work completed

- Added `RetryTimeout` to `UartEndpointConfig` and UART config parsing.
- Updated `UartEndpoint::setup()` so initial open failure is non-fatal when retry is enabled.
- Implemented reconnect scheduling:
  - `_schedule_reconnect()`
  - `_retry_timeout_cb()`
  - `reopen()`
  - `_close()` / `_configure_port()`
- Detected runtime UART failures in `UartEndpoint::_read_msg()` and scheduled reconnect instead of exiting.
- Added/updated unit tests in endpoints_test.cpp for:
  - missing device retry scheduling
  - runtime disconnect retry scheduling
- Documented `RetryTimeout` in config.sample.
- Verified with `ninja -C build test` and created branch `uart-reconnect-retry-timeout` with the commit.

### DIFF
--- a/examples/config.sample
+++ b/examples/config.sample
@@ -116,6 +116,11 @@
 # Default: <false>
 #FlowControl = false
 
+# Retry timeout in seconds when opening the UART device fails or the device is
+# disconnected. Set to 0 to disable automatic reconnect.
+# Default: 0 (disabled)
+#RetryTimeout = 5
+
 # Only allow specified MAVLink message IDs to be sent via this endpoint. An
 # empty list allows all message IDs.
 # Format: Comma separated list of integers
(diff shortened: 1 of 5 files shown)


fallenmi (changes requested):
`RetryTimeout` currently stops after the first failed retry. `_retry_timeout_cb()` returns `true` when `reopen()` fails, which should leave the periodic timeout active, but `reopen()` first calls `del_timeout(_retry_timeout_timer)` and clears the pointer before attempting `open()`. `del_timeout()` marks that timeout for deletion, so `Mainloop::_del_timeouts()` removes it after the callback even though the callback returned `true`.

I reproduced this on exact head `a5d01f75636674f406f061484acc17546ad88259` with the sanitized daemon and `RetryTimeout=1`. As a control, a valid PTY appearing at 0.4 seconds connected on the first retry tick. When the same valid PTY appeared at 1.4 seconds, after the first tick had failed, it was still disconnected at 3.2 seconds: the log contained only the initial open failure and the first retry failure, and the daemon had no fd for the tty. Removing only the premature `del_timeout()`/pointer-clear block made that late device reconnect on the second tick in the same oracle.

Please keep the timeout alive across failed `open()`, configuration, and `add_fd()` attempts, and let the callback cancel/clear it only after a successful reconnect. Please also extend the test past one failed timer callback: make the device unavailable for the first retry, expose it before the second, and assert that the endpoint reconnects.

The exact head built with Clang 21 plus ASan/UBSan, and `endpoints_test` (30 tests) and `mainloop_test` (2 tests) each passed twice, 64/64 executions. Upstream CI is separately stopped at the clang-format check, so its two build jobs are skipped.

Disclosure: I used OpenAI Codex to inspect and build the exact revisions and run the PTY reconnect oracle; I verified the result.

PR #495, opened 2026-07-13
CI: Omit "float-equal" warning in dependencies

jbeyerstedt:
The "gtest" library uses some exact float comparison for some reason which is disallowed by our default compiler flags (which enable the "float-equal" warning and use warnings as errors".

This PR attempts to only apply the compiler warnings to our sources, but not the dependencies.

Builds on top of #483 which fixes some deprecated CI dependencies.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,10 +109,10 @@ jobs:
           overwrite: true
 
   alpine-linux:
-    name: alpine 3.14 (musl)
+    name: alpine 3.23 (musl)
     runs-on: ubuntu-24.04
     needs: [formatting-check]
-    container: alpine:3.14
+    container: alpine:3.23
     steps:
       - name: install dependencies
         run: apk update && apk add build-base git linux-headers pkgconf meson ninja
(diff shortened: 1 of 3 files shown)


examples/meson.build
```diff
@@ -2,17 +2,23 @@ exe_arm_authorizer = executable('arm-authorizer',
         ['arm-authorizer.cpp'],
         include_directories: [mavlink_inc, src_inc],
         install: false,
+        c_args: cc.get_supported_arguments(possible_cc_flags),
+        cpp_args: cc.get_supported_arguments(possible_cxx_flags),
```
lucasdemarchi: I don't think that works since gtest is a header-only library... the code is coming from the header when compiling our own compilation unit.

lucasdemarchi: Let me push an alternative to see if it passes CI.

jbeyerstedt: Yeah, makes sense when it's header only. But also makes it harder to fix

lucasdemarchi: I tried building it locally and it fixed it for me. Let me check why CI is still failing.

Issue #386, opened 2023-02-17
Mavlink-router is no longer defaulting to a TCP V4 Server Port.


SGDucharme:
Is this be design?  I could not find any information in the README.md file that this behavior was changed from TCP V4 to TCP V6.

My usage of Mavlink-router needs to use TCP V4.

We have modified Mainloop::tcp_open to use TCP V4 instead of TCP V6 for our use.

Would it be better to add a new command line argument that selects which version of TCP is used for the default port?

lucasdemarchi:
it should be listening to both

lucasdemarchi:
If it's not. After starting mavlink-proxy can you check all the ports it's listening to with something like netstat?  If it's not listening on TCPv4, can you send a patch to fix it?

jbeyerstedt:
I definitely have a mavlink-router instance (code version from about 2022-08) running which is reached via IPv4 (VPN) on it's TCP port. So I would be surprised, if it doesn't work any more.

If you open an IPv6 socket in linux, it will also listen to IPv6 on that port by default. It might be confusing, that `netstat -tlpn` only lists the port as "tcp6". (You can explicitly disable this dual-protocol behavior when creating the socket, but that's not something we did.)

SGDucharme:
I did some more testing, and it indeed does work.

My confusion was that netstat did not list the TCP v4 listening socket, but only the V6 address.

Thanks for the quick answers.

PR #499, opened 2026-09-21
Add IPv4 fallback for TCP server

yarreg:
When `TcpServerPort` is enabled, mavlink-router creates the TCP server socket using `AF_INET6`.

On systems where IPv6 is unavailable, socket creation fails with:

```text
TCP Server: Could not create tcp socket (Address family not supported by protocol)
```
As a result, the TCP server cannot be started even though IPv4 networking is available.

  ### Changes
  - Extract TCP socket creation and binding into tcp_bind_socket().
  - Keep the existing IPv6 TCP listener as the first choice.
  - Fall back to an IPv4 listener when IPv6 is unavailable with EAFNOSUPPORT.
  - Preserve the existing wildcard bind behavior:
      - IPv6: [::]:<port>
      - IPv4 fallback: 0.0.0.0:<port>


yarreg:
Related to #386.

The TCP server was changed from IPv4 to IPv6 by #348. This is correct for dual-stack systems, but it breaks on systems where IPv6 is unavailable: socket(AF_INET6, ...) fails with EAFNOSUPPORT.

This patch preserves the IPv6-first behavior and adds an IPv4 fallback. It does not add an explicit address-family configuration option.

